PR #33, opened 2014-12-19
Enable to modify threshold for each PD individually

YujiOkajima:
In order to introduce the threshold of new PD, I have added survival probability from threshold cut to WCSimPMTObject.cc and WCSimPMTObject.hh, and delete it from WCSimWCDigitizer.hh.
The figure and text below is the result of check with "electrontest.mac" and "verification_HitsChargeTime.C"

Processing verification_HitsChargeTime.C("../wcsimtest.root","../../WCSim_clean/verification-test-scripts/wcsimtest.root")...
Stats for the first event in your version of WCSim using ../wcsimtest.root
Number of tube hits 2540
Number of digitized tube hits 2045
Number of photoelectron hit times 4408
***********************************************************
Stats for the first event of WCSim version on GitHub using ../../WCSim_clean/verification-test-scripts/wcsimtest.root
Number of tube hits 2540
Number of digitized tube hits 2045
Number of photoelectron hit times 4408
***********************************************************
FIRST EVENT TEST PASSED: Number of hit tubes matches
FIRST EVENT TEST PASSED: Number of digitized tubes matches
FIRST EVENT TEST PASSED: Number of hit times matches
***********************************************************
ks test for # of digitized hits: 1
ks test for average charge: 1
ks test for average time: 1

check

The probability of HPD has still not been analyzed for WCSim. However the events which are cut by threshold are much less than that of PMT.
I will remove the threshold from HPD provisionally.

### DIFF
--- a/include/WCSimPMTObject.hh
+++ b/include/WCSimPMTObject.hh
@@ -20,6 +20,7 @@ public:
   virtual G4float* GetQEWavelength()=0;
   virtual G4float  GetmaxQE()=0;
   virtual float  GettimingResolution(float)=0;
+  virtual void  Threshold(double&, int&){};
   virtual G4double GetPMTGlassThickness()=0;
 };
 
@@ -41,6 +42,7 @@ public:
   G4float* GetQEWavelength();
   G4float  GetmaxQE();
   float  GettimingResolution(float);
+  void  Threshold(double&, int&);
   G4double GetPMTGlassThickness();
 };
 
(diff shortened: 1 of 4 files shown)


cwwalter:
@YujiOkajima I think this pull request is no longer relevant since we merged number #59.  Is that correct or is there anything else new in here.  If it is no longer needed please close this one.

YujiOkajima:
All right, I will close.

The threshold cut efficiency function is based on R3600, and it is possible to change the cut efficiency of new PDs, especially HPD. 
This is not implemented in #59 and I think It will be necessary in future.
If the cut efficiency of new PDs are different from R3600, I will pull request again.

cwwalter:
OK. Please raise this issue again in a new pull request if necessary after the new work on separating the digitizer and DAQ etc.

PR #34, opened 2015-01-08
Introduce HPD parameters.

YujiOkajima:
Introduced HPD parameters.
Not only parameters in WCSimPMTObject, but also dark rate in macro file is added 
(and changed PMT dark rate to 4.2 kHz following to fiTQun and SKDetSim).
Conversion factors have been also added based on the number of digitized hit from dark rate.

In order to compare HPD performance with PMT, Hyper-Kamiokande geometry with HPD has also been introduced.
I compared result of simulation in HyperK geometry with PMT and HPD (condition: 500MeV e- from origin to x-axis) by using verification_HitsChargeTime.C
The figure below is the result.
varidation
This is almost the same as the result I uploaded in issue12 comment which is the result of simulation in the same condition with old version WCSim


YujiOkajima:
I'm sorry for delayed response.
I checked the HPD parameter with different name work correctly, and there are not any problems.
Therefore I changed the name and uploaded the code.

eosullivan:
Ok, maybe one last thing. Could you make the change to vis.mac where you adjust the dark rate from 4 to 4.2 a separate pull request? It will just make it a bit easier to find later if it has it's own commit history.

eosullivan:
And I guess the change to the default conversion number for the dark noise should also be done in the pull request where you change the default dark noise rate.

eosullivan:
@YujiOkajima, Can you tell us a bit more about where you get the 4.2 kHz for the updated dark noise rate number? You mention that it is from @wilking? @fdilodov is looking at maybe using it in the new production data, so we want to make sure it is right.

YujiOkajima:
@eosullivan , I got 4.2 kHz from fiTQun code (fiTQun.parameter.dat) and some SKDetSim cards in Kamioka computer system.
fiTQun code set the dark noise rate of SK-I to 4.2 kHz.
And some card for SKDetSim simulation seems to have 4.2 kHz dark noise.
However, I checked cards for the SKDetSim simulation again and it seems not to be the only value for SK-1 simulation (maybe 4.2 kHz is set for atmpd validation and outer detector tuning. On the other hands 4.91 kHz is set for low energy events cards).
Maybe the dark noise is set for high energy events, but it have to be confirmed.
I will revert the macro file modification.

cwwalter:
@YujiOkajima Where did you get the original value you set?

YujiOkajima:
I get the dark rate from SKDetSim cards in 
/usr/local/sklib_gcc4.8.2/skdetsim-v13p90_for_atmnu_neutron/ @sukap
DS-DARK, the parameter of dark rate,  is set to 4.2 in validate_ap_sk1.card and sk1_odtune.card

cwwalter:
yes, but here you are changing from 4.0 to 4.2. I remember you originally set to 4.0 to agree with SK-I.  Where did 4.0 come from?  

Anyway, yes please revert the macro file modifications for this pull request.  We can put them in a separate one.

YujiOkajima:
I'm sorry, when I introduced the dark rate, I only heard the dark rate is 4 kHz.
and I did not know the value of SKDetSim because I did not have account of Kamioka computer.

YujiOkajima:
I reverted the vis.mac.
and modified the commit history.

eosullivan:
Ok, I will merge this pull request. Please modify vis.mac with the appropriate dark rate for both the regular PMTs and the HPDs and make another pull request with that change.

cwwalter:
OK, good job.  @YujiOkajima you might also want to announce this to the hkwg-software list and tell people how to use it.

PR #36, opened 2015-01-20
Further improved how the PMT is created WCSimDetectorConfig.cc

eosullivan:
I now return the pointer to PMT with the CreatePMTObject command to make things more concise. I also added an exit command if an unknown type of PMT is specified. Finally, I cleaned up the indentation a bit in the WCSimDetectorConfigs.cc file.

### DIFF
--- a/include/WCSimDetectorConstruction.hh
+++ b/include/WCSimDetectorConstruction.hh
@@ -85,7 +85,7 @@ public:
 
   G4float GetPMTQE(G4float, G4int, G4float, G4float, G4float);
 
-  void CreatePMTObject(G4String);
+  WCSimPMTObject *CreatePMTObject(G4String);
 
   WCSimPMTObject *  PMTptr;
   void    SetPMTPointer(WCSimPMTObject* PMT) {PMTptr = PMT;} //currently you can only save one PMT here. When we move to multiple PMTs as a future upgrade, this can be changed to an array of PMT pointers.
--- a/src/WCSimDetectorConstruction.cc
+++ b/src/WCSimDetectorConstruction.cc
@@ -218,31 +218,37 @@ G4VPhysicalVolume* WCSimDetectorConstruction::Construct()
   return physiExpHall;
 }
 
-void WCSimDetectorConstruction::CreatePMTObject(G4String PMTType)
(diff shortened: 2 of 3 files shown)


src/WCSimDetectorConstruction.cc
```diff
@@ -218,31 +218,37 @@ G4VPhysicalVolume* WCSimDetectorConstruction::Construct()
...
     WCSimDetectorConstruction::SetPMTPointer(PMT);
+    return PMT;
   }
   else if (PMTType == "HPD20inchHQE"){
     WCSimPMTObject* PMT = new HPD20inchHQE;
     WCSimDetectorConstruction::SetPMTPointer(PMT);
+    return PMT;
   }
-  else { G4cout << "Not a recognized PMT Type" << G4endl;}
+  else { G4cout << PMTType << " is not a recognized PMT Type" << G4endl; exit(1);}
```
cwwalter: Probably good to also print out that we are stopping the program for this reason.

include/WCSimDetectorConstruction.hh
```diff
@@ -85,7 +85,7 @@ public:
 
   G4float GetPMTQE(G4float, G4int, G4float, G4float, G4float);
 
-  void CreatePMTObject(G4String);
+  WCSimPMTObject *CreatePMTObject(G4String);
 
   WCSimPMTObject *  PMTptr;
   void    SetPMTPointer(WCSimPMTObject* PMT) {PMTptr = PMT;} //currently you can only save one PMT here. When we move to multiple PMTs as a future upgrade, this can be changed to an array of PMT pointers.
```
cwwalter: I wonder if SetPMTPointer and GetPMTPointer (which I guess is not being used now) should now be private member functions?  Is there a reason to keep them public now that people can call CreatePMTObject?

eosullivan: In order to get information needed by the digitizer, the GetPMTPointer function is used in WCSimWCPMT.cc and WCSimWCDigitizer.cc. I guess the SetPMTPointer function could be a private member function, in principle.

cwwalter: Oh, OK.  Let's keep the two together.

cwwalter:
OK this looks good. I will merge it.

PR #47, opened 2015-02-10
Feature/issue46

eosullivan:
Adding in the option for a 12 inch HPD photodector. This is necessary to begin the design studies for an alternative configuration in the HK detector. Except for the name, radius, and expose height, the 12 inch HPD is identical to the entry for the 20 inch HPD. The expose height is assumed to be the same as the 12 inch PMT.


eosullivan:
I didn't create a detector configuration entry for this HPD yet since I assume whoever is doing the design study will want to do this on their own with the appropriate radius, length, etc.

cwwalter:
This looks fine to me.  

As Erin said in her comment there is currently no detector defined that uses these tubes, but they are there for studies if you need them.

PR #54, opened 2015-02-28
Feature/issue50

samshort25:


aihimmel:
The file changes look good to me.  What do you think @cwwalter, @eosullivan?

cwwalter:
Nice job Sam!  

It it is not too much trouble I would like to ask you to redo this a bit.  Your code itself looks fine on a quick check but this stuff:

" Chris Walter	Merge branch 'hotfix/HK_geometry_check'	 4744f64
 Chris Walter	Merge branch 'release/hyperk-v1.2.0'	 0824503
 Sam Short	Merge remote branch 'upstream/master' into develop"

That came from your learning how to do everything is going to make the history look confusing.  I'm not sure exactly how it happened but you merged old things from as far back as August onto your current branch including the master which only contains the releases.

In fact, this should still work OK since the changes to the files are really all that matters, but the graph of the branches will have some confusing things related to Master that will make it harder to understand when people are looking at them and trying to understand what happened in the past.

So, can I ask you to try the following?  Now that you are an expert it should go fast.

1) Close this pull request.
2) Rename/move the repo on your disk to somewhere else so you still have the work you did.
3) From the github site, delete the issue50 branch on your fork (click on "branches" to do that)
4) Re-clone your fork
5) Make and change to a new issue50 branch
5) Reapply your changes, commit and then push to your fork
6) Make a new pull request

This will make the pull request look nice and clean.

Thanks!

cwwalter:
BTW, to be safe you better give the new issue50 branch a slightly different name.

I'm now sure but there may be problems if you try to use the same name as it keeps history for everything.

Maybe issue50_2 or something.

eosullivan:
Hi Sam! Congrats on your first WCSim pull request!

Once you have fixed up your branch a bit, I've noticed something in your proposed changes:
- You left WCCapPMTSpacing as the old SK value for the two SK configs. However, since we are changing photocoverage, we also need to change the photocoverage of the caps. Set WCCapPMTSpacing = (pi*WCIDDiameter/WCBarrelNumPMTHorizontal);
-I would also set the  WCCapEdgeLimit = WCIDDiameter/2.0 - WCPMTRadius; to avoid problems there with the changing number of PMTs. 
- The spacing of code for the WCBarrelNRings line is a bit off, probably just from where you are cutting and pasting. It does seem to be working ok, but it would be a bit easier to read if the extra spaces were removed. 

Have you tried running your new detectors with the visualizer turned on (using vis.mac)? Does everything look ok (no overlapping volumes, etc)?

Issue #50, opened 2015-02-18
Implement candidate HK detector options


eosullivan:
For the HK design studies, new detector options must be implemented. The detector configurations that we need right now are: 
1. SK detector with 12" HPD and 14.59% photocoverage
2. SK detector with 20" HPD and 13.51% photocoverage
3. A cylindrical detector with a height of 100m and a diameter of 69m with 12" HPD and 14.59% photocoverage
4. SK detector with 20" normal PMT and 20.27% photocoverage 
5. SK detector with 20" HPD and 20.27% photocoverage

eosullivan:
This issue is closed with pull request #55

PR #55, opened 2015-03-02
Feature/issue50 2

samshort25:
- Added the three new detector geometries as described in Issue 50. 
- Also implemented Erin's suggestions regarding the definitions of WCCapPMTSpacing and WCCapEdgeLimit.
- Ran WCSim with the new geometries. Occasionally get the warning "Potential overlap in geometry! *** This is just a warning message.". Ran WCSim with the (already implemented) SuperK geometry and notice a similar number of warnings. Assume this is OK?

### DIFF
--- a/include/WCSimDetectorConstruction.hh
+++ b/include/WCSimDetectorConstruction.hh
@@ -59,6 +59,7 @@ public:
   // Related to the WC geometry
   void SetSuperKGeometry();
   void SuperK_20inchPMT_20perCent();
+  void SuperK_20inchHPD_20perCent();
   void SuperK_12inchHPD_15perCent();
   void SuperK_20inchHPD_14perCent();
   void Cylinder_12inchHPD_15perCent();
--- a/include/WCSimDetectorConstruction.hh
+++ b/include/WCSimDetectorConstruction.hh
@@ -59,9 +59,9 @@ public:
   // Related to the WC geometry
   void SetSuperKGeometry();
   void SuperK_20inchPMT_20perCent();
-  void SuperK_12inchHPD_14perCent();
+  void SuperK_12inchHPD_15perCent();
   void SuperK_20inchHPD_14perCent();
-  void Cylinder_12inchHPD_14perCent();
+  void Cylinder_12inchHPD_15perCent();
   void DUSEL_100kton_10inch_40perCent();
   void DUSEL_100kton_10inch_HQE_12perCent();
   void DUSEL_100kton_10inch_HQE_30perCent();
--- a/include/WCSimDetectorConstruction.hh
+++ b/include/WCSimDetectorConstruction.hh
@@ -58,6 +58,9 @@ public:
 
   // Related to the WC geometry
   void SetSuperKGeometry();
+  void SuperK_12inchHPD_14perCent();
+  void SuperK_20inchHPD_14perCent();
+  void Cylinder_12inchHPD_14perCent();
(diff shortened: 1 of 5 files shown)


eosullivan:
Great, the commit history looks a lot cleaner in this pull request. I also just checked out and ran your pull request branch and things seem to be working properly. This looks good to me! I will merge this today if there are no other comments.

eosullivan:
@samshort25 , do you mind adding a fourth detector configuration with this pull request? An SK detector with 20" PMTs and 20% photocoverage (this is just like the SK detector in SetSuperKGeometry() but with a 20% photocoverage). I will update issue #50 to include this detector.

samshort25:
No problem at all. I will do it tomorrow (as I'm in the process of writing a talk for tomorrow at the moment). Do you want me to change the WCCapPMTSpacing and WCCapEdgeLimit parameters too? Or would you prefer to have them left at the SK values?

eosullivan:
Change WCCapPMTSpacing and WCCapEdgeLimit  as you did for the other detector configurations so that everything gets properly updated when you change the photocoverage.

samshort25:
Should I implement all of the proposed designs here: https://wiki.hyperk.org/DesignReport/DetectorOptimization before you merge my branch (I'm not sure if that's the correct terminology?) All just cylinders, yes?

cwwalter:
I don't want to implement every possible tank that we might simulate in the future.  Things get left around that way and everyone is too afraid to get rid of things that area already there.  

So, lets start with the proper SK options + the 1 100m tank you already put in.

@eosullivan will compare agains the table and confirm if we have the correct ones. 

In the next step if we need to do more we can add them then.

eosullivan:
Let's let the dust settle a bit today so we can make sure we have agreement among the working groups about which configurations they would like to see. I suspect in the end we will implement the four detectors that you already have in there, but the photocoverage might need to be updated slightly. We will try and match this pull request to the "Detector Configurations in WCSim" table on the detector optimization wiki page at: https://wiki.hyperk.org/DesignReport/DetectorOptimization#preview

cwwalter:
Thanks for adding the table Erin.  

 Maybe it should be:
Detector Configurations in WCSim ->
Detector Configurations Currently implemented in WCSim

And then to the table also add the WCSim detector configuration name.

eosullivan:
Good idea to add the configuration name. I've just put it on the wiki.

eosullivan:
Ok, I've updated the description of Issue 50 to reflect the new photocoverages given the information provided by the photosensor group. @samshort25 can you update the photocoverage values in this pull request? It seems like these are now finalized.

samshort25:
@eosullivan I think that's all done now. Let me know if there's anything else that needs adding.

eosullivan:
I think you should add some comments to the detector configurations in the macros (novis.mac and vis.mac) about what the actual photocoverage is, just so it is really obvious to those running the configurations that they are slightly different than the label suggests. Other than that, I think the changes look good. Let's review them one more time in the software meeting tonight and then hopefully we can get these changes merged in.

eosullivan:
Ok, as per the meeting tonight, we will add a fifth detector option (SK with 20" HPD and 20.27% photocoverage). @samshort25 , sorry about the many changes to the issue! There was consensus in the group that after the fifth detector is added in, the pull request can be merged in. I updated issue #50 with the fifth detector option.

eosullivan:
Ok, everything looks good. Thanks for implementing these, Sam!

PR #56, opened 2015-03-12
Feature/geant4.9.6

tfeusels:
This feature branch also compiles with geant4.9.6. The function call SetModel of G4OpticalBoundaryProcess that was already not used in Geant4.9.4 is removed in the newer version. Instead Geant4 uses the function call from the G4OpticalSurface to determine which model (glisur, unified or Look-up tables) to use. These calls are NOT deprecated as mentioned incorrectly in WCSimConstructMaterials.cc. If these are not set, Geant4 will use the defaults of G4OpticalSurface (see G4OpticalSurface.hh), which is "glisur" and not "unified" which we are currently using.
Expected physics (model or data) differences between geant4.9.6 and geant4.9.4 will be discussed in the Physics Lists issue.

### DIFF
--- a/src/WCSimPhysicsList.cc
+++ b/src/WCSimPhysicsList.cc
@@ -280,9 +280,6 @@ void WCSimPhysicsList::ConstructOp(){
   // theCherenkovProcess->SetMaxBetaChangePerStep(10.0);
   // theCherenkovProcess->SetTrackSecondariesFirst(true);
 
-  G4OpticalSurfaceModel themodel = unified;
-  theBoundaryProcess->SetModel(themodel);   // deprecated in 4.9.6
-
   theParticleIterator->reset();
   while( (*theParticleIterator)() )
   {
--- a/README.md
+++ b/README.md
@@ -56,3 +56,5 @@ doc/DetectorDocumentation.pdf.
 ```
 WCSim development is supported by the United States National Science Foundation.
 ```
+
+Test
--- a/README.md
+++ b/README.md
@@ -56,5 +56,3 @@ doc/DetectorDocumentation.pdf.
 ```
 WCSim development is supported by the United States National Science Foundation.
 ```
-
-Test
--- a/src/WCSimConstructMaterials.cc
+++ b/src/WCSimConstructMaterials.cc
@@ -736,7 +736,7 @@ void WCSimDetectorConstruction::ConstructMaterials()
      new G4OpticalSurface("WaterBSCellSurface");
 
    OpWaterBSSurface->SetType(dielectric_dielectric);
-   OpWaterBSSurface->SetModel(unified); // deprecated in 4.9.6
+   OpWaterBSSurface->SetModel(unified); 
    OpWaterBSSurface->SetFinish(groundfrontpainted);
    OpWaterBSSurface->SetSigmaAlpha(0.1);
 
@@ -758,7 +758,7 @@ void WCSimDetectorConstruction::ConstructMaterials()
    OpGlassCathodeSurface =
      new G4OpticalSurface("GlassCathodeSurface");
    OpGlassCathodeSurface->SetType(dielectric_dielectric);
-   OpGlassCathodeSurface->SetModel(unified);  // deprecated in 4.9.6
+   OpGlassCathodeSurface->SetModel(unified);  
    //   OpGlassCathodeSurface->SetFinish(groundbackpainted);
    OpGlassCathodeSurface->SetFinish(polished);
    //OpGlassCathodeSurface->SetSigmaAlpha(0.002);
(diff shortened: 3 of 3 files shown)


eosullivan:
Just so it is stated explicitly, does this branch also still compile with Geant4.9.4?

eosullivan:
Also, I would maybe recommend on holding off on merging this until after we tag out the release for the HK design study. In principle the change seems simple and I think it shouldn't affect things, but since it doesn't add anything for the design study I think it would benefit from having a bit more time for people to use it in case there is some change.

tfeusels:
I indeed forgot to mention it, but I compared the results with and without the very minor changes (essentially just a code cleanup) using the verification scripts and compiling with geant4.9.4. I got exactly the same results as expected.

So I don't see any reason not merge it in. It doesn't to anything to the current version, but gives people the option to start looking into the newer geant4.9.6.

As I'm working now on getting 4.10.0 working, I need to include this fix as well, so hope to branch off the merged develop.

eosullivan:
Can you attach the verification plots you made comparing the output from 4.9.4 and 4.9.6?

tfeusels:
As explained, I compared geant4.9.4 with and without the changes. The changes are the removal of an unused function call. The plots are highly uninteresting as it shows distributions on top of each other with a K-S test of 1:
FIRST EVENT TEST PASSED: Number of hit tubes matches
FIRST EVENT TEST PASSED: Number of digitized tubes matches
FIRST EVENT TEST PASSED: Number of hit times matches
***********************************************************
ks test for # of digitized hits: 1
ks test for average charge: 1
ks test for average time: 1

This feature just makes sure people CAN use geant4.9.6. This is not a detailed physics comparison as I would like to investigate that when I look into the Physics List changes that might be needed (or at least documented).

src/WCSimPhysicsList.cc
```diff
@@ -280,9 +280,6 @@ void WCSimPhysicsList::ConstructOp(){
   // theCherenkovProcess->SetMaxBetaChangePerStep(10.0);
   // theCherenkovProcess->SetTrackSecondariesFirst(true);
 
-  G4OpticalSurfaceModel themodel = unified;
-  theBoundaryProcess->SetModel(themodel);   // deprecated in 4.9.6
```
aihimmel: As far as we can tell, this function call was not actually doing anything since the actual model is set as a property of each individual surface (which we do), rather than being set globally.

aihimmel:
I agree with Erin, if possible can you post the verification plots?

As for merge order, even if the chance of problems is low, we should try to be careful since we're under some time pressure for the configuration studies.  I would propose we merge the configurations first, and tag that as a release.  Then, immediately after we merge this pull request so people can start working on sukap.  If the results really do look identical, we can tag it as a patch to the previous release and we don't slow down the start on things like MC production and fiTQun tuning.

tfeusels:
Your choice. I still don't understand what the concern with this one line removal of an unused function call is. 

compare_feature_4 9 6_fix

eosullivan:
Can you run the verification script comparing 4.9.4 and 4.9.6? I am curious to see if there are any changes to this basic output when you move between versions of Geant. My concern is that as soon as we make it possible for people to run with this newer version of Geant, people will start to do so and if we haven't characterized the differences between the versions then it might be hard to recognize differences in the outputs later.

tfeusels:
That is of course a very different test and gives different results (see below)
For all current studies and analysis one should freeze the current Geant4 version. For development of the code itself though, users should have the freedom to be able to compile in newer Geant4's. Different Geant4 will probably mean a new fitQun tuning. The discussion on the physics differences (and updated physics lists) is a separate one. 

The current tests show that
- the user CAN compile with both geant4.9.4 and 4.9.6 and that this change has no effect if one keeps on using the same geant4.9.4
- Switching to geant4.9.6 for physics studies will show differences. So people that can only use this version should only longer term development (and explicitly state their used geant4 version.
- If we at some point want to change our default Geant4 (after the upcoming release of WCSim for production), I believe we should immediately jump to 4.10.0 or even 4.10.1. The feature branch where this is possible can be ready quite soon.

Differences between Geant4 versions:
compare_feature_4 9 6_vs_4 9 4

eosullivan:
Ok, I agree with your logic that people should be able to use different versions of Geant and the capabilities of WCSim shouldn't be restricting people from doing so. I guess, as you say, we just need to make it clear to people who want to use Geant4.9.6 that it will affect the physics outputs. 

Thanks for doing this. It's a nice warm up to the big switch over to Geant 10.

cwwalter:
Thanks for these checks Tom!

After seeing this and having several discussions I think what we want to do is the following:

We will merge this to dev as soon as we make 1.2.1 for the design studies  (#55) hopefully in ~ 5 days. If anyone needs to do work with a newer G4 version please comment out lines:

G4OpticalSurfaceModel themodel = unified;
theBoundaryProcess->SetModel(themodel);     

until we are ready to merge this in.  You can uncomment them before and then bring your branch up-to-date afterwards if necessary.

Then we are going to ask everyone to keep using 4.9.4.  This means installing it in a few places like Kamioka.  When the work of the dev branch is ready we will switch everyone over to 4.10.X together. When we start doing the comparisons we will need to do we won't want the extra confusion of people using different versions of G4 in the mix. In the meantime people doing the development tests requiring 4.10 will be able to use it.

PR #57, opened 2015-03-25
Enhancement/daq1

samshort25:
Added a uniqueID for each raw hit, which is then saved to each digit object; this will enable us to link easily between the different TClonesArrays in the wcsim.root output.

### DIFF
--- a/include/WCSimRootEvent.hh
+++ b/include/WCSimRootEvent.hh
@@ -85,16 +85,21 @@ class WCSimRootCherenkovHit : public TObject {
 private:
   Int_t fTubeID;
   Int_t fTotalPe[2];
+  Int_t fUniqueID;
 
 public:
   WCSimRootCherenkovHit() {}
   WCSimRootCherenkovHit(Int_t tubeID,
 			Int_t totalPe[2]);
+  WCSimRootCherenkovHit(Int_t tubeID,
+			Int_t totalPe[2],
+			Int_t uniqueID);
 
   virtual ~WCSimRootCherenkovHit() { }
 
   Int_t GetTubeID()       const { return fTubeID;}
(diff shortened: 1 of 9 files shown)


eosullivan:
Will you be giving a presentation about this pull request at the software meeting tonight @samshort25?

samshort25:
@eosullivan I'm not sure if I can attend the meeting this evening. If I can attend, is there anything specific you'd like me to talk about for this pull request? If a one slide summary is sufficient, I can make one and post it on the meeting page?

eosullivan:
Sure, if you can't make the meeting we can talk about it here too. I guess I am just confused about why you need this unique ID. Is this a tool which is needed for some bigger project?

samshort25:
It was something we think we will need as part of the DAQ trigger studies so that we can tell which raw hits make it through the digitisation and become digits. 
This may all change once we've rewritten the digitisation, but I thought I'd add it in now so that it's available if we need it.
If you'd prefer, we can wait until the digitisation is rewritten before merging?

cwwalter:
Haven't looked at the details yet but why couldn't you just use the (tube, arrival time) pair as the unique hit ID. I think I wondered about that in email before but can't remember the answer...

samshort25:
I hadn't thought of that... I guess we could. I thought this variable might be life easier, as it's obvious what it is, but perhaps the fewer variables the better?

eosullivan:
Sam, what is the status of this pull request? Is it still necessary to include these variables (in which case, we should all look at the details of you pull request more closely) or can you get by with Chris' suggestion of using tube and arrival time (in which case, we should close this pull request)?

samshort25:
Hi Erin, we talked about this at our DAQ meeting and, whilst it's not especially transparent for non-experts, we're happy to just use the tube and arrival time. I'll close the pull request :)

PR #58, opened 2015-03-25
Add link to the nuance format documentation.

aihimmel:
No changes to code function, only adding comments.


cwwalter:
Good idea. Maybe also note is in important to exactly follow the format.

cwwalter:
OK this looks good.

PR #59, opened 2015-03-26
New PD Implementation

YujiOkajima:
The parameters of 12 inch and 20 inch Box and Line PMTs are implemented.
And, in order to reproduce the timing distribution of new PDs and collection
efficiency of all PDs,
the process of hit time smearing is moved from WCSimWCDIgitizer to
WCSimPMTObject and
implement WCSimPMTCE in source code.
In original code, There is a table of collection efficiency (CE, possibility
of output any signal from photoelectron) 
in WCSimWCSD.cc and you can use it when you set /WCSim/PMTCollEff on in
macro file.
However, this table seems not to come from SKDetSim (CE of SKDetSIm is 100%)
The figure below is the comparison between Normal QE R3600 before and after modification of code.
validation_pmt

From this figure, The new code seems to work correctly.
And dark noise was set to 8.4 kHz The conversion factor is 1.126 and validated by using the signals whose hit time is in the timing window between 600 ns and 900 ns (trigger is 950 ns) of the events in SK whose incident particle (10MeV/c electron) was generated from (1000 cm ,0,0) 
(the below figure shows hit time distribution of the events.)
time_distri

The below figure shows the result

validation_darknoise
The peak is almost 1, so the setting dark noise and measured dark noise is
the same value.
The Results of validation and setting of classes to introduce are uploaded to 
https://www-sk.icrr.u-tokyo.ac.jp/~okajima/slides/Implementation_of_newPDs3.pdf


include/WCSimDetectorConstruction.hh
```diff
@@ -89,6 +89,8 @@ public:
   G4double GetPMTSize1() {return WCPMTSize;}
 
   G4float GetPMTQE(G4float, G4int, G4float, G4float, G4float);
+  G4float GetPMTCE(G4float);
```
aihimmel: Perhaps something is missing from this pull request?  I cannot see the implementation of this function anywhere in the commit.

src/WCSimWCSD.cc
```diff
@@ -181,7 +164,7 @@ G4bool WCSimWCSD::ProcessHits(G4Step* aStep, G4TouchableHistory*)
      G4double local_y = localPosition.y();
      G4double local_z = localPosition.z();
      theta_angle = acos(fabs(local_z)/sqrt(pow(local_x,2)+pow(local_y,2)+pow(local_z,2)))/3.1415926*180.;
-     effectiveAngularEfficiency = Interpolate_func(theta_angle,10,collection_angle,collection_eff)/100.;
+     effectiveAngularEfficiency = fdet->GetPMTCE(theta_angle);
```
YujiOkajima: @aihimmel GetPMTCE is used in WCSimWCSD.cc.

cwwalter:
I think Alex's point is that the function GetPMTCE is not defined anywhere.  It looks like you may have forgotten to commit a file.

YujiOkajima:
I'm sorry, I forgot to add WCSimWCCE.cc
I have introduce it and confirmed it works correctly by setting "/WCSim/PMTCollEff on" and execute novis.mac .

aihimmel:
OK.  This all seems reasonable to me.  It looks like a lot of changes, but mostly it is the same change repeated for each PD.  It appears to be nearly the minimum-necessary change to include the new photosensors.  With the validation plots above, I think this is ready to merge.

cwwalter:
OK, I will merge this.  We want to restructure the way you did things to make getCE have a better name and be part of the base phototube class but we will do that as a separate pull request.

PR #60, opened 2015-03-27
Reorganizing tube-by-tube collection efficiency implementation.

aihimmel:
This should produce no change in behavior.  It simply moves around where things are defined so we are better prepared when we run with multiple tube types in the same detector.


src/WCSimPMTObject.cc
```diff
@@ -20,7 +20,46 @@
...
+
+////////////////////////////////////////////////////////////////////////////////////////////////
+// PMT Base Class
+
+WCSimPMTObject::WCSimPMTObject()
+{
+    collectionEfficiencyAngle = { 0., 10., 20., 30., 40., 50., 60., 70., 80., 90.};
+    collectionEfficiency      = { 100., 100., 100., 100., 100., 100., 100., 100., 100., 100.};
+}
+
```
aihimmel: Previously there was no constructor in the base class.  Now there is a constructor which is called first, setting the collection efficiency to 100%, which can then be overridden in the derived class constructor.  I've added a comment which explains this.

cwwalter:
Looks good.  In addition to removing those functions from the new file did you "rm" the now empty file also?  It looks like no but maybe I'm wrong..

aihimmel:
Yes, the file has been removed.  It still appears to be there for the sake of the diff, but once it is merged it should be gone.

cwwalter:
OK I will merge.

PR #61, opened 2015-03-27
Changing HPD to HQE box and line PMTs in the detector configurations.

aihimmel:
Changing HPD to HQE box and line PMTs in the detector configurations.

### DIFF
--- a/src/WCSimDetectorConfigs.cc
+++ b/src/WCSimDetectorConfigs.cc
@@ -65,9 +65,9 @@ void WCSimDetectorConstruction::SuperK_20inchPMT_20perCent()
 
 
 // Note: the actual coverage is 20.27%
-void WCSimDetectorConstruction::SuperK_20inchHPD_20perCent()
+void WCSimDetectorConstruction::SuperK_20inchBandL_20perCent()
 {
-  WCSimPMTObject * PMT = CreatePMTObject("HPD20inchHQE");
+  WCSimPMTObject * PMT = CreatePMTObject("BoxandLine20inchHQE");
   WCPMTName           = PMT->GetPMTName();
   WCPMTExposeHeight   = PMT->GetExposeHeight();
   WCPMTRadius         = PMT->GetRadius();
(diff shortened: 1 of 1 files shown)


cwwalter:
OK, this looks good.  I updated the table at 

https://wiki.hyperk.org/DesignReport/DetectorOptimization
To reflect this.  

The configuration names are now:
SuperK_20inchBandL_20perCent
SuperK_20inchBandL_14perCent
SuperK_12inchBandL_15perCent

YujiOkajima:
I tried to compile the WCSim, but it was failed with the error message below.

src/WCSimDetectorConfigs.cc:68: error: no ‘void WCSimDetectorConstruction::SuperK_20inchBandL_20perCent()’ member function declared in class ‘WCSimDetectorConstruction’
src/WCSimDetectorConfigs.cc:92: error: no ‘void WCSimDetectorConstruction::SuperK_12inchBandL_15perCent()’ member function declared in class ‘WCSimDetectorConstruction’
src/WCSimDetectorConfigs.cc:116: error: no ‘void WCSimDetectorConstruction::SuperK_20inchBandL_14perCent()’ member function declared in class ‘WCSimDetectorConstruction’

I think src/WCSimDetectorMessenger.cc and include/WCSimDetectorConstruction.hh are not updated.
Pleas check it.

aihimmel:
I have pushed a fix.  Sorry for the trouble.

YujiOkajima:
I tried to use the up-to-date version WCSim with different condition with different condition from usual, However it was failed to compile with the error message like the message below

src/WCSimPMTObject.cc:1357:78: warning: extended initializer lists only available with -std=c++11 or -std=gnu++11 [enabled by default]
     collectionEfficiency = { 95., 95., 95., 95., 95., 95., 95., 95., 95., 95.};   
                                                                              ^
src/WCSimPMTObject.cc:1357:26: error: assigning to an array from an initializer list
     collectionEfficiency = { 95., 95., 95., 95., 95., 95., 95., 95., 95., 95.};   

I also tried to compile the code I pulled request as #59 in the same environment, and succeeded to do it.

In this condition, The gcc version is 4.8.2. 
I confirmed it works correctly in the environment I usually use (gcc version 4.4.6).

Please check it.

aihimmel:
On the Duke machines (gcc version 4.4.7), I see a similar warning, but do not get the compiler error.  Using gcc 4.4.6 on sukap I see neither errors nor warnings.  I am currently trying to run with 4.8.2 on sukap to see if I can reproduce the problem.

aihimmel:
This is a feature of C++11, so I am surprised that the issue only shows up with the newer compiler and not the older compiler.

PR #62, opened 2015-03-30
Update header and messenger to match new configuration names.

aihimmel:


aihimmel:
Commit changes missing from pull request #61.  Will merge ASAP so work can continue.

PR #63, opened 2015-03-30
Change novis.mac and vis.mac with new detector configuration names

eosullivan:
Remove defunct detector configuration names and add new Box and Line SK configurations.


cwwalter:
Looks good.

PR #64, opened 2015-04-01
Update README.md to update the recommended version of gcc and to put the recommended version of ROOT.

eosullivan:
We are now all working with at least gcc 4.4, so I updated the recommended compiler version to the one we are currently using at Duke (gcc4.4.7). I also put which version of ROOT should be used, since WCSim will not work with all versions of ROOT.


eosullivan:
Just to be clear, this is a change only to the README file and doesn't affect the code.

PR #65, opened 2015-04-01
Fix gcc4.8.2 Issues

aihimmel:
Revert back to the static variable in a member function method of storing the collection efficiency.  Tested to compile with gcc4.4.6 and gcc4.8.2.

### DIFF
--- a/include/WCSimPMTObject.hh
+++ b/include/WCSimPMTObject.hh
@@ -12,8 +12,6 @@ class WCSimPMTObject
 {
 
 public:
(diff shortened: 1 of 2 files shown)


eosullivan:
This seems reasonable since it at least matches what we currently do to get the QE info.

aihimmel:
Exactly.  Let's put off the discussion of if there's a better way to structure how we store PMT information until a later date.

eosullivan:
Ran the verification script and it seems to be working fine. Merging.

Issue #38, opened 2015-01-20
Make it possible to store multiple PMT pointers. 


eosullivan:
This will be needed to complete issue #13 . This needs to be done with issue #37 in mind as the way we implement the readout of multiple PMTs will affect how the pointers should be stored.

eosullivan:
Closed with pull request #66.

PR #66, opened 2015-04-07
Feature/issue38

eosullivan:
## Motivation

This pull request contains code which allows for mulitple photodetectors to be built in the same detector, which is the first step in being able to implement a functional outer detector. It should be noted, however, that while you can build mutiple photodetectors in the same detector, there still exists only one electronics readout channel so only one photodetector type will be recorded and digitized. Furthermore, currently only a single sensitive detector volume can be specified. The ability to record hits from multiple photodetectors and to declare multiple sensitive detector volumes will be implemented in subsequent pull requests. This pull request is in response to issue #38.  

### Main code flow changes 

When a detector configuration is specified (in WCSimDetectorConfigs) the user instantiates a PMT using CreatePMT("PMTType", "CollectionName"). In the current version of the code the user would just specify the "PMTType", where PMTType is the type of photodetector (for example, 20inchPMT). After this pull request, the user must now also specify the "CollectionName" where CollectionName is the name of the hits collection (currently all ID tubes are set to glassFaceWCPMT). 

CreatePMT will instantiate a PMT of the specified type and will also store the PMT pointer in a map which is keyed with CollectionName. When information about the PMT is required (for example, in the digitizer when the timing resolution is needed) the PMT pointer can be retrieved using GetPMTPointer(CollectionName). Currently, all electronics are set up to find the glassFaceWCPMT CollectionName. The code will look up the CollectionName in the CollectionName array, will retrieve the index, and will call the corresponding PMTPointer from the PMTPointer array using that index. 

### Minor code changes

ConstructPMT now takes PMTName and CollectionName as inputs (before this pull request radius and expose was used). This allows the user to create an inner and outer detector with the same photodetector type, and also allows for the CollectionName to be passed in and used to create the proper hits collection name. 

Now that multiple PMT types can be properly stored in memory, the HyperK configuration in WCSimDetectorConfigs is changed to use the proper CreatePMT command for the OD tubes.

## Verification 

The verification_HitsChargeTime.C was run for both the default SK setup and for the HK setup. Both matched the clean checkout of the code. The HK detector was also run using vis.mac to ensure the new way of declaring the OD tubes looked ok by eye.

### DIFF
--- a/include/WCSimDetectorConstruction.hh
+++ b/include/WCSimDetectorConstruction.hh
@@ -92,25 +92,26 @@ public:
 
   WCSimPMTObject *CreatePMTObject(G4String, G4String);
 
-  WCSimPMTObject *  PMTptrID[10];
-  WCSimPMTObject * PMTptrOD[10];
-  int IDPMTCounter;
+  G4int total_array_number;
+  G4int array_index;
   int ODPMTCounter;
-  void SetPMTPointer(WCSimPMTObject* PMT, G4String PMTVolume) {
-    if(PMTVolume == "ID"){
-      PMTptrID[IDPMTCounter] = PMT;
-      IDPMTCounter++;
+  WCSimPMTObject * PMTptr[10];
+  G4String CollectionNameArray[10];
+ 
+ void SetPMTPointer(WCSimPMTObject* PMT, G4String CollectionName) {
+   CollectionNameArray[total_array_number] = CollectionName;
+   PMTptr[total_array_number] = PMT;
+   total_array_number++;
+ }
+  
+  WCSimPMTObject* GetPMTPointer(G4String CollectionName){
+    array_index = 1000; // some large number
+    for (int i=0; i<total_array_number; i++){
+      if (CollectionName == CollectionNameArray[i]){array_index = i; break;}
     }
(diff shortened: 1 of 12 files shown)


include/WCSimDetectorConstruction.hh
```diff
@@ -88,18 +88,35 @@ public:
...
-  WCSimPMTObject*  GetPMTPointer(){return PMTptr;}
+  WCSimPMTObject *CreatePMTObject(G4String, G4String);
 
+  G4int total_array_number;
+  G4int array_index;
+  int ODPMTCounter;
+  WCSimPMTObject * PMTptr[10];
+  G4String CollectionNameArray[10];
+ 
+ void SetPMTPointer(WCSimPMTObject* PMT, G4String CollectionName) {
```
aihimmel: Perhaps these functions should use the already existing maps?

eosullivan: Ok, I have updated the code to use a map (though, not an existing one). The map uses CollectionName as the key and stores/returns the PMTObject pointer.

cwwalter:
OK, this looks good after Alex's changes.  I'm merging this in.

PR #69, opened 2015-04-16
Now set the glass thickness parameter in ConstructPMT

eosullivan:
## Summary  

The WCPMTGlassThickness parameter was being hard set in WCSimDetectorConfigs, and wasn't being called in ConstructPMT. This meant that only a single glass thickness could be specified for a given detector configuration and therefore the glass thickness used in ConstructPMT didn't necessarily correspond to the PMT being built. This pull request adds in the call for glass thickness in ConstructPMT and takes out the line which hard sets the glass thickness in DetectorConfigs. 

## Verification  

Though this change is fairly minor, a number of verification steps were done to ensure the removal of the WCPMTGlassThickness from DetectorConfigs did not affect the code output.  

The only detector configurations that build multiple PMT types are the two Hyper-K geometries (SetHyperKGeometry and SetHyperKGeometry_withHPD). In the current version of WCSim, the glass thickness for the OD tubes would have been the same as the ID tubes, whereas after the changes proposed in this pull request the glass thickness of the OD tubes will match the glass thickness specified for the given PMT type. I ran a high statistics run for both these geometries and found that the new glass thickness in the OD tubes did change the number of hits on an event-by-event basis, but that these changes are statistically consistant. 

###  SetHyperKGeometry  
glassthickness_hkwpmt  

###  SetHyperKGeometry_withHPD  
glassthickness_hkwhpd

All other detector configurations were tested by running a single event and comparing the output with the current version of WCSim. As expected, the outputs from the two versions of WCSim matched in all cases.

### DIFF
--- a/src/WCSimConstructPMT.cc
+++ b/src/WCSimConstructPMT.cc
@@ -38,12 +38,12 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructPMT(G4String PMTName, G4Str
   
   G4double expose;
   G4double radius;
-  G4double glassthickness;
+  G4double glassThickness;
   
   WCSimPMTObject *PMT = GetPMTPointer(CollectionName);
   expose = PMT->GetExposeHeight();
   radius = PMT->GetRadius();
-  glassthickness = PMT->GetPMTGlassThickness();
+  glassThickness = PMT->GetPMTGlassThickness();
 
   G4double sphereRadius = (expose*expose+ radius*radius)/(2*expose);
   G4double PMTOffset =  sphereRadius - expose;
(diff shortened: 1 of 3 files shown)


src/WCSimConstructPMT.cc
```diff
@@ -38,11 +38,13 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructPMT(G4String PMTName, G4Str
...
   G4double expose;
   G4double radius;
+  G4double glassthickness;
   
   WCSimPMTObject *PMT = GetPMTPointer(CollectionName);
   expose = PMT->GetExposeHeight();
   radius = PMT->GetRadius();
-  
+  glassthickness = PMT->GetPMTGlassThickness();
+
```
cwwalter: It's a minor thing.  But, let's try to stick to our capitalization convention and make this glassThickness.

cwwalter:
Should 

G4double WCPMTGlassThickness;

in the .hh file also be removed?  Any of those other variables?

cwwalter:
OK. This looks good now.

PR #70, opened 2015-04-24
Modify sample macro code to use the collection efficiency parameter.

YujiOkajima:
In the current sample macro files, the collection efficiency process is disabled.
In order to enable collection efficiency process, I fixed novis.mac and vis.mac.
Because R3600 has 100% collection efficiency, the simulation result is not changed.
The figure below shows the comparison between the result of simulation with novis.mac
before and after modification. (Red: before modification, Blue: after modification)
novis_compare

### DIFF
--- a/novis.mac
+++ b/novis.mac
@@ -63,6 +63,9 @@
 #/WCSim/PMTQEMethod     Stacking_And_SensitiveDetector
 #/WCSim/PMTQEMethod     SensitiveDetector_Only
 
+turn on or off the collection efficiency (05/27/11 XQ)
+#/WCSim/PMTCollEff on
+
 # command to choose save or not save the pi0 info 07/03/10 (XQ)
 /WCSim/SavePi0 false
 
--- a/novis.mac
+++ b/novis.mac
@@ -63,7 +63,7 @@
 #/WCSim/PMTQEMethod     Stacking_And_SensitiveDetector
 #/WCSim/PMTQEMethod     SensitiveDetector_Only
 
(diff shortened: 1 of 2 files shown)


eosullivan:
I think this was left commented out purposely so that you have make an effort to turn it on. I don't have any strong opinions about it though. It could almost be argued that we shouldn't enable it from the macro at all and that it should just be properly set in the definition of the photodetector.

cwwalter:
Yes, this was done on purpose.  Until we make the use of angular dependent collection efficiency the default for our running I would like to keep things this way.  Some of the tubes from the LBNE studies (like 8" tubes) have non flat responses.  So, if we turn this on by default results might change for people even though they are not expecting it.

YujiOkajima:
OK, then I will stop changing the collection efficiency to enable.
But I think novis.mac should be modify because the macro file does not mention about this command and this command will be necessary for Box & Line PMT study.
I think 
#/WCSim/PMTCollEff on
should be added in novis.mac.

cwwalter:
Yes, that makes sense.  I agree.  Can you modify this pull request?

YujiOkajima:
I have reverted vis.mac and novis.mac, and modified novis.mac after that.

cwwalter:
Thanks.  Can you also remove the "(05/27/11 XQ)"  from both files?  It is left over from LBNE development.

YujiOkajima:
I removed it.

cwwalter:
OK, this looks good.

PR #72, opened 2015-05-06
Feature/macro cleanup

eosullivan:
Just a general cleanup of the vis.mac and novis.mac files. The only changes in this pull request are the addition/deletion/rearrangement of some comments which describe the functions in the macros. The main motivation for this pull request is the addition of a comment on the /WCSim/Construct line that reminds users to uncomment this out if they construct a new detector configuration, as this seems to be a common mistake. I also removed the section relating to the WCPMTsize command as this has been removed for some years now. Finally, I fixed the comments related to the /WCSim/HK/waterTankLength line and moved the comment describing the usage of this function to directly above the command lines.

### DIFF
--- a/novis.mac
+++ b/novis.mac
@@ -21,16 +21,11 @@
 # diameter of 69m with 12" HPD and 14.59% photocoverage
 #/WCSim/WCgeom Cylinder_12inchHPD_15perCent # Note: the actual coverage is 14.59%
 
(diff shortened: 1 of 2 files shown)


eosullivan:
Although these changes are all related to comments, I did do a quick validation where I ran vis.mac and novis.mac to ensure they still work.

cwwalter:
Looks good thanks.

Issue #73, opened 2015-05-19
Incorrect hit timing resolution


YujiOkajima:
I'm sorry,

I found the function which I used for hit timing resolution analysis of measured data is different from the function which has been already introduced. It will effect in high charge region.

I will fix it as soon as possible.

fdilodov:
Hi, just to understand this issue a bit better. There is a 10-12% difference between what we should have had and what we actually have in the code. Do we have an idea of how much this going to affect the reliability of our results using the SK framework? For the HK tank production I think we can from now work in using the right code (ie fixing WCSim and tuning fiTQun) so the problem is just for SK. I presume that the resolutions we will get will not be correct, but the actual relative comparisons between the different options may still be OK? Can we discuss it at tomorrow's sw meeting?

YujiOkajima:
In low energy study, it may not affect so much, however when the charge of PMT > 23.96 p.e. , the timing resolution becomes worse.
I'm going to describe this problem at the tommorow meeting.

fdilodov:
Thanks.
PS it seems that my comment didn't appear by email but just on the web, 
it is attached below.


On 20/05/15 13:42, YujiOkajima wrote:
>
> In low energy study, it may not affect so much, however when the 
> charge of PMT > 23.96 p.e. , the timing resolution becomes worse.
> I'm going to describe this problem at the tommorow meeting.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/WCSim/WCSim/issues/73#issuecomment-103869905>.
>
Hi, just to understand this issue a bit better. There is a 10-12% 
difference between what we should have had and what we actually have in 
the code. Do we have an idea of how much this going to affect the 
reliability of our results using the SK framework? For the HK tank 
production I think we can from now work in using the right code (ie 
fixing WCSim and tuning fiTQun) so the problem is just for SK. I presume 
that the resolutions we will get will not be correct, but the actual 
relative comparisons between the different options may still be OK? Can 
we discuss it at tomorrow's sw meeting?

cwwalter:
Hi Francesca, I got it in email.  You don't get it yourself.  But other people get it.

cwwalter:
BTW, I also want to know how this mistake happened given the checks that were done before.  We waited for many checks to make sure it was correct.  But, it was still wrong.  So, we should understand how it happened so we don't do it again.

eosullivan:
This was corrected in pull request #74

PR #74, opened 2015-05-19
Fix the timing resolution of Box and Line PMT

YujiOkajima:
This is modification for #73 issue

I analyzed the timing resolution of Box and Line PMT and fitted with correct function.
And introduced the fitted curve in WCSim.

I checked the 1 p.e. timing resolution by comparing PDs hit time and true hit time with that of nominal PD, and the MC output seems to be correspond to timing resolution calculated with input parameters.
(nominal PD is changed event by event)
(timing resolution, calculated value: 2.30 ns, MC value: 2.30 +/- 0.01 ns)
So, new code seems to work correctly.

boxandline_checktts_modified
(timing resolution is sqrt(((p2)^2+2*(p3)^(-2))/2) )

### DIFF
--- a/src/WCSimPMTObject.cc
+++ b/src/WCSimPMTObject.cc
@@ -1191,7 +1191,7 @@ G4double BoxandLine20inchHQE::GetRadius() {return .254*m;}
 G4double BoxandLine20inchHQE::GetPMTGlassThickness() {return 0.4*cm;}
 
 float BoxandLine20inchHQE::HitTimeSmearing(float Q) {
-  G4float sig_param[4]={0.6315,0.06277,0.5728,23.9};
+  G4float sig_param[4]={0.6314,0.06260,0.5711,23.96};
   G4float lambda_param[2]={0.4094,0.06852};
   G4float sigma_lowcharge = sig_param[0]*(exp(-sig_param[1]*Q)+sig_param[2]);
 
@@ -1360,7 +1360,7 @@ G4double BoxandLine12inchHQE::GetRadius() {return 152.4*mm;}
 G4double BoxandLine12inchHQE::GetPMTGlassThickness() {return 0.4*cm;}
 
 float BoxandLine12inchHQE::HitTimeSmearing(float Q) {
-  G4float sig_param[4]={0.6315,0.06277,0.5728,23.9};
+  G4float sig_param[4]={0.6314,0.06260,0.5711,23.96};
   G4float lambda_param[2]={0.4094,0.06852};
 
   G4float highcharge_param[2];
--- a/src/WCSimPMTObject.cc
+++ b/src/WCSimPMTObject.cc
@@ -1191,10 +1191,16 @@ G4double BoxandLine20inchHQE::GetRadius() {return .254*m;}
 G4double BoxandLine20inchHQE::GetPMTGlassThickness() {return 0.4*cm;}
 
 float BoxandLine20inchHQE::HitTimeSmearing(float Q) {
-  G4float sig_param[4]={0.8928,0.03278,0.07054,16.4};
+  G4float sig_param[4]={0.6315,0.06277,0.5728,23.9};
   G4float lambda_param[2]={0.4094,0.06852};
+  G4float sigma_lowcharge = sig_param[0]*(exp(-sig_param[1]*Q)+sig_param[2]);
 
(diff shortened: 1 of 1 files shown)


YujiOkajima:
Because I fail to save the fitting result, I analyzed the sigma and get fitting parameters again.
And I validated the transit time spread again.

hittime_resolution
(timing resolution, calculated value: 2.30 ns, MC value: 2.28 +/- 0.01 ns)

eosullivan:
Can you show us a plot with the old and new timing distributions so we can see how things will change?

YujiOkajima:
@eosullivan 

In low charge region, there is not so much difference between old code and new code.
the difference is shown in high charge region and I don't have the method to analyze TTS of high charge.
So I will describe with calculated values
TTS of Box and Line PMT follows exponentially modified Gaussian and the timing resolution consists of sigma and lambda. (sigma: deviation of Gaussian component, lambda: parameter of exponential component)
And the value of sigma is not correct.
The below figure shows the relationship between sigma and charge before and after modification.
(Red: old code, Blue: new code)
compare_resolution_highq

Because the fitted curve for analysis and implementation was different, the timing resolution before modification is much worse than that after modification in >20 p.e. region.

The below figure shows the relationship between "(sigma_old)/(sigma_new)-1" and charge.
In >20 p.e. region, the timing resolution has >10% difference.

compare_resolution_difference_highq

YujiOkajima:
I have uploaded slide about this problem
https://dl.dropboxusercontent.com/u/61208795/HK_meeting_2015_May_21.pdf

eosullivan:
Your sigma_correct for Q<3 is different between slide 3 and 4. Is this just some typo or is it a mistake in your code? Should it be e^{p1} rather than e^{p0} in slide 3?

eosullivan:
Also, I just looked at the Q>3 sigma function in your pull request and it does not match the sigma_correct on slide 4. The function in your pull request gives a leading factor of 2p0p1p3^3/2 rather than the leading factor in your slides of 2p0p2p3^3/2 (ie. in your slides, p2 is in the leading factor and in your code p1 is in the leading factor). Can you double check which is right? Also, the way you have put it into the code makes it a bit hard to translate to your slides.

YujiOkajima:
@eosullivan 

I'm sorry, There are some typos in slides

the leading factor of the function in slide 2 is 2p0p1p3^3/2, and all analysis is done by using function with 2p0p1p3^3/2 leading factor.
The functions in the slides are wrong but all figures in the slides are correct
the sigma_correct had typo. It is e^{p1} in slide 3

eosullivan:
Okajima-san, is this pull request all up-to-date and ready to be merged?

YujiOkajima:
Yes, the timing resolution was fixed to correct function and analyzed the measured data with the function.
And I validated the timing resolution and sigma.
sigma_bl_mc
I will also describe the validation result about timing resolution of box and line PMT later.

eosullivan:
Ok, so perhaps we will wait until we have discussed this validation result before merging in the code. When you say you will describe the validation result later, do you mean at the next software meeting?

YujiOkajima:
I will make slide about these validation and upload it on this page.
Because all analysis has been already finished, maybe I will upload it tomorrow.

YujiOkajima:
I made the slide to describe the validation result of timing resolution
https://dl.dropboxusercontent.com/u/61208795/Description_hittime_validation.pdf
From this result, the parameters seems to be introduced correctly.

By the way, as I described in this slide, timing resolution of Q<0.5 p.e. signal is set to that of Q=0.5 p.e. in the simulation process of WCSim.
Because the timing resolutions of Box and Line PMT and HPD are not diverge at 0p.e., 
I think this process should move to WCSimPMTObject.cc.
If my idea is correct, I will make this issue.

eosullivan:
So, is this Q=0.5 pe a property of only the PMTs? Can you check with an expert to see if this is correct? It does seem to not be fitting the low Q region very well. 

As for taking out the trigger time and replacing it with first hit time in the digitizer code, is this what the digitizer actually does? We want the code to do what the actual hardware does.

cwwalter:
Let's have the discussion about 0.5 etc somewhere else (another issue?).  If necessary.  So, we can keep this issued focused on the B&L timing function fix.  If this is all ready let's close this and merge it.

YujiOkajima:
@eosullivan 
The Q=0.5 p.e. property is currently also applied to HPD, therefore HPD also seems not to be fitted very well.
I will make this issue later.

This code modification is not included in this pull request.
This is the code modification for only validation.

PR #75, opened 2015-05-24
Implement new HPD parameters

YujiOkajima:
The new HPD parameters based on 20 inch HPD (Prototype version) measurement result are introduced in WCSim and validated.

I uploaded the validation result.
https://dl.dropboxusercontent.com/u/61208795/newHPD_implementation.pdf
From this result, 1 p.e. distribution, Conversion factor of dark noise, timing resolution about 1p.e. region seems to work correctly.  (QE is the same value as Box and Line PMTs)
Now I have been considering how to validate high charge region timing resolution, but I do not have the method at the moment.

And, based on the geometry of detectors with Box and Line PMT, new detectors with new HPDs are added.
I execute simulation in those geometry and there is no error in the process.


eosullivan:
Why do we need two different HPDs in WCSim? Is this just updated information about HPDs (in which case we should just replace the old info in the HPD entry we already have) or is this actually a new PD? Is there any reason to keep the old HPD around?

YujiOkajima:
@eosullivan 
The data source of old and new HPD parameters is absolutely different (measurement result of 8" HPD in water and 20" prototype HPD in the air) except for QE and PD shape (QE is based on High QE R3600). 
I think we should use new parameters for the evaluation of photosensor performance in detector.
However I also thought the condition of existing geometry should not be changed.
So I did not delete the parameters.

eosullivan:
Ok, so is it that the new parameters better describe the HPD we are considering for HK? If so, I see no reason to keep the old HPD around.

YujiOkajima:
Yes, I think the new parameter seems to be described better. 
Then, I will modify it.

cwwalter:
If that is the case, I agree. I think you should just change the current HPD parameter description and not add any new configurations.

eosullivan:
Also, is there any motivation for making all of these detector configurations with this HPD? We made many SK detector configurations with B&L for the design report study, but we will probably not be doing this with HPDs. Probably just one configuration, SK with 40% HPD coverage, would suffice.

eosullivan:
Sorry, I'll revise that last suggestion. Maybe we just need SK with 20" HPD and 20% coverage and SK with 12" HPD and 15% coverage.

YujiOkajima:
@eosullivan 
There is not so big reason. I thought the HPD should be done the same evaluation as that of Box & Line PMT to compare the result.
Then I will remove the 20" HPD 14% coverage geometry.

vis.mac
```diff
@@ -75,7 +78,8 @@
 # (14/03/10)
 # Factor of Normal PMT is 1.367
 /DarkRate/SetConvert 1.367
-#/DarkRate/SetConvert 1.120 #For HPDs
+#/DarkRate/SetConvert 1.120 #For HPDs (Detector: Cylinder_12inchHPD_15perCent and HyperK_withHPD)
+#/DarkRate/SetConvert 1.082 #For HPD_20150522 (Detector: SuperK_20inchHPD_20perCent, SuperK_20inchHPD_14perCent and SuperK_12inchHPD_15perCent)
```
eosullivan: How is this conversion rate calculated? This number always seemed a bit mysterious to me.

eosullivan:
The timing response for the HPDs follow the same complicated formula that you implemented for the B&Ls. Can you show us some validation plots which compare data with your function?

YujiOkajima:
@eosullivan 

Please read the slide I uploaded.
https://dl.dropboxusercontent.com/u/61208795/newHPD_implementation.pdf
From page.7, when the conversion factor is set to 1, the ratio of dark rate MC output to setting value is 0.9238.
In order to set the ratio to 1, I set the conversion factor to 1/0.9238=1.082.

And I validated the timing resolution by changing the code today.
sigma_hpd_mc_compare
Each point shows the timing resolution of MC output and the blue line shows the setting function.
From this result, the sigma of HPD seems to be correctly introduced (Box and Line PMT was also validated).
I will describe the validation result of timing resolution in detail later.

cwwalter:
I think we should not add any configuration we are not actually using now.  I think there are already too many!

Let's just add the ones we need when we use them.  It is also good to have that in its own pull request not in the pull request of the HPD parameter setting change.

YujiOkajima:
All right. I will delete the old HPD parameters.

eosullivan:
And please rename the new HPD to not include the date. Also, I will agree with Chris that there are many detector configurations and we don't necessarily need to add these new ones. So, please remove the new detector configurations. If we decide that we need them, they can be a separate pull request.

PR #78, opened 2015-06-01
Feature/issue67

eosullivan:
This commit allows for multiple sensitive detectors to be instantiated and to have unique hit collections. This is a necessary step towards our goal of having a working outer detector volume. 

## Modified file descriptions

WCSimDetectorConstruction.hh: Now contains a method to recall a hit collection name. Currently this is just whatever the user has specified as the ID collection.  

WCSimConstructCylinder.cc: Now calls ConstructPMT with the ID collection name rather than a hard coded "glassFaceWCPMT" collection name.  

WCSimConstructGeometryTables.cc: Will now assign a PMTID to tubes which belong to the ID or the OD hit collections.  

WCSimConstructHyperK.cc: Now calls ConstructPMT once to build an ID tube and once to build an OD tube. Tubes are built with the corresponding hit collections which have been specified in WCSimDetectorConfigs.cc.

WCSimConstructPMT.cc: The logical volume for the sensitive detector now matches the hit collection name rather than being the hard coded "glassFaceWCPMT". Multiple sensitve detectors can now be instatiated. 

WCSimDetectorConfigs.cc: Each configuration now has a detector name associated with it. This is to make the hit collection names distinct for the given configuration and to avoid the situation where we repeat a hit collection name when we rebuild (for example, if we build a glassFaceWCPMT in the default SK configuration, and then build glassFaceWCPMT in HK using the macro). 

WCSimEventAction.cc: Gets the hit collection associated with the ID for use by the digitizer electronics. This can be extended to be used with the OD (or with multiple hit collections) in the future. 

WCSimStackingAction.cc: Does the pruning for the quatum efficiency using the tube associated with the ID. For the OD to be properly implemented, we will need to think of a clever way to account for the different tubes in the configuration and choose an appropriate way to prune for the QE. For now, the only configuration this will affect is Hyper-K (since it is the only one with an OD) and there will be no change in performance since the QE correction was done for the ID collection previous to this pull request. 

WCSimWCDigitizer.cc and WCSimWCPMT.cc: Still gets the pointer for the ID tubes, but now finds the appropriate hit collection name for those tubes, rather than hard coding it in. 

WCSimWCSD.cc: Now applies the appropriate QE if method 3 (sensitive detector only) is applied. For method 2 (stacking action and sensitive detector) is applied, then the maxPe of the ID tube (which was used in stacking action) is used in the QE scaling. When filling the hit collection, the routine now must find which volume was hit and find the appropriate hit collection pointer.

### DIFF
--- a/src/WCSimConstructPMT.cc
+++ b/src/WCSimConstructPMT.cc
@@ -124,21 +124,21 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructPMT(G4String PMTName, G4Str
                        0.0*deg,90.0*deg);
   
   G4SubtractionSolid* solidGlassFaceWCPMT =
-      new G4SubtractionSolid(    "glassFaceWCPMT",
+      new G4SubtractionSolid(    CollectionName,
                                  tmpGlassFaceWCPMT,
                                  solidCutOffTubs); 
 
   G4LogicalVolume *logicGlassFaceWCPMT =
     new G4LogicalVolume(    solidGlassFaceWCPMT,
                             G4Material::GetMaterial("Glass"),
-                            "glassFaceWCPMT",
+                            CollectionName,
                             0,0,0);
 
   G4VPhysicalVolume* physiGlassFaceWCPMT =
       new G4PVPlacement(0,
                         G4ThreeVector(0, 0, -1.0*PMTOffset),
                         logicGlassFaceWCPMT,
-                        "glassFaceWCPMT",
+                        CollectionName,
                         logicWCPMT,
                         false,
                         0,
@@ -146,15 +146,15 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructPMT(G4String PMTName, G4Str
 
   //logicGlassFaceWCPMT->SetVisAttributes(G4VisAttributes::Invisible);
   logicGlassFaceWCPMT->SetVisAttributes(WCPMTVisAtt);
-
-  if (!aWCPMT)
-  {
+  
+  //  if (!aWCPMT)
+  //  {
     G4SDManager* SDman = G4SDManager::GetSDMpointer();
     G4String SDName = "/WCSim/";
     SDName += CollectionName;
     aWCPMT = new WCSimWCSD(CollectionName,SDName,this );
     SDman->AddNewDetector( aWCPMT );
-  }
+    // }
   logicGlassFaceWCPMT->SetSensitiveDetector( aWCPMT );
 
   PMTLogicalVolumes[key] = logicWCPMT;
--- a/src/WCSimConstructPMT.cc
+++ b/src/WCSimConstructPMT.cc
@@ -147,6 +147,7 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructPMT(G4String PMTName, G4Str
   //logicGlassFaceWCPMT->SetVisAttributes(G4VisAttributes::Invisible);
   logicGlassFaceWCPMT->SetVisAttributes(WCPMTVisAtt);
   
+  // Instantiate a new sensitive detector and register this sensitive detector volume with the SD Manager. 
     G4SDManager* SDman = G4SDManager::GetSDMpointer();
     G4String SDName = "/WCSim/";
     SDName += CollectionName;
--- a/src/WCSimConstructPMT.cc
+++ b/src/WCSimConstructPMT.cc
@@ -147,14 +147,12 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructPMT(G4String PMTName, G4Str
   //logicGlassFaceWCPMT->SetVisAttributes(G4VisAttributes::Invisible);
   logicGlassFaceWCPMT->SetVisAttributes(WCPMTVisAtt);
   
-  //  if (!aWCPMT)
-  //  {
     G4SDManager* SDman = G4SDManager::GetSDMpointer();
     G4String SDName = "/WCSim/";
     SDName += CollectionName;
     aWCPMT = new WCSimWCSD(CollectionName,SDName,this );
     SDman->AddNewDetector( aWCPMT );
-    // }
+
   logicGlassFaceWCPMT->SetSensitiveDetector( aWCPMT );
 
   PMTLogicalVolumes[key] = logicWCPMT;
--- a/src/WCSimConstructPMT.cc
+++ b/src/WCSimConstructPMT.cc
@@ -146,14 +146,14 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructPMT(G4String PMTName, G4Str
 
   //logicGlassFaceWCPMT->SetVisAttributes(G4VisAttributes::Invisible);
   logicGlassFaceWCPMT->SetVisAttributes(WCPMTVisAtt);
-  
+ 
   // Instantiate a new sensitive detector and register this sensitive detector volume with the SD Manager. 
-    G4SDManager* SDman = G4SDManager::GetSDMpointer();
-    G4String SDName = "/WCSim/";
-    SDName += CollectionName;
-    aWCPMT = new WCSimWCSD(CollectionName,SDName,this );
-    SDman->AddNewDetector( aWCPMT );
-
+  G4SDManager* SDman = G4SDManager::GetSDMpointer();
+  G4String SDName = "/WCSim/";
+  SDName += CollectionName;
+  aWCPMT = new WCSimWCSD(CollectionName,SDName,this );
+  SDman->AddNewDetector( aWCPMT );
+  
   logicGlassFaceWCPMT->SetSensitiveDetector( aWCPMT );
 
   PMTLogicalVolumes[key] = logicWCPMT;
--- a/src/WCSimConstructPMT.cc
+++ b/src/WCSimConstructPMT.cc
@@ -151,7 +151,7 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructPMT(G4String PMTName, G4Str
   G4SDManager* SDman = G4SDManager::GetSDMpointer();
   G4String SDName = "/WCSim/";
   SDName += CollectionName;
-  aWCPMT = new WCSimWCSD(CollectionName,SDName,this );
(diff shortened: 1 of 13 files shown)


cwwalter:
@aihimmel Do you have time to help review this now?

src/WCSimConstructPMT.cc
```diff
@@ -146,15 +146,14 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructPMT(G4String PMTName, G4Str
...
-    SDName += CollectionName;
-    aWCPMT = new WCSimWCSD(CollectionName,SDName,this );
-    SDman->AddNewDetector( aWCPMT );
-  }
+ 
+  // Instantiate a new sensitive detector and register this sensitive detector volume with the SD Manager. 
+  G4SDManager* SDman = G4SDManager::GetSDMpointer();
+  G4String SDName = "/WCSim/";
+  SDName += CollectionName;
+  aWCPMT = new WCSimWCSD(CollectionName,SDName,this );
```
aihimmel: It looks like aWCPMT is now being overwritten each time a new PMT is created. Will this cause a problem? Probably not if aWCPMT is not used anywhere else. If that's the case, we should move it from a member variable to a local variable in this function. If it turns out we do need to use aWCPMT, we probably need to make it a map keyed to the CollectionName.

src/WCSimDetectorConfigs.cc
```diff
@@ -20,7 +20,10 @@
 
 void WCSimDetectorConstruction::SetSuperKGeometry()
 {
-  WCSimPMTObject * PMT = CreatePMTObject("PMT20inch", "glassFaceWCPMT");
+  WCDetectorName = "SuperK";
```
aihimmel: Not required, but this might be a nice thing to store as a branch in the geometry tree. That way you can quickly determine what geometry was used without needing to look at the PMT arrangement or rely on the filename. Could be especially handy when evaluating multiple types of geometry.

src/WCSimWCSD.cc
```diff
@@ -163,11 +164,16 @@ G4bool WCSimWCSD::ProcessHits(G4Step* aStep, G4TouchableHistory*)
      G4double local_y = localPosition.y();
      G4double local_z = localPosition.z();
      theta_angle = acos(fabs(local_z)/sqrt(pow(local_x,2)+pow(local_y,2)+pow(local_z,2)))/3.1415926*180.;
-     effectiveAngularEfficiency = fdet->GetPMTCollectionEfficiency(theta_angle, collectionName[0]);
+     effectiveAngularEfficiency = fdet->GetPMTCollectionEfficiency(theta_angle, volumeName);
      if (G4UniformRand() <= effectiveAngularEfficiency || fdet->UsePMT_Coll_Eff()==0){
-
+       //Retrieve the pointer to the appropriate hit collection. Since volumeName is the same as the SD name, this works.
```
aihimmel: Is this a G4 requirement, or is this just a convention of our code? If it is just a convention, perhaps we need to do something more robust, or at least put a comment where the SD and volume get named that makes it clear that this is required. It actually looks like the collection name, SD name, and volume name are all the same. Is that the case?

eosullivan: Yes, currently collection name, SD name and volume name are all the same. The collection name and volume name are the same so that you can find which volume you are in and easily assign it to the proper collection. Also, when the WCSimGeometryTable code assigns PMT numbers, you want it to do it for both the ID and OD PMT volumes. It is possible, in principle, for volume name and collection name to be different, but you would then need to keep track of both and have the user define both in the DetectorConfigs. For the sensitive detector, I just wanted it to be distinct for the different volumes and making it equal to the hits collection achieves this.

src/WCSimConstructPMT.cc
eosullivan: Ok, it doesn't seem like aWCPMT is used anywhere and so I have made it a local variable.

src/WCSimDetectorConfigs.cc
eosullivan: Yes, I like this idea. I could do it now or wait until the next step where the digitizer is able to read out from various hit collections. Then I could put the hit collection which is recorded by the digitizer (which will also contain the detector name) into the geometry tree.

aihimmel: That makes sense. Better to make all the changes to the output trees at once.

aihimmel:
@cwwalter, this all looks good to me. Once aWCPMT is made local I think this is ready to merge.

eosullivan:
Great, thanks for checking everything over @aihimmel. I am just running a few final validation tests. I will post them here once they are complete.

cwwalter:
Nice job on this PR and the review @eosullivan and @aihimmel.  Thanks!

The only thing I didn't follow is the aWCPMT discussion.  I didn't have time to follow up in detail but it looks like we are making a new SD to add to the SD manager.  There are two questions I have related to this and making it local:

- Is the name aWCPMT good?  I mean is it really a PMT?  I think it is a real a SD part of a PMT right?  I know this is old code I probably wrote BTW :)

- One thing I want to check on: if you make this variable local then it will go out of scope when the function exits. Since you did a 'new' the memory should still exist but you will never-ever-ever be able to access it again. Is that OK? Or can you still get at it through the SDMan?

cwwalter:
BTW, I suspect the aWCPMT was done that way because it is a "standard" G4 approach.  You might want to peak at the examples (for example). Is this done once for each time a new kind of PMT is defined?  If this does cause some sort of problem another solution might be to make a class variable which is a list of pointers.

aihimmel:
I don't think we need to explicitly keep track of the pointer. It's not called anywhere else in the code. Both the SDManager and logicGlassFaceWCPMT have the pointer and it seems like the actual sensitive detector object is only used under the hood. If this does break things it should show up pretty obviously when Erin runs her validation.

aihimmel:
I just checked the examples in N02 the sensitive detector is a local variable in the Construct() function, much like what we do here.

cwwalter:
Great thanks.

eosullivan:
I don't think the name aWCPMT is too off the mark. If we wanted to be super technical about it then it is the sensitive part of the PMT that we are registering, but I think it's fine to think of it conceptually as just registering a PMT to be a sensitive detector. As far as the variable disappearing from memory goes, the SD manager seems to be able to find things ok.

eosullivan:
For verification, I began by running a single event in the SK and cylinder detector configurations and found that there was no change between the current WCSim and the version in this pull request (as expected since there is no OD in these configurations). I then ran 500 MeV electrons with no dark noise in HK and found that there were event-by-event differences, but things seem statistically significant (plot attached below). Finally, I ran 5000 MeV mu- events with no dark noise in HK. In this case, light is reaching the OD and so, as expected, we see (slightly) more hits in the current WCSim version than the pull request version. This is as expected since in the current version OD tubes are being read out with the ID tubes, whereas in the pull request version the OD tubes are in their own hit collection and are not being recorded by the digitizer. This plot is also attached below. 

As a sanity check, I verbosely read out the PMTID of the tubes being hit. Using the geofile.txt, I made sure that when the hit is registered in the SD name I associated with the OD volume , it matched the tubes that are outward-pointing and that the hits associated with the ID volume were using the inward-pointing tubes. I then made sure that it was using the correct pointer address for the ID/OD hit collection as appropriate and that the PMTHitMap was being filled in a sensible way with the multiple hit collections. 

## HyperK with 500 meV e-

multisd_hk_500mev_e-

## HyperK with 5 GeV mu-

multisd_hk_5000mev_mu-

cwwalter:
Great.  But, I'm confused by something.  In the 2nd set of plots I don't see that they are systematically lower as you write in the text.

BTW, can you make a ID and OD set of plots for HK.  They should add to the total right?  That is sort of your long term goal anyway.

eosullivan:
In the second set of plots (HyperK with 5 GeV mu-) the # of digitized hits is systematically lower for the red curve (the red is always higher in the rising edge of the peak, has the max bin in a lower hit bin, always lower on the falling edge of the peak). But, you are right that it isn't so obvious. Because these are high energy muons (which I chose since they will penetrate into the OD), they produce lots of light and so the OD contribution is small compared with the ID contribution (like, tens of hits in the OD vs thousands in the ID). I'll see what I can do about making ID and OD plots separately to make things a bit clearer.

cwwalter:
I just realized after reading the comments there is something I haven't thought through/understand.  Now you have separate hit collections for the ID and the OD.  Which is great. But what get's written out to the the ROOT file and are there two digitized collections written out or still just one?  If one, does that one include both the OD and ID or just the ID?

I was thinking it was trivial to plot the OD hits because it was in a separate collection but now I realize that may not be the case.

eosullivan:
There is still exists just one digitized collection and that still gets written to the ROOT file. Before this pull request, the digitzed collection used the hit collection which contained both ID and OD hits. After this pull request, the digitized collection will use the hit collection which contains only ID hits. So, the change here is only in the hit collection. The last step to all this is changing the digitzer to have different digitzed collections, and that will be in a separate PR.

It is, in principle, trivial to plot the digitized collection based on the OD hit collection. The only issue is that the trigger for the digitizer is 25 hits (which I could turn off) and the fact that I need to rerun everything to get the OD HC digitzed since the simulation and digitization isn't separate in the code. It takes ~ a day to run this. What are we trying to see here? I think seeing that it sums to the total is a bit hard to do on a visual level. Perhaps just running a single event and showing that when you digitize the pre-PR WCSim hit collection (ID+OD) that it matches the case where you digitize the post-PR WCSim hit collection (ID) + the post-PR WCSim where you swap the HC to be the one which corresponds to the OD hit collection. Does that make sense?

eosullivan:
Ok, actually trying to get the exact number when you compare pre and summed post-PR WCSims won't work because the presence of the additional SD changes the random numbers and gives rise to differences.

cwwalter:
Sorry I didn't get to this yesterday.  I was just suggesting the separate ID OD plots as a good way to confirm and display to others it was performing as expected.  But, I missed that this PR alone was not enough to do that.

If you (@eosullivan) have convinced yourself using the other checks that the changes are working properly with no side effects, go ahead and merge it now.

Issue #82, opened 2015-06-10
More precise analysis of HPD parameters


YujiOkajima:
The introduced timing resolution of HPD, which was calculated from sigma and lambda of Exponentially modified Gaussian, seems to be different from that timing resolution analyzed from the measurement result in the low charge region. 
I changed the analysis method (make the number of qbin twice as big as past analysis by changing the qbins width, change the fitting target from lambda to 1/lambda) , and analyzed HPD timing resolution again.

1/lambda (new analysis, fitted by 1/(pol1))
![hpd_invert_lambda2](https://cloud.githubusercontent.com/assets/8589041/8082186/4c3d69d8-0fb6-11e5-907e-3eb0d1fe8e39.png)

sigma (new analysis)
![sigma_hpd_test](https://cloud.githubusercontent.com/assets/8589041/8082649/c03ceb76-0fb9-11e5-97ce-f7b094fde642.png)

Timing resolution (sqrt(sigma^2+1/lambda^2), blue: old parameters, red: new parameters)
![hpd_timing_resolution_compare](https://cloud.githubusercontent.com/assets/8589041/8098481/a94885b6-102b-11e5-99d2-c2c1da8696e9.png)

On the other hand, Box and Line PMT seems not to be such a big difference between data and fitted function.
The below figure shows the timing resolution of Box and Line PMT and function which has been introduced in WCSim.
![boxandline_timingresolution](https://cloud.githubusercontent.com/assets/8589041/8082372/c00c60c0-0fb7-11e5-9fc3-d17f432c5f11.png)

YujiOkajima:
As a reference, the timing resolution of Box & Line PMT was analyzed by using the result of 1/lambda fitting.
The figure below shows the relationship between 1/lambda and charge
(Blue: current parameters, Red: new parameters)
![boxandline_timingresolution_newlambda](https://cloud.githubusercontent.com/assets/8589041/8105180/d10a9c9c-1071-11e5-8ccc-d513b827845d.png)

sigma is not changed.

And the figure below shows the timing resolution (Blue: current parameters, Red: new parameters)
![boxandline_newlambdainvert_compare](https://cloud.githubusercontent.com/assets/8589041/8105142/755135e6-1071-11e5-9d9e-1fd056851a97.png)

from this figure, the difference of fitted curve was shown in the middle region of charge.
the difference is less than 0.1 ns but the new timing resolution seems to be better fitted than old version in ~10 ns region.
[...]

PR #83, opened 2015-06-11
fix PD timing resolution parameters

YujiOkajima:
As I described in #82, the timing resolution parameters of HPD should be fixed and I analyzed with new analysis method.

### DIFF
--- a/src/WCSimPMTObject.cc
+++ b/src/WCSimPMTObject.cc
@@ -1206,7 +1206,7 @@ G4double BoxandLine20inchHQE::GetPMTGlassThickness() {return 0.4*cm;}
 
 float BoxandLine20inchHQE::HitTimeSmearing(float Q) {
   G4float sig_param[4]={0.6314,0.06260,0.5711,23.96};
-  G4float lambda_param[2]={0.4094,0.06852};
+  G4float lambda_param[2]={0.4113,0.07827};
   G4float sigma_lowcharge = sig_param[0]*(exp(-sig_param[1]*Q)+sig_param[2]);
 
   G4float highcharge_param[2];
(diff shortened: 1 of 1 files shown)


eosullivan:
It will be a bit confusing if we have different fitting methods for the box & line and the HPDs. Even though this new method doesn't change the box & line results, I wonder if it is worth using just so we are consistent. Also, is there some way to make the box & line function match the data better? Are you still investigating?

YujiOkajima:
@eosullivan  I have analyzed the Box and Line PMT result with the same fitting method as HPD as a reference.
The result is shown in #82. 
The timing resolution of Box & Line PMT with new fitting method seemed to be better than current parameters.
The parameters of Box & Line PMT with new method has been already analyzed.
So, I can change it immediately.

I am not searching for new analysis method to make the Box & Line PMT function match the data better now.

If we search for more suitable analysis method, I have 3 ideas below.

One idea for making the Box & Line PMT function match the data is use more narrow width qbin.
The timing resolution of the new photodetectors are analyzed after TQ correction, but if the dependence is still remained, the timing resolution will be change.

another idea is changing TQ correction method. In current TQ correction, the TQ map is corrected based on timing distribution peak. The peak of EMG distribution is based on the Gaussian peak and 1/lambda.
So, even if the EMG peak is fixed, the Gaussian peak will be changed when the charge is changed. I think the difference of Gaussian peak is possible to make the timing resolution worse.
Therefore, I think correction of TQmap based on the Gaussian peak of EMG is possible to be better for analysis than current analysis.

last candidate is searching for more suitable fitting function.

YujiOkajima:
Based on the result of analysis with fitting of 1/lambda, the parameter of lambda was changed.
The figure below shows the result of 1/lambda fitting.
boxandline_newlambdainvert
From this figure, the lambda parameters were changed to (0.4113, 0.07827)

eosullivan:
Sorry for the delay in getting this merged. I still don't really understand how switching to 1/lambda improved the fit. Perhaps the improvement came mostly from the finer binning. Either way, it's safe to say that these new parameters are an improvement for the case of the HPDs so I'll merge this in. We should continue to think about how to improve the box and line fit.

Issue #85, opened 2015-06-18
Hard-coded .mac files not found should be a clearer error


tdealtry:
After working with WCSim for a couple of months, I've just found out that I've not been reading jobOptions.mac, jobOptions2.mac, or tuning_parameters.mac while running WCSim. This is because I've been running WCSim from a directory that isn't the root of the package (i.e. not the one with these files in). I see a few options for rectifying this:
1. If any of these files aren't found, throw a fatal exception.
2. Make it clear in the documentation that WCSim must be run from the root directory, or copies (or soft links) of these files must be present in your running directory (e.g. make it explicit at https://github.com/WCSim/WCSim/wiki/Tutorial:-How-to-set-up-GitHub-%28in-the-context-of-WCSim%29#test-that-your-code-is-working and elsewhere)
3. Add relative paths to WCSim.cc, so that these files are looked for at $WCSIMDIR/jobOptions.mac. Running a simple test on this, it seems that variable expansion must be performed first.
4. Specify the .mac file locations on the command line (with sensible defaults if they're not given).
I'd suggest doing 1 (so no one has a chance to get burnt again) and either 2 or 3 or 4.

cwwalter:
Good catch!

I think the entries there now set the defaults anyway so hopefully it hasn't affected any running results. 

I think we should do #1, at least to start.  Also see issue #3.

eosullivan:
Can I assign this issue to you @tdealtry? I agree that at the very least we can do option 1. And then maybe option 3. Anything we can do in an automated way, we should try to do.

tdealtry:
Yeah, I can do this. I'll submit a pull request for 1 soon.

I'm not certain that 3 is the best alternative option; having the DAQ options in there (e.g. selecting which digitizer and trigger classes to use) makes it difficult to save run history, or run with different options on the batch. Perhaps waiting until the pull request for that is finished would be the best thing before deciding which way to go.

A few days ago I was looking at something similar to #3 - actually consolidating all .mac files into a single .mac file. This would of course make this issue obsolete. However, some options complain that they've already been initialised. Would doing something like the following in SetNewValue() (i.e. adding a new bool class data variable) be acceptable?

  if (command == physListCmd && !physListCmdRead) {
    thisWCSimPhysicsListFactory->SetList(newValue);
    physListCmdRead = true;
}

cwwalter:
[...]

PR #86, opened 2015-06-29
WCSim exectuable will exit if jobOptions.mac, jobOptions2.mac, or tun…

tdealtry:
…ing_parameters.mac are not found. For issue #85

### DIFF
--- a/WCSim.cc
+++ b/WCSim.cc
@@ -113,7 +113,6 @@ int main(int argc,char** argv)
     G4UIsession* session =  new G4UIterminal(new G4UItcsh);
 
     // Visualization Macro
-    file_exists("vis.mac");
     UI->ApplyCommand("/control/execute vis.mac");
 
     // Start Interactive Mode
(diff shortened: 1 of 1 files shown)


cwwalter:
As I mentioned I don't think we want to require vis.mac.

WCSim.cc
```diff
@@ -114,6 +126,7 @@ int main(int argc,char** argv)
     G4String command = "/control/execute ";
     G4String fileName = argv[1];
 
+    file_exists(fileName.c_str());
```
cwwalter: What happens if you just run WCSim with a non-existent file name on the command line now?  I thought it already didn't work...

tdealtry:
I made the commit, but perhaps github doesn't make this a notification? I think this is now ready to merge.

eosullivan:
You usually don't get a notification for your own commits, although in the case of this commit I didn't get a notification either. Anyways, I checked out this pull request branch and it works, so I will merge this in. Thanks!

Issue #10, opened 2014-08-07
Add a validation suite for running after updates


cwwalter:
We should add a set of validation input files and .mac files so that we can check after updates that things still work properly. 

This might be in a separate repository.

tdealtry:
I was just looking at my assigned issues, and saw this (related to #162). From what I understand, @brichards64 has got Travis CI working, and it just needs someone with WCSim admin privilege to turn on.
This will be very useful in the near future (i.e. when #178 is merged) so that the other pending (and all future) pull requests don't have to be revalidated by authors. Instead this can be done automatically by Travis CI.
Of course after #162 is solved, we can leave this open until the exact physics checks are defined.

eosullivan:
Ok, so let's turn this on so we can at least start playing around with it a bit more. @brichards64, can you let me know what you need me to do to make this live? I also had some future improvement suggestions in our last email exchange, which would be a nice to have. I will attach them below:

 For future improvement, I can probably think of a couple other tests that would be nice for WCSim. A test to make sure a pull request is not significantly increasing the time to generate events would be great. And we also probably want to test with various detector configurations (SK, the HK options, etc) in case a change has been made to a specific detector component (if someone makes a change to the 20" B&L, for example).

brichards64:
Sure, Its been a long time since i made it and i think wcsim has advanced since then so ill check it all still works and make any changes needed then ill let you know how to integrate it.

From memory requires an admin to link the github account with the build server and add a couple of scripts. If you want validation as well as build testing adding a webpage to github is the best way to go. But ill outline the steps needed when i check its all still works well. I might also make some improvements to the validation such that anyone can add validation of any quantity they want easily.

PR #88, opened 2015-07-10
Start of a validation suite

tdealtry:
This is for issue #10 

This pull request has 2 main parts
1. a root wrapper ('rootwc') that automatically loads the relevant WCSim libraries and includes, so as to run root macros in compiled mode. This is good for both speed and debugging purposes.
   This wrapper also loads the T2K plotting style, to give WCSim output some consistency. Of course, this can be modified/removed as required
2. a group of python scripts (and helper root macros) that people can run to go through the steps of:
   - Creating a series of .mac files
   - Submitting jobs (either locally, or to a cluster)
   - Running a 'sample_readfile.C' style macro
   - Comparing the outputs of the 'sample_readfile.C' macro, either 
   - a) Identical files in different directories (e.g. checking the results before/after a change)
   - b) Different files in the same/different directory (e.g. checking how a new feature compares to an old feature)

I've copied in the README.md information for below convenience

This is not a complete solution to issue #10; additional requirements will include:
- Deciding which particle types/energies are to be verified
- Creating a macro (or macros) like sample_readfile.C which will create all the distributions that are to be verified
- Collating the results of the KS tests & flagging any distributions that are inconsistent
- Automating the whole process (i.e. a master python script)
### Full scripts suite

An automated way to go from .mac file creation & job submission through to WCSim output analysis and histogram comparison
## generate_mac_files.py

This script will generate a series of .mac files for a particle gun for the set of options given, and submit the jobs to run (either locally, or on a cluster)
Specify position/direction with either 3 vectors, or the allowed options of $WCSIMDIR/sample-root-scripts/MakeKin.py
Creates subdirectories for each set of 'jobOptions.mac' options specified, and uses filenames to distinguish 'novis.mac' options
Note: not all possible options are included yet, but can be easily added
## analyse_dir.py

This script will run a given .C analysis script over all files in a given directory

The default macro is daq_readfile.C, based on sample-root-scripts/sample_readfile.C but with (many) more histograms created, dealing mostly with number/times of hits/digits, and whether hits/digits are noise/photon-induced
## plot_lots.py

This script will glob over all files in a given directory, and create histograms comparing multiple other files using compare_lots.C
The other files either come from:
- The same directory; use --difference to specify differences e.g. noise rate
- A different directory
  The 'multiple' files are specified with --diffplots. e.g. _0e-?_10e-:50e-_ will compare 0, 10, and 50 MeV electrons with identical other options
  The histograms to compare are read from the file comparison_hists.py
## plot.py

Very similar to plot_lots.py, but only allows a 1 vs 1 comparison. Has a couple of extra run options:
- filesize - compares filesize and running time between 2 files (note runtime will only be valid if you used the 'local' runmode in generate_mac_files.py)
- compare_stacks - creates stacked histogram comparisons (currently only works for 2)
- compare - calls compare.C, which is very similar to compare_lots.C, just for exactly 2 files
### Usage

Setting up
- $WCSIMDIR must be setup to point to the folder where WCSim.cc resides
- $PATH should contain the place where WCSim resides, and also where rootwc lives e.g.:
  export PATH=${WCSIMDIR}/rootwc:${WCSIMDIR}/bin/Linux-g++:${PATH}

Running
- Run generate_mac_files.py to create WCSim output files
- Run analyse_dir.py to create the relevant histograms in an output file that can be compared
- If comparing histograms or stacks, modify comparison_hists.py to choose which histograms/stacks to compare
- If comparing histograms, run plot_lots.py to make the histograms
- If comparing speeds/sizes or stacks, run plot.py
### Example

Compare 5, 10, 50 MeV electrons for the SKDETSIM-like and new-style digitisation routines, with/without noise, submitting WCSim jobs locally
- python $WCSIMDIR/verification-test-scripts/full_scripts_suite/generate_mac_files.py--batchmode local --WCgeom SuperK --HKwatertanklength 24750 --PMTQEMethod Stacking_Only --SavePi0 false --DAQdigitizer SKI,SKI_SKDETSIM --DAQtrigger NHits,SKI\
  _SKDETSIM --DAQnhitsthreshold 25 --DAQnhitsignorenoise --DAQnhitswindow 200 --DAQsavefailuresmode 0 --DAQsavefailurestime 250  --DarkNoiseRate 0,8.4 --DarkNoiseConvert 1.367 --DarkNoiseMode 1 --DarkNoiseWindow 2000 --GunParticle e- --GunEnergy\
  5,10,50 --GunPosition 0,0,0 --GunDirection 1,0,0 --DAQnhitsignorenoise --NEvents 2000
- python $WCSIMDIR/verification-test-scripts/full_scripts_suite/analyse_dir.py --dir .
- Compare like-for-like
  - python $WCSIMDIR/verification-test-scripts/full_scripts_suite/plot_lots.py --dir1 SKI_NHits_fails0_250_NHits25_200 --dir2 SKI_SKDETSIM_SKI_SKDETSIM_fails0_250_NHits25_200 --legend 'New_code:Old_code' --mode compare_lots
- Compare 5,10,50 MeV on the same plot
  - python $WCSIMDIR/verification-test-scripts/full_scripts_suite/plot_lots.py --dir1 SKI_NHits_fails0_250_NHits25_200 --dir2 SKI_SKDETSIM_SKI_SKDETSIM_fails0_250_NHits25_200 --legend 'New_5MeV:Old_5MeV:New_10MeV:Old_10MeV:New_10MeV:Old_10MeV' \
    --mode compare_lots --diffplots _5e-:_10e-,_50e- --required_filename_part _5e-
- Compare speed/filesize
  - python /home/other/tdealtry/Documents/myWCSIM/WCSim/verification-test-scripts/full_scripts_suite/plot.py --dir1 SKI_NHits_fails0_250_NHits25_200 --dir2 SKI_SKDETSIM_SKI_SKDETSIM_fails0_250_NHits25_200 --legend 'New_code,Old_code' --mode fil\
    esize


eosullivan:
I think these are great tools to have in WCSim. They will be useful for the DAQ verification, and can be modified for more general verification. I can see it being helpful to have the ability to generate macros in an automated way, both for verification and for production runs. I also think it is great to have the flexibility swap in your own .C analysis file into the framework that loops over the macros and generates plots. 

For now, I will just comment on generate_mac_file.py since the files after that are very DAQ specific and the functionality needed for them are not yet in the code. 

* The generate_mac_file.py requires the argparse module, which is not by default included in python (at least not in my version). We should say in the README that this is required, and perhaps even include it in the directory.  
* Also in generate_mac_file.py, it's going to be a pain to have to change the options each time we put a new variable into DetectorMessenger and allow it to be set in the macro. I wonder if it is possible to have a python script that generates the generate_mac_file.py file by looking in DetectorMessenger for the available options?
* In your pull request example (and your README), you have: 
python $WCSIMDIR/verification-test-scripts/full_scripts_suite/generate_mac_files.py--batchmode ...
There should be a space between .py and --. 
* You should make some defaults. The defaults should be the settings in novis.mac. 
* HKtanklength should not go into the macro unless someone wants to specify it. 
* In the execution line of generate_mac_file.py, change WCSim to $WCSIMDIR/bin/Linux-g++/WCSim. 

And a general comment:

* It would be nice if you had an easy option to compare against a clean checkout of the code rather than just other flag options (or having to change $WCSIMDIR to point to a different location). Ideally, we would have a script that could check between your feature branch and your up-to-date develop branch (or, even better, a script that checked out your pull request and compared that to your develop branch). At the very least we should have some option like there is in the current verification setup where you check out a clean version of WCSim in a directory called WCSim_clean. Then, you could have a $WCSIMDIR_CLEAN path and everything that was done in $WCSIMDIR could be simultaneously done in $WCSIM_CLEAN and the plotting scripts could compare between files run with the two different executables. 

How easy is it to make this more general so we can merge this in without the DAQ code in place? Could you set it up with the basic verificationhitchangetime.C or samplereadfile.C files and take out the DAQ-specific options from the macro generator? Then, once the DAQ code is in place, we can add daq_readfile.C and put the DAQ options back into generate_mac_file.py.

tdealtry:
Thanks for the positive response.

I've included argparse.py in the folder. I'd be happy to do the alternative of giving a download location in the readme. Let me know what you prefer.
I've also specified defaults, now set HKtanklength appropriately, removed (commented out) the new DAQ options, and fixed the readme typo.
I've not changed to $WCSIMDIR/bin/Linux-g++/WCSim, but instead to $WCSIMDIR/bin/$G4SYSTEM/WCSim, in order to be portable.
I also changed the default macro of analyse_dir.py to be sample_readfile.C. There's obviously no output root files, but it works. I also added an option to plot.py to run verification_HitsChargeTime.C

Thinking about your idea of automatically generating generate_mac_file.py - I agree that it'd be nice, but it'd be tricky to do in practice. You need to think about command ordering (e.g. WCSim/Construct) and command dependency (e.g. /WCSim/HyperK/waterTank_Length).
What would be nice (and I've thought about before but not worked out how to code it up yet, other than brute force parsing of the .cxx) is to grab the allowed options (of e.g. /WCSim/WCgeom) & defaults from the Messenger code. Would this be a happy medium? To get the defaults correctly, adding a call to SetDefaultValue() for every Messenger variable would be a good idea, but I'm not sure this is what you want the WCSim code to actually do?

In terms of your general comment, I've added a new (currently unfinished) script. I think forcing the current macros to use $WCSIMDIR and $WCSIMDIR_CLEAN takes away a bit of the generality of the scripts, but having a wrapper that'll do the checking out, building, running, etc of code into 2 branches is a useful way to go.
It's split into 3 parts:
grabcode - download & compile the current develop, as well as the new repo/branch you want to compare to
run - call generate_mac_file.py
analyse - call analyse_dir.py, plot_lots.py, etc.

eosullivan:
I haven't forgotten about this pull request @tdealtry! I actually have tried to go back and look through it twice this week. 

I think the challenging thing is that since it isn't fully functional, it's hard to make specific suggestions. What we really need are some programs that test the general outputs of the WCSim files and that perform the functions that this validation routine intends (like saving plots so they can be compared later). Then we can use as these as defaults in the analysis scripts. Do you have time to keep going with this?

As for generating generate_mac_file.py, are the only complications the examples of command ordering and dependencies that you used? If so then I think you can get around them. If anything other than SuperK is chosen as the detector configuration, then put /WCSim/Construct on the next line. And we can remove the /WCSim/HyperK/waterTank_Length from generate_mac_file.py if it's going to be a problem. That is really only specifically for the egg-shape design which is probably going to become obsolete. 

I like the idea of the script that downloads and compiles the current develop branch and then allows you to compare with your current branch (this is even better than having a WCSIMDIR_CLEAN). So, keep going to implement that idea if you have time.  

A couple of other things I wrote down while using your scripts:
* Can we set a default for --batchmode? Probably should be local. 
* Set the default for PMTCollEff to ON. I know you pulled this from novis.mac where off was the default, but this PMTCollEff being incorrectly set has lead to lots of headaches for people using the code so let's just leave it on. This has recently been fixed in novis/vis.mac. 
* I also noticed that when I ran plot.py is is throwing an error perhaps related to the syntax of how you are calling verification_HitsChargeTime.C? The error is below:
```
 File "plot.py", line 151
    command = 'rootwc -l -b -q \'$WCSIMDIR/verification-test-scripts/verification_HitsChargeTime.C("'file1 + '","' + file2 + '")\''
                                                                                                         ^
SyntaxError: invalid syntax
```

tdealtry:
Hi @eosullivan. I've got a bit of time to do this. You can assign me #10 if you like.

I agree that we should do the review in a step-by-step process (and I think there's no reason to rush it?). I think what is ready to be checked is:
* the rootwc directory
* run_verification.py in 'grabcode' mode
* run_verification.py in 'run' mode (I've added a suggested set of low energy verification particle gun runs, but the actual set of runs can (and should) be discussed and iterated). But the idea of how to do it is there.
* generate_mac_files.py
* analyse_dir.py (sample_readfile.C now has the option to save its histograms, so it's a little easier to see what's going on)

Then once we're both happy with these, move on to the rest. I.e.:
* run_verification.py in 'run' mode (this is intended to call analyse_dir.py)
* plot*.py, compare*.C
* run_verification in 'compare' mode (this is intended to call plot*.py)

I've done the things you suggested. I've also updated the allowed geometries (removed DUSEL, added the new HyperK geometries).
One thing I've not tested are python dependencies (since argparse happily lives in the versions I have access to). But it should be working ok with the inclusion of argparse.py in the directory (I even get an argparse.pyc). So I'm >99% that it's working ok.


Just for reference, I've run the following successfully (which should go a long way to verifying the first set of bullet points work):
* python run_verification.py --setup-script /home/hyperk/tdealtry/Documents/myWCSIM/source_at_start.sh grabcode --new-repo git@github.com:tdealtry/WCSim.git --new-branch feature/issue10
* python run_verification.py --setup-script /home/hyperk/tdealtry/Documents/myWCSIM/source_at_start.sh run --low-energy
* for d in wcsim-work/*; do cd $d; python ../../analyse_dir.py --additional-macro-options true; cd -; done;

eosullivan:
Thanks @tdealtry for agreeing to work on this important task; I've assigned it to you in the issue tracker. 

Your plan for review sounds good. We can do it in steps and there is no real urgency here. As always (but especially since this is something that everyone will end up using once implemented), other WCSim developers should also feel free to check out this PR branch, try things out, and give feedback.

eosullivan:
With the new travis implementation, which automatically performs validation, I think it's time to close this PR. Thanks for putting in the groundwork for this, @tdealtry. This will continue to be accessible in the closed pull request section if we need to refer to this in the future.

Issue #11, opened 2014-09-09
Further subdivide the digitizer into electronics and DAQ.


cwwalter:
This depends on issue #4 being resolved.  But, after that work is done the current digitizer can be further sub-divided in electronics and DAQ pieces.  This way one could implement things like the current SK software triggers cleanly.

eosullivan:
I think this plan looks good in principle. Extending this idea of deriving from a base class to include the digitizer and trigger/DAQ is certainly where we want to go with this. For the actual details of how the electronics will be set up in the code, this would probably benefit from having an SK electronics expert look over it. In the meeting last night, Chris suggested that an SK electronics expert could perhaps come and give a talk at the HK software meeting to refresh us on the details of how the new digitization scheme works.

eosullivan:
It looks like I can only assign one person to this, but this will also be the responsibility of @dewhurst, and @stewartt1982. Once the code is split up into the digitizer and the DAQ, @thomaslindner can make the digitizer more modular and can add in the FADC.

Issue #48, opened 2015-02-12
Move the PMT threshold cut from WCSimWCDigitizer.cc to WCSimWCPMT.cc


eosullivan:
This is a bug that was discovered by Debra Dewhurst. The digitizer threshold is implemented before the PMT threshold, which leads to some events in the final digitized collection with less hits than the digitizer threshold. This is an issue only for low energy events. To solve this, the PMT threshold function should be moved into the WCSimWCPMT.cc file (which is probably where it belongs anyways).

dewhurst:
I looked at moving the PMT threshold cut from WCSimWCDigitizer to WCSimWCPMT but as far as I can see this isn't possible at the moment as the information needed for the threshold cut isn't available in WCSimWCPMT. For example, I think information about the gates that's in WCSimWCDigitizer is needed and it's not there at the moment. And also the containers are not the same - WCSimWCPMT deals with hits, whereas WCSimWCDigitiser deals with PMTs hit. So I came to the conclusion that in the way the digitisation is currently done the threshold cut can't be moved to WCSimWCPMT yet. I think this is something that will be easier to do when we look at separating the digitisation and triggers but at the moment isn't really possible. 

I have implemented a possible quick fix in WCSimWCDigitizer.cc where DigitizeGate() makes a check to ensure that after the PMT Threshold is applied the number of hit PMTs > GlobalThreshold (currently 25 hits). This removes the triggers with < 25 hits. However, this won't deal with the fact that some events may have > 25 hits in the entire event window but < 25 hits (after the PMT threshold cut) in the 200ns window that caused the trigger - therefore, if we could go back and say "try to find a trigger again", it might find a different trigger. The only way to get around this is to do:

1. the trigger to be found, 
2. the PMT threshold to be applied,
3. check whether the 200ns window still fulfils the trigger requirements 
4. if it doesn't go back to 1 and repeat.

But again this involves re-writing the digitisation, which is what we're hoping to do in the future but is not going to be quick. So my quick fix does steps 1-3 and removes the ones that no longer fulfil the trigger requirements but then it doesn't do 4 - looking for a new 200ns window with > 25 hits.

[...]

Issue #13, opened 2014-09-09
Simulate SK-IV Electronics


cwwalter:
After #4 and #11 are finished we could simulate the SK-IV electronics setup.  This is necessary to use WCSim for real SK/T2K based analysis work.

eosullivan:
Now that issue #11 is resolved, new digitizer and trigger schemes can be added easily. This should make implementing this feature now possible.

jeimber:
I've written an SK-IV trigger class that produces SHE, HE, LE and SLE triggers.  I've tested it with simple particle gun events.  At the moment, trigger overlaps are handled differently from the real SK-IV code.

eosullivan:
Great, it's nice to see this work is coming together. I know the software meeting is late in the UK, but it would be great if you could present on your progress at tomorrow night's meeting, if possible.

eosullivan:
Actually, Chris just pointed out the software meeting is next week (Feb 24/25), and not this week.

jeimber:
I'm currently in Japan, so it'll be 8am on 25th for me, right?
I've improved the trigger overlap handling but it's not tested yet.  I know how to increase the number of particles launched into the tank, is there a method in \gps to space them in time to get multiple peaks in the same digit stream?

eosullivan:
If you are in Japan, the meeting is at 7:30 am on the 25th. 

The documentation for the available gps commands is here: https://geant4.web.cern.ch/geant4/UserDocumentation/UsersGuides/ForApplicationDeveloper/html/ch02s07.html. I haven't tried this myself, but you might have some luck defining multiple sources with different t0's. @tdealtry, did you ever try and do this while testing your digitizer code?

tdealtry:
I've not tried adding multiple particles at different times. But this is also a good resource (has some more intuitive & well-described examples) http://nngroup.physics.sunysb.edu/captain/reference/master/detSim/dox/detSimGPS.html I think the last example would give you fine control, when combined with /gps/time

jeimber:
Thanks Tom, Clark's Captain simulation documentation (minus a couple of commands that weren't recognised, perhaps Geant4 version specific) did exactly what i needed.

jeimber:
Am i right in thinking that there isn't a trigger time stored in WCSimRootTrigger?  Is there a reason for that?

tdealtry:
The information is already stored in the header:
trigger->GetHeader()->GetDate()
(a confusingly named variable I know...)

PR #89, opened 2015-07-16
Feature/issue11 - separate the digitization and triggering code

stewartt1982:
This commit is to provide a re-write of the digitisation and triggering in WCSim (see issue <a href=https://github.com/WCSim/WCSim/issues/11>#11</a>).  The digitisation and triggering are independent from each other and can new digitisation and trigger algorithms can be developed and easily changed.  More details of this pull request can be seen in the talk <a href=http://www-sk.icrr.u-tokyo.ac.jp/indico/getFile.py/access?sessionId=0&resId=0&materialId=0&confId=2396>here</a>. Verification plots can be found  <a href=https://gist.github.com/tdealtry/0efb539cd2ab425900e8>here</a>.

In this pull request the existing WCSim digitiser and trigger are available.  This will be removed in a future pull request.

## New File Descriptions
**WCSimWCAddDarkNoise.cc/WCSimWCAddDarkNoise.hh** - The original **WCSimWCDigitiser** class adds dark noise hits to the found gates.  The new routine **WCSimWCAddDarkNoise** adds hits before gates have been found. To add noise there are two methods available described in the section on **WCSimDarkRateMessenger**
with **time** given in nanoseconds.

**WCSimWCDAQMessenger.cc/WCSimWCDAQMessenger.hh** - Class handles options set in the jobOptions.mac file to allow the trigger and digitisation algorithms to be specified.
* **/DAQ/Digitizer `<Digitizer Name>`** - current options: SKI_SKDETSIM uses the existing digitizer and trigger in WCSim. SKI and SKIV use digitizer defined in WCSimWCDigitizerSK.hh/WCSimWCDigitizerSK.cc (just with different deadtime values). It should be noted that SKIV isn’t strictly SKIV electronics.
* **/DAQ/Trigger `<Trigger Name>`** - current options: NHits (use WCSimWCTriggerNHits - a simple Nhits-type trigger), NHits2 (use WCSimWCTriggerNHits2), SKI_SKDETSIM (use existing digitiser & trigger).
* **/DAQ/TriggerNHits/Threshold `<Trigger threshold for NHits trigger>`**
* **/DAQ/TriggerNHits/Window `<Width of the NHits window in nanoseconds>`**

**WCSimWCDigitizerNew.hh/WCSimWCDigitizerNew.cc** - Defines a new base class from which all digitizers inherit. Defines new WCSimWCDigitsCollection* DigiStore where digits are stored. These digits are used as input into a trigger class which inherit from the base class in WCSimWCTrigger.cc. Derived class is WCSimWCDigitizerSKI, an SKI-like digitization.
 
* **void AddNewDigit(int tube, int gate, float digihittime, float peSmeared, std::vector`< std::pair<int,int> >` digi_comp)** - Adds a new digit to the DigiStore WCSimWCDigitsCollection object
* virtual void DigitizeHits(WCSimWCDigitsCollection* WCHCPMT) - virtual method which must be implemented by classes which derive from *WCSimWCDigitizerNew*.

**WCSimWCTrigger.hh/WCSimWCTrigger.cc** - Defines a new base class from which all triggers inherit. The input WCSimWCDigitsCollection*  DigiStore defined in the class **WCSimWCDigitizerBase** used as input.  Input from WCSimWCDAQMessenger Nhits threshold, trigger window width. Classes which inherit from the WCSimWCTrigger base class must define **virtual void DoTheWork(WCSimWCDigitsCollection* WCDCPMT)** which calls the trigger algorithm.  Currently one trigger algorithm is implemented **void AlgNHits(WCSimWCDigitsCollection* WCDCPMT, bool remove_hits, bool test=false)** which defines a SK-like Nhits trigger.

Currently all algorithms are defined in the base class. This is to avoid code duplication in the case where one wants to apply multiple trigger algorithms consecutively, and the second trigger acts on the ‘failed digits’ of the first trigger. It is unclear as to whether this is algorithmically the best way to do things; multiple triggers may instead be defined in the same algorithm, in which case the algorithms should be moved to the concrete implementation classes.

The derived classes in WCSimWCTrigger are: 

**WCSimWCTriggerNHits** - Defines a NHits trigger.  Inherits from WCSimWCTriggerBase.

**WCSimWCTriggerNHits2** - A second NHits trigger. This calls the NHits trigger algorithm 2 times. The second call to the NHits algorithm is called on hits that failed the first algorithm. This is a test of applying multiple trigger algorithms consecutively, with the second trigger acts on the ‘failed digits’ of the first trigger. Still in development.

**WCSimEnumerations.hh** - Defines what kTriggerNDigits, etc mean. This enumeration information (TriggerType_t) is saved in the output file. Having this in a separate class allows for this to be implemented in user-defined analysis macros using #include.  

## Modified file descriptions
**WCSimDarkRateMessenger.hh/WCSimDarkRateMessenger.cc** 
Adds the ability to use the new dark noise class WCSimWCAddDarkNoise.  There are two modes for adding dark noise available, implemented in a new constructor **WCSimDarkRateMessenger(WCSimWCAddDarkNoise*)**

**/DarkRate/SetDarkMode 0** adds noise hits to a window specified by a start and end time
* **/DarkRate/SetDarkLow `<time1>`**
* **/DarkRate/SetDarkHigh `<time2>`**

and

**/DarkRate/SetDarkMode 1** adds dark noise hits to a window of width specified by:
* **/DarkRate/SetDarkWindow `<time>`**

where **time** is the size of the window in nanoseconds to add dark noise to centred on the hit (e.g.  Hit time +/- **time**/2).  If a window around a hit overlaps with another window around a hit then the windows are merge taking the window which maximises the window size.

The existing **/DarkRate/SetFrequency** and **/DarkRate/SetConvert** are still used.  

**WCSimDarkRateMessenger::WCSimDarkRateMessenger(WCSimWCAddDarkNoise* darkratepoint):WCSimAddDarkNoise(darkratepoint)** added to allow parameters to be passed to **WCSimWCAddDarkNoise**.

**WCSimWCAddDarkNoise.cc/WCSimWCAddDarkNoise.cc** - Dark noise routine to add noise before digitization and triggering.  Called in WCSimEventAction using parameters set in .mac file.  Noise is added before digitisation and triggering.  Either **mode 0** or **mode 1** must be used as described above. If **WCSimWCDigitizer** class is used the original dark noise routine is used.

**WCSimEventAction.cc** - Now calls user specified digitisation and trigger classes in addition to the existing WCSimWCDigitizer. Details of selecting between digitisers described in section on **WCSimWCDAQMessenger.cc**.  

**WCSimEventAction.hh** -  Implements void SetDigitizerChoice(G4String digitizer) and void SetTriggerChoice(G4String trigger) to set strings for the chosen digitizer and trigger. Also passes the DigitizerBase & TriggerBase pointers around appropriately. Also saves the raw hits in the correct order for the digit composition information to be valid.

**WCSimRootEvent.hh/WCSimRootEvent.cc** - Write two new pieces of information included in the root file. 
* The ID for each photon making up a digit.
* Information on the trigger algorithm used and generic trigger information (e.g. how many hits in the NHits trigger window)

**WCSimWCDigi.hh** - Multiple changes to WCSimWCDigi.hh.  
A helper routine **void SortArrayByHitTime()** which sorts **std::map`<int,float>` pe** and **std::map`<int,float>` time** by time (smallest to largest). 

**std::vector`< std::pair<int,int> >` fDigiComp** which stores an ID for each photon making up a digit. There can be more than one digit in stores in a WCSimWCDigi object, hence the vector contains: `<digit number, photon id>` where the photon id is the index of the photon in the input DigitsCollection.

**inline void AddPhotonToDigiComposition(int digi_number, int photon_number)** Used to add an photon id to a digit

**void AddDigiCompositionInfo(std::vector`< std::pair<int,int> >` &digi_comp)**
Similar to the above but adds an entire vector of photon id to a digit

**WCSimWCDigi.cc**

**std::vector`<int>` WCSimWCDigi::GetDigiCompositionInfo(int gate)** 
Defines a method to return a vector of photon ids for a specific trigger.

**void WCSimWCDigi::RemoveDigitizedGate(G4int gate)**
This removes an element from the maps, vectors, and sets, and counters that were filled by WCSimWCDigitizerBase::AddNewDigit().

**WCSim.cc** - Added the creation of a **WCSimWCDAQMessenger** object which reads from the jobOptions.mac file to get DAQ configurations (to specify which trigger and digitisation to be used).

**WCSimDetectorConstruction.hh** - Add method GetDetectorName() to return the detector name WCDetectorName.

**WCSimWCDigitizer.hh/WCSimWCDigitizer.cc** - Class now inherits from WCSimWCTriggerBase. Also correctly fills the new variables with trigger type/additional information & photon composition of the digit.


AlexFinch:
I will be reviewing this code.

eosullivan:
Great to see this moving forward. This will be one of the biggest changes to WCSim yet! I look forward to hearing the internal review results from @AlexFinch.

src/WCSimWCAddDarkNoise.cc
```diff
@@ -0,0 +1,254 @@
...
+
+#ifndef WCSIMWCADDDARKNOISE_VERBOSE
+//#define WCSIMWCADDDARKNOISE_VERBOSE
+#endif
+
+WCSimWCAddDarkNoise::WCSimWCAddDarkNoise(G4String name,
+                                   WCSimDetectorConstruction* myDetector)
+  :G4VDigitizerModule(name)
+{
+  this->myDetector = myDetector;
```
AlexFinch: Should probably move to initialiser list myDetector(myDetector)

src/WCSimWCAddDarkNoise.cc
```diff
@@ -0,0 +1,254 @@
...
+  DigiHitMap.clear();
+  DarkRateMessenger = new WCSimDarkRateMessenger(this);
+}
+
+WCSimWCAddDarkNoise::~WCSimWCAddDarkNoise(){
+  DarkRateMessenger = 0;
```
AlexFinch: Why are you doing this?

src/WCSimWCAddDarkNoise.cc
```diff
@@ -0,0 +1,254 @@
...
+#ifndef WCSIMWCADDDARKNOISE_VERBOSE
+//#define WCSIMWCADDDARKNOISE_VERBOSE
+#endif
+
+WCSimWCAddDarkNoise::WCSimWCAddDarkNoise(G4String name,
+                                   WCSimDetectorConstruction* myDetector)
+  :G4VDigitizerModule(name)
+{
+  this->myDetector = myDetector;
+  DigiHitMap.clear();
```
AlexFinch: Should you not also initialize: PMTDarkRate; ConvRate; DarkHigh; DarkLow; and DarkMode; ?

src/WCSimWCAddDarkNoise.cc
```diff
@@ -0,0 +1,254 @@
...
+  DarkRateMessenger = new WCSimDarkRateMessenger(this);
+}
+
+WCSimWCAddDarkNoise::~WCSimWCAddDarkNoise(){
+  DarkRateMessenger = 0;
+}
+
+void WCSimWCAddDarkNoise::AddDarkNoise(){
+  G4DigiManager* DigiMan = G4DigiManager::GetDMpointer();
```
AlexFinch: Does it ever make sense to call AddDarkNoise without having first called ReInitialize()? If the answer is No then it would be safer for AddDarkNoise to call ReInitialize() and make the latter private.

src/WCSimWCAddDarkNoise.cc
```diff
@@ -0,0 +1,254 @@
...
+  // Get the PMT collection ID                                                                                                                                             
+  G4int WCHCID = DigiMan->GetDigiCollectionID("WCRawPMTSignalCollection");
+  // Get the PMT Digits collection                                                                                                                                        
+  WCSimWCDigitsCollection* WCHCPMT =
+    (WCSimWCDigitsCollection*)(DigiMan->GetDigiCollection(WCHCID));
+  
+  if (WCHCPMT && this->PMTDarkRate>1E-307) {
```
AlexFinch: I prefer to be explicit  WCHCPMT != NULL

src/WCSimWCAddDarkNoise.cc
```diff
@@ -0,0 +1,254 @@
...
+    //Determine ranges for adding noise
+    if(DarkMode == 1)
+      FindDarkNoiseRanges(WCHCPMT,this->DarkWindow);
+    else if(DarkMode == 0) {
+      result.push_back(std::pair<float,float>(DarkLow,DarkHigh));
+    }
+    //Call routine to add dark noise here.
+    //loop over pairs which represent ranges.
+    //Add noise to those ranges
+    std::vector<std::pair<float, float> >::iterator it2 = result.begin();
```
AlexFinch: This while loop is really a for loop, it would be clearer to write it as such.

src/WCSimWCAddDarkNoise.cc
```diff
@@ -0,0 +1,254 @@
...
+      PMTindex[l] =0;
+    }
+    int num_hit_b4=0;
+    //int num_hit_after=0;
+    //    std::cout<<"entries before "<<WCHCPMT->entries()<<"\n";
+    //Set up proper indices for tubes which have already been hit
+    for (int g=0; g<number_entries; g++){
+      G4int tube = (*WCHCPMT)[g]->GetTubeID();
+      //std::cout<<"totalpe "<<tube<<" "<<(*WCHCPMT)[g]->GetTotalPe()<<"\n";
+      for (int gp=0; gp<(*WCHCPMT)[g]->GetTotalPe(); gp++){
```
AlexFinch: Isn;t this for loop just the same as 
 PMTindex[tube]+=(*WCHCPMT)[g]->GetTotalPe();
 num_hit_b4+=(*WCHCPMT)[g]->GetTotalPe();
???

src/WCSimWCAddDarkNoise.cc
```diff
@@ -0,0 +1,254 @@
...
+      ranges.push_back(std::pair<float, float>(t1, t2));
+    }
+  }
+  
+  //we need to ensure that the ranges found above are sorted first
+  //for the algorithm below to work
+  //output are pairs stored in the result vector
+  sort(ranges.begin(),ranges.end());
+  std::vector<std::pair<float, float> >::iterator it = ranges.begin();
+  std::pair<float, float> current = *(it)++;
```
AlexFinch: This while loop is really a for loop, so it should be coded as one. Or did I miss something?

src/WCSimWCAddDarkNoise.cc
```diff
@@ -0,0 +1,254 @@
...
+  while (it != ranges.end()){
```
AlexFinch: Please add a comment to explain what is happening here.

src/WCSimWCDigitizerBase.cc
```diff
@@ -0,0 +1,127 @@
...
+
+//RawSignalHitCollection *collection = new RawSignalHitCollection;
+
+WCSimWCDigitizerBase::WCSimWCDigitizerBase(G4String name,
+					   WCSimDetectorConstruction* myDetector,
+					   WCSimWCDAQMessenger* myMessenger)
+  :G4VDigitizerModule(name)
+{
+  G4String colName = "WCDigitizedStoreCollection";
+  this->myDetector = myDetector;
```
AlexFinch: move this->myDetector = myDetector; to  myDetector(myDetector) in initializer list

src/WCSimWCDigitizerBase.cc
```diff
@@ -0,0 +1,127 @@
...
+    DAQMessenger->TellDigitizer();
+  }
+}
+
+WCSimWCDigitizerBase::~WCSimWCDigitizerBase(){
+  //DarkRateMessenger = 0;
+}
+
+void WCSimWCDigitizerBase::Digitize()
+{
```
AlexFinch: Does it ever make sense to call  Digitize without first calling   ReInitialize?
 If not, then put call to ReInitialize inside Digitize as that is safer.

src/WCSimWCTriggerBase.cc
```diff
@@ -0,0 +1,270 @@
...
+const double WCSimWCTriggerBase::LongTime = 100000.0 ; // ns = 0.1ms. event time
+
+
+WCSimWCTriggerBase::WCSimWCTriggerBase(G4String name,
+				       WCSimDetectorConstruction* myDetector,
+				       WCSimWCDAQMessenger* myMessenger)
+  :G4VDigitizerModule(name)
+{
+  G4String colName = "WCDigitizedCollection";
+  this->myDetector = myDetector;
```
AlexFinch: move this->myDetector = myDetector; to  myDetector(myDetector) in initializer list

src/WCSimWCTriggerBase.cc
```diff
@@ -0,0 +1,270 @@
...
+#ifdef WCSIMWCTRIGGERBASE_VERBOSE
+    G4cout << "Saving trigger " << itrigger << " of type " << WCSimEnumerations::EnumAsString(triggertype)
+	   << " in time range [" << lowerbound << ", " << upperbound << "]"
+	   << " with trigger time " << triggertime
+	   << " and additional trigger info";
+    for(std::vector<Float_t>::iterator it = triggerinfo.begin(); it != triggerinfo.end(); ++it)
+      G4cout << " " << *it;
+    G4cout << G4endl;
+#endif
+
```
AlexFinch: Please add some comments to explain what is going on here. I can't  follow it.

include/WCSimWCDAQMessenger.hh
```diff
@@ -0,0 +1,50 @@
...
+
+  ~WCSimWCDAQMessenger();
+
+  void SetNewValue(G4UIcommand* command, G4String newValue);
+
+  void TellEventAction();
+  void TellTrigger();
+  void TellDigitizer();
+
+  void TellMeAboutTheDigitizer  (WCSimWCDigitizerBase* digitizer)   { WCSimDigitize = digitizer; }
```
AlexFinch: I don't like the name much. What's wrong with SetDgitizer ?

AlexFinch:
I have a general comment, not just about this new code. The make file runs doxygen, but the code does not contain doxygen style commenting, so the doxygen output is not that helpful. It would be useful to at least ensure each class and  method had a basic description of what it does and what the arguments are.

src/WCSimWCAddDarkNoise.cc
tdealtry: Perhaps @eosullivan can confirm, but I think having the default values setup in the WCSimDarkRateMessenger means that these variables in WCSimWCAddDarkNoise will always be set (whether included in the .mac file or not)

src/WCSimWCAddDarkNoise.cc
tdealtry: Done

src/WCSimWCAddDarkNoise.cc
tdealtry: This is how it was done in the old class - added a delete DarkRateMessenger to bring it in line with the use of other messenger classes elsewhere in the code

src/WCSimWCAddDarkNoise.cc
tdealtry: Done

src/WCSimWCAddDarkNoise.cc
tdealtry: Yes, done

src/WCSimWCAddDarkNoise.cc
tdealtry: Done (please ask if it's still unclear)

src/WCSimWCDigitizerBase.cc
tdealtry: Done

src/WCSimWCTriggerBase.cc
tdealtry: Done

src/WCSimWCTriggerBase.cc
tdealtry: Done (please ask if it's still unclear)

include/WCSimWCDAQMessenger.hh
tdealtry: Changed names to SetDigitizerOptions etc.

src/WCSimWCAddDarkNoise.cc
eosullivan: The standard is to put the default value into the .cc file, and to let the macros write over them, if specified. If you don't call something in the macro, then the variable won't be set, despite the default (I believe).

tdealtry: Hmm you're right. The purpose of the 'Default' in a G4UIcmdWithA* is a mystery to me then...
I've just committed a fix where the defaults in WCSimWCAddDarkNoise are set when the constructor of WCSimDarkRateMessenger is called. I've done it this way (rather than being explicit in the WCSimWCAddDarkNoise constructor) so as to not have to change defaults in 2 places in the code (which is of course likely to get out of sync).

tdealtry: I've just made another commit solving a similar initialisation problem in the constructors of WCSimWCDigitizerBase and WCSimWCTriggerBase - if the WCSimDarkRateMessenger pointer is NULL, exit.

src/WCSimWCDigitizerBase.cc
tdealtry: Digitize() was already doing what ReInitialize() does anyway. So I've implemented your suggestion (and the same for WCSimWCTriggerBase & WCSimWCDigitizer).

src/WCSimWCAddDarkNoise.cc
eosullivan: Actually, looking through it a bit more, it's confusing that the variable doesn't get set if not called in the macro. In the other messenger classes in WCSim this is the case (WCSimDetectorMessenger for example), but I see in your WCSimDarkRateMessenger class that you have set the "omittable" Boolean for each parameter to true, which should set the value to the default, for example: 
```C++
SetFrequency->SetParameterName("DarkRate",true);
```
Anyways, I agree that the defaults shouldn't be in two places. I also think that since the parameters are not set by default in the other messenger classes, having the explicit call to set the default could avoid confusion for users later (it is very clear that even if the parameter is not in the macro, the value will be set).

tdealtry: @eosullivan I'm having a little bit of trouble parsing your comment fully. It seems like you like the new way to initialise WCSimDarkRateMessenger, and may change the other messenger classes to do similar - is that correct?

eosullivan: Ok, let's try again post-caffeine. My points were:
- Despite my earlier comment, now that I see that you are setting the Boolean in SetParameterName to true, I am surprised that the value is not being set to the default if not specified in the macro. I thought that's what that Boolean did. 
- Since the Boolean is subtle (and I guess not working like I thought it did), I am on board with explicitly setting it in the constructor like you are doing. You should probably set that Boolean in SetParameterName to false if you are going to do that, though.

tdealtry: Looking into this a bit more, it seems that if both omittable and currentValueAsDefault are set to true, GetCurrentValue() should be called and SetNewValue() called afterwards with this information. A quick look at doing things this way was unsuccessful.
So I just set omittable to false as you suggested.

AlexFinch:
Just a quick note to say I am away on holiday until 10th August, but provided the comments I have made have been addressed I am happy for the code to be merged into the main branch.

src/WCSimWCAddDarkNoise.cc
tdealtry: Made it private & called in AddDarkNoise()

src/WCSimWCAddDarkNoise.cc
tdealtry: Done

src/WCSimWCAddDarkNoise.cc
tdealtry: Done

src/WCSimWCAddDarkNoise.cc
tdealtry: Improved the commenting. Please let me know if it's still unclear

tdealtry:
Thanks @AlexFinch I've just made another commit to address the outstanding comments.
I think it's time for the gatekeepers (@eosullivan, @cwwalter, et al) to take a look.

eosullivan:
Ok, thanks for your review @AlexFinch. I will start to look through in more detail. Before I do, though, I just wanted to revisit the issue of putting the dark noise routines in WCSimWCPMT as we discussed briefly at the HK meeting. Is it technically possible to do this? If so, I like the idea of having dark noise in the WCSimWCPMT class since this is where the dark noise will actually be generated. Are there any downsides to doing it this way?

tdealtry:
Both @stewartt1982 and I have had a look at this and there are no obvious showstoppers. Of course, there may be some unforeseen subtleties, but we're fairly confident it'll work.
This does need to be done AFTER this pull request though, as it will break the old routine (which of course will be pulled out in a future pull request...).

tdealtry:
Just a heads up - I've started to add some doxygen to make it easier on the reviewer(s). It's not perfect though - check #99

novis.mac
```diff
@@ -79,11 +84,21 @@
...
+/DarkRate/SetDarkHigh 100000
+/DarkRate/SetDarkLow 0
+/DarkRate/SetDarkWindow 4000
+
 # Or you can use the G4 Particle Gun
 /mygen/generator normal
 /gun/particle e-
 #/gun/particle pi0
-/gun/energy 500 MeV
+/gun/energy 10 MeV
```
eosullivan: Keep the default energy at 500 MeV in novis.mac.

novis.mac
```diff
@@ -91,5 +106,5 @@
 
 /WCSimIO/RootFile wcsim.root
 
-/run/beamOn 10
+/run/beamOn 1000
```
eosullivan: Keep the default number of events at 10 in novis.mac.

src/WCSimDetectorConfigs.cc
```diff
@@ -40,7 +40,6 @@ void WCSimDetectorConstruction::SetSuperKGeometry()
   WCAddGd               = false;
 }
 
-
```
eosullivan: Put this extra line back in so that this DetectorConfig file doesn't show up in the diff for this PR.

src/WCSimPMTObject.cc
```diff
@@ -1222,6 +1222,7 @@ float BoxandLine20inchHQE::HitTimeSmearing(float Q) {
 
 G4float* BoxandLine20inchHQE::Getqpe()
 {
+
```
eosullivan: Take out this extra line so WCSimPMTObject doesn't show up in the diff for this PR.

src/WCSimSteppingAction.cc
```diff
@@ -100,7 +100,7 @@ G4int WCSimSteppingAction::G4ThreeVectorToWireTime(G4ThreeVector *pos3d,
 } 
 
 
-void WCSimSteppingAction::Distortion(G4double x,G4double y)
+void WCSimSteppingAction::Distortion(G4double /*x*/,G4double /*y*/)
```
eosullivan: Did you mean to comment out the x and y here? It doesn't currently affect anything since this entire routine is commented out, but this should be left as is unless there is some reason to change it for this PR.

src/WCSimSteppingAction.cc
```diff
@@ -150,7 +150,7 @@ void WCSimSteppingAction::Distortion(G4double x,G4double y)
 }
 
 
-double WCSimSteppingAction::FieldLines(G4double x,G4double y,G4int coord)
+double WCSimSteppingAction::FieldLines(G4double /*x*/,G4double /*y*/,G4int /*coord*/)
```
eosullivan: Same comment as above. x,y, and coord are commented out in this PR. It should be left the same unless there is a reason to change it for this PR.

src/WCSimSteppingAction.cc
tdealtry: Sorry this was my fault - commiting some stuff to get rid of most of the (numerous) benign compiler warnings - when not quite understanding the github workflow. I think most (if not all) of these types of changes are in 495cb78ff834bf1b65d7cfebde497cb299bdbc85
I think the changes are valuable (so we can see actual compilation problems clearly) but how you want to proceed is up to you - either do it as part of this pull request, or as another one. If the latter, do you know the git voodoo to apply only commit 495cb78ff834bf1b65d7cfebde497cb299bdbc85 to a branch based off develop?

eosullivan:
Sorry for the lag in comments on this pull request. I am still reviewing the details of this pull request, but I have a couple of general comments. 
* What was the reasoning behind setting the digitizer/trigger options in jobOptions.mac (as opposed to the novis.mac file, for example)? Could we set a default and then write over it in the macro (like we do for the detector configuration)? My concern is with spreading the options over many places, though I could be convinced to agree with this method if there is some good reason for it.
* Can there be a compiled file with digitizer options and trigger options, rather than have a new file for each new thing? This is how things are currently done for the photosensors where everything is compiled into WCSimPMTObject.cc.
* Does this new dark noise method need to be in this pull request or could it be it's own? If it can be done without the rest of this digitizer/trigger stuff, then having the dark noise in it's own pull request could make this PR less of a monster change. If it can't be decoupled, or if it's too much work to separate it at this point, then I could be convinced to leave it in. 
* I know you have been doing some validation. Could you show some plots here in this pull request that shows the code is behaving as you would expect. I guess at this point, this would mean that the WCSim output is not changing when you use SKI_SKDETSIM.

jobOptions.mac
```diff
@@ -15,4 +15,17 @@
 /WCSim/physics/list WCSim # included for backward-compatibility.  Requires jobOptions2.mac
 
 
-
+#choose the Trigger & Digitizer type (and options)
+# set at least one (both to be sure) to SKI_SKDETSIM if you want to use the old combined Digitization & Trigger classes
```
eosullivan: Does this mean set either /DAQ/Digitizer or /DAQ/Trigger to SKI_SKDETSIM? This comment could maybe be clearer.

jobOptions.mac
```diff
@@ -15,4 +15,17 @@
...
+/DAQ/Digitizer SKI
+/DAQ/Trigger NHits
+#do we want to save only triggered events (mode 0), both triggered events & failed events (mode 1), or only failed events (mode 2)?
+/DAQ/TriggerSaveFailures/Mode 0
+#what trigger time to use for the failed events?
```
eosullivan: There are still some questions in the comments here regarding the electronics settings. Is there a plan to ask an expert about these numbers?

tdealtry: The current Digitizer/Trigger are direct analogues of the SKI implementation of the old class. But of course, since the reordering, we're now some hybrid of SKI & SKIV electronics/DAQ that has never been built. So the default numbers have been chosen for a direct comparison to the old code. @jeimber is working on implementing real SKIV with #13, which will of course have "reality" as a major driving force.
Just to be clear, the TriggerSaveFailures options are for trigger development, and not to be seen as a 'real' trigger.

tdealtry:
* The use of jobOptions.mac (rather than novis.mac) was due to what Chris said at the software meeting in June, after I said the implementation using novis.mac was somewhat complicated. To be honest, the jobOptions.mac is similarly complicated, so I'd be ok with going back to novis.mac. I agree that it's best to keep options together. Please check the novis.mac implementation we had (1bc2d20c6df89550c7d9bee74a982d93fe2001f9 is the diff when the switch to jobOptions.mac was made) and let me know what you think (I'm not a fan of the detector configuration implementation, because /WCSim/Construct is something very important and is easily missed in user .mac files).
* I guess we could but everything into WCSimWCTrigger.cc and WCSimWCDigitizer.cc. They'll be long files though! We'll make this change sometime over the next few days. We'll name the WCSimWCDigitizer file WCSimWCDigitizerNew until we get rid of the old class in a future pull request.
* We're putting together a document with a load of plots showing the differences between using SKI_SKDETSIM and the old code, and the differences between the new algorithms and SKI_SKDETSIM. That'll be ready early next week. But until then, the spoiler is that we still get the same results as section 3 of http://www-sk.icrr.u-tokyo.ac.jp/indico/getFile.py/access?sessionId=0&resId=0&materialId=0&confId=2396 - i.e. identical.
* In terms of attaching validation documentation, it's a real shame that github doesn't let you attach pdfs. I do have an idea though - could a folder be made in WCSim-Documentation where such pdfs could be added for posterity?

stewartt1982:
Regarding whether the new dark noise method can be removed from the pull request.  I would prefer not to move this routine to a separate pull request.  
* It is technically possible to remove the WCSimWCAddDarkNoise the pull request but this will leave the new digitizer/trigger code without the ability to add dark noise.  The existing dark noise method in WCSimWCDigitizer is incompatible with the new digitization/trigger classes and would require modification to work properly.  As the new WCSimWCAddDarkNoise class is already a somewhat modified version of the existing dark noise method I think this would gain few time/effort savings.
* WCSimWCDigitizer could be modified to use the new WCSimWCAddDarkNoise class such that the addition of dark noise is uniform for all digitizer classes.  As WCSimWCDigitizer will likely be removed after the completion of this pull request it would be mostly wasted effort.
* The changes are confined to 3 classes: WCSimWCAddDarkNoise, WCSimDarkRateMessenger, (the first two classes having the bulk of the changes) and WCSimEventAction

tdealtry:
I just wanted to explain the motivation of the previous 4 commits. In real-life electronics, multiple digits per PMT per trigger window are created. We had this in mind when we were creating the new classes, but discovered a couple of issues. These are now fixed in these commits, so multiple digits per PMT per trigger are now possible, and filled correctly (both internally and in the output file).
The motivation for not using the WCSimWCDigi container is that is setup for exactly one digit per gate, which would make using it for the triggered digits tricky and convoluted. Therefore, a new container class was written (WCSimWCDigiTrigger) to store the triggered digits. It's heavily based on WCSimWCDigi of course, but uses multimaps, and strips out the variables/methods not used by the trigger classes.

eosullivan:
Ok, so you have made it possible to have multiple digits per trigger, but is it being utilized yet? Or is this just setting up for a future enhancement?

tdealtry:
It's being utilized now. (It's always been the idea to do this for this PR; it was just incorrectly implemented until the recent changes)

eosullivan:
Have you ever used the Gist feature in GitHub? I just spent a bit of time this morning playing around with it, and I'm not super in love with it, but it could work for our verification document needs. Gist seems like a place where you can store documents in a more casual way than an official repository, but still have it be more permanent than, say, a dropbox file that may be deleted if you purge your computer. Take a look and let me know what you think. 

The basic recipe is the same as for a normal repository: create a Gist, clone the Gist to your computer, git add your PDF, and push your PDF to your Gist. Then you can include the URL for your Gist in the pull request (maybe even in the PR description so it is easy to find). I did a little test and the URL seems to stay static despite making commits to the Gist. 

If you haven't used Gist before, some resources I used are [this page on how to create Gists](https://help.github.com/articles/creating-gists/) and [this page on how to add documents](http://www.d3noob.org/2012/12/loading-thumbnail-into-gist-for.html).  This is [my test Gist] (https://gist.github.com/eosullivan/f43defebf36e5f7e8e66) that I made.

eosullivan:
As for our discussion before, let's keep the dark noise method in this PR, then. 

Also, just so I understand, what was complicated about having the options in novis.mac originally? Is it that the order of the commands were important and easy to mess up? Did something need to be specified before WCSim.cc gets called and SuperK gets built?

tdealtry:
Hi @eosullivan. Sorry for the delay (holidays). For the .mac options, the way to do it is either:
* In jobOptions.mac (as the code is now)
  * Create the WCSimWCDAQMessenger in WCSim.cc
  * Pass it to the WCSimEventAction constructor
  * Tell the WCSimEventAction the options with WCSimWCDAQMessenger::SetEventActionOptions()
  * Create the dark noise / digitizer / trigger classes in the WCSimEventActionConstructor
  * Tell the digitizer / trigger classes the options with WCSimWCDAQMessenger::Set*Options() called in the digitizer / trigger constructors
* In novis.mac (read the diff 1bc2d20c6df89550c7d9bee74a982d93fe2001f9 backwards)
  * Create the WCSimWCDAQMessenger in the WCSimEventAction constructor
  * Have methods in WCSimEventAction (CreateDigitizerInstance() and CreateTriggerInstance()) that are called in WCSimWCDAQMessenger::SetNewValue() (i.e. the class instances are created as the options are read in)
  * Thinking about this now, it'd be better to move the calls to Create*Instance() into WCSimEventAction (maybe in BeginOfEventAction()) to deal with the cases 1) no option specified (i.e. using default) 2) option multiply specified (shouldn't happen of course, but typos can be made)
  * Or yet another alternative would be something like /WCSim/Construct that'd be the trigger to call Create*Instance() (I'm not a fan of this though - it'd lead to mysterious seg faults if the user doesn't supply the line)
  * Tell the digitizer / trigger classes the options with WCSimWCDAQMessenger::Set*Options() called in the digitizer / trigger constructors

eosullivan:
Ok, thanks for the summary. I think I like the novis.mac way since it doesn't look so much more complicated than the jobOptions.mac way and it's more in line with how other options are set. Let's avoid a /WCSim/Construct situation for the reasons you mention. 

So, what's left for this PR? I would like to see the validation plots. I will also go through this one last time this week to see if there are any outstanding questions/concerns from me. Are all the features that you want to go in with this PR already committed?

tdealtry:
{% gist 0efb539cd2ab425900e8 WCSim-Verification-PR89.pdf %}
<script src="https://gist.github.com/tdealtry/0efb539cd2ab425900e8.js"></script>

tdealtry:
[Verification plots](https://gist.github.com/tdealtry/0efb539cd2ab425900e8) Sorry for the delay on this; it was almost ready then we discovered the multi-digit problems & had to remake all the plots.
Everything* is there & ready to go (just done another merge with develop).
*I'll work on the .mac option swap today; assuming BeginOfEventAction() is called where I hope it is, it'll be fairly quick.

tdealtry:
So I've had to do exactly what I didn't want to do and add a '/DAQ/Construct' option, but this is only temporary (until the old WCSimWCDigitizer class is removed, which I believe is going to be ~directly after this PR is merged) and I've documented how to change things (in novis.mac, and various TODO comments in the code).
So now everything is ready to go.

src/WCSimSteppingAction.cc
eosullivan: Ah, just realizing this wasn't resolved. Let's do these changes to avoid compiler warnings in a separate PR. You could either go through and do them manually after this PR gets merged (kind of annoying, but less of a chance to screw things up), or there is a cherrypick command in git (https://ariejan.net/2010/06/10/cherry-picking-specific-commits-from-another-branch/) which should do what you want but I would recommend using caution (and maybe making a copy of your feature branch in case things go sideways).

vis.mac
```diff
@@ -111,7 +114,7 @@
 # Or you can use the G4 Particle Gun
 /mygen/generator normal
 /gun/particle mu+
-/gun/energy 1000 MeV
+/gun/energy 20 MeV
```
eosullivan: Revert this back to 1000 MeV as a default

eosullivan:
What issue did you run into that made the /DAQ/Construct necessary? We should try and think of another way to implement this, even if it is a short term thing. The /WCSim/Construct line has been really problematic.

eosullivan:
I just read through the verification documentation. Looks nice and thorough. Could you put the link in the PR description so it doesn't get buried in this discussion thread?

A couple of questions/comments after reading the documents:
* Sorry if this information exists somewhere, but could you just remind me why we are introducing multiple digits per PMT? Are multiple digits per PMT allowed in the real life SKI electronics? Is the fact that this feature didn't exist before a bug?  
* I'm a bit confused about the comparison in  Fig 13, which seems to suggest the difference should be in the high energy region since the new code accepts more single pes at high energy. The difference is in the low E region, though? Or am I not seeing some difference that's there? 
* How do you know it is right to make the trigger time = the first digit above the threshold rather than the first hit above the threshold?
 * It's nice to see that hard 25 hit cutoff for the low E events! It shows that this PR is fixing issue #48. 
* If I understand correctly, the only thing the SKIV digitzer does right now is introduce some deadtime. Furthermore, it isn't even it's own digitizer, but exists as an if statement in src/WCSimEventAction.cc. I wonder if it might be worth waiting for issue #13 to introduce the SKIV option since this incarnation isn't even really a placeholder for what will be done in the actual implementation of the SKIV digitizer. Or do you think it is useful somehow to have this option available now?

tdealtry:
The /DAQ/Construct line is to do with WCSimDarkRateMessenger - currently the instance is created in the constructor of WCSimWCAddDarkNoise or WCSimWCDigitizer (depending on whether you're running 'old' or 'new' code), and because the construction of these classes is delayed until after novis.mac is read (i.e. in WCSimEventAction::BeginOfEventLoop()), errors along the lines of "unknown .mac option" come up.
The way to get around this is to create the WCSimDarkRateMessenger instance somewhere else (e.g. in the WCSimEventAction constructor) and pass it to the classes that need it when they're created (i.e. the constructor of WCSimWCAddDarkNoise or WCSimWCDigitizer). This is more work, since we need to store the values as they're read in, instead of setting things in the relevant classes directly as is done now (see WCSimWCDAQMessenger methods SetTriggerOptions() and TellMeAboutTheTrigger(), and the Store* variables for an example of how this works).
I just thought it was a lot more work for something that is incredibly short term. Perhaps @cwwalter can comment, but the feeling I got when discussing in Kashiwa is that we'll merge this PR, and then about 24 hours later we'll merge another PR with the old class removed.

tdealtry:
@stewartt1982 can add the link to the pdf at the top.
* Re multiple digits per PMT: quote from the [SKI NIM](http://www-sk.icrr.u-tokyo.ac.jp/sk/pub/sknimpaper.pdf):  "There are two TACs and QACs for each channel so that events in rapid succession, such as a muon followed by its decay electron, can be processed without deadtime.'' So yes, it was a bug.
* Re Fig 13: remember the trigger efficiency is much lower at low energy in the new code (hence the normalisation differences here, and elsewhere (e.g. the timing plots))
* Re what's used to set trigger time: email discussion with Hayato-san: "Trigger timing is defined when the sum of hitsum signals exceeds the discriminator." i.e. the trigger time is the moment the trigger threshold is reached.
* I know a while ago @thomaslindner was interested in studying the effect of digitizer deadtime, so the concept is a useful one. Perhaps instead of the boolean switch between 0 & 500 ns, you'd prefer a new parameter (digitizer deadtime), defaulting to 0 ns (like SKI), and renaming the class WCSimWCDigitizerSKI (and getting rid of the SKIV option for now)?

eosullivan:
* re multiple digits per PMT: Ok, I'm having a hard time relating this multiple digit feature to deadtime, so let's see if I understand this. If there is only one digit per PMT, then once a PMT is hit that PMT is "dead" until the next event so there is deadtime? And by introducing multiple digits per PMT you are allowing for there to be no dead time? If so, should we only be allowing each PMT to have two digits? And how would this change when we go to SKIV?
* re Fig 13: So, should I see some difference in the high energy sample? Looking at the plots I can't see the difference. 
* re what's used to set trigger time: So...how does using the digitized time help this? Could we use the time of the 25th hit rather than the 26th hit and get the exact time? Isn't hitsum the raw hits?
* Love the idea of making digitizer deadtime a variable. Then it is super flexible when people want to put in their own options.

tdealtry:
* multiple digits. Assume you have hits spaced evenly at 0, 100, 200, ... and an integration window of 200 ns. So the first digit has two hits with times 0, 100. If you have no more than one digit per PMT, this is all you have. If you have no deadtime, you'll also get digits with 200;300, 400;500, .... If you have 500 ns deadtime, you'll get digits with 0;100, 700;800, ...
   * So this was for quite an unrealistic set of hits - if you check e.g. Fig 15, you see that most hits are in a narrow (~100 ns) peak, with a long (few 100 ns) tail. This tail is short enough however that, except at the very high energies, 500 ns deadtime is essentially equal to the 'infinite' deadtime in the old code
   * In SKIV, the trigger windows are not just 1350 ns (~ the SKIV value used for super low energy) but up to 100s of microseconds. So putting a limit on the number of digits per PMT per trigger window won't make sense.
* Fig 13 - you really need to zoom in on the first bin. At 5 GeV, the blue line is somewhat obscured by the axis tick unfortunately.
* Trigger times - I've just realised my explanation needs clarifying. It's not: the labelled eqn in the pdf (- trigger time + 950 + first hit time in digit + PMT- and charge-dependent smearing). It's in fact just 'first hit time in digit'. Does that make more sense?
   * From the SK NIM again: "When the PMT signal is above the threshold level [-1 mV ~ 1/4 p.e.], a ‘‘hit’’ signal with 200 ns width and 15 mV pulse height is asserted on the ATM front panel. The HITSUM signal, which is used to generate the global trigger signal, is the analog sum of all these 15mV/channel pulses"
   * To me this reads as it not being NHits, and potentially not quite being NDigits.
      * e.g. if two photons interact simultaneously, only one HITSUM signal will be generated. Of course, the PMT response isn't a delta function, and so "simultaneous" will extend out to a few 10s of ns.
      * e.g. if two photons interact at 0 and 180 ns, then you either get HITSUM of 15 mV from 0 to 380 ns, or 15 mV from 0 to 380 ns with an additional peak at 180-200 ns. How the HITSUM works (i.e. if a single channel can produce stacked 15 mV signals) is unclear.
* I'll add the deadtime variable tomorrow.

src/WCSimSteppingAction.cc
tdealtry: Would you mind explaining in a bit more detail why these changes should be taken out of this PR? It's likely to be an hour or twos work, for (from what I can see at the moment) no gain - whether the changes are merged here or later they'll still be in the develop branch & the next release.

tdealtry:
I've just committed the deadtime variable option. I also noticed that WCSimWCDAQMessenger wasn't setting defaults correctly, so I fixed that too.
I also renamed the new NHits triggers as NDigits triggers - this is what they actually are & the less chance for confusion between hits & digits the better.

src/WCSimSteppingAction.cc
eosullivan: Ok, I'm not looking to make wasted work for anyone, and we will have the next PR to focus on once this gets merged, so let's leave these compiler warning changes in this PR. But, for future pull requests here is what the philosophy should be: the changes that get made in the PR should be directly related to the issue or new feature that is being implemented. So, if you had found that there were compiler warnings, you would make a pull request just for that, we would have had a little discussion about what the warnings were and how you fixed them, and it would have been merged in awhile ago. Then, in the future, if I was wondering why this line was like it is, I would click the blame button, see which pull request that line was last changed in, and would see our discussion on just that topic rather than having to wade through the long digitizer discussion like is the case now. This is a small change that probably won't cause many questions in the future, but it's good to get in the habit of having focused pull requests and I think it makes for easier bookkeeping if nothing else.

tdealtry: Totally agree with everything you say. Thank you for making the exception (for my mistake) in this case though.

eosullivan:
Shouldn't some of the options that get set in novis/vis,mac be inherent properties of the digitizer or trigger? I like the flexibility to set them in the macro in case someone wants to do a study on the effect of each option, so don't undo the work you've done, but I want to avoid the case that we have now with the PMT classes where dark rate changes with PMT type and you have to manually change it in the macro to get it to match the photosensor you are using. I'm just trying to picture what we want to happen when we add in more configurations, like SKIV for example. I'm not so sure what the best breakdown is between what would go in the trigger and what would go in the digitizer, so let me know what you think, but maybe things like threshold, integration window, and deadtime could be set by the derived trigger class in the same way that properties of the photosensor like size, rn1pe, QE, etc are set in the derived PMT classes?  And then things that are related to how you want to store the information, like whether or not to save failed triggers, what time to give failed events, and whether or not to adjust for noise (or should this be a property of the trigger?) could stay set in the macro like they are.  You could leave the options that get set in the derived class in the macro, but comment them out and indicate that you could use these commands to overwrite the default values for the given configuration. How does that sound?

tdealtry:
I've just implemented this for the digitizer integration window & digitizer deadtime. Let me know what you think (it's slightly different to #101 because you can't call a pure virtual function in a constructor). If you like it, I'll do the same for some trigger options (e.g. window within which digits are saved).
Regarding changing the threshold in the .mac file - this is a function, rather than a single number, so making a new digitizer is probably the way to do it.
I've also set the default SKI digitizer integration time to 400 ns (reference: private communication with Hayato-san) but kept a 200 ns line in the .mac file uncommented, since 200 ns is what is used in WCSim currently (for at least 5 years according to git blame).

eosullivan:
Ok, I like how things are set up in src/WCSimWCDAQMessenger.cc and novis.mac, but in  src/WCSimWCDigitizerNew.cc, can we make it more like src/WCSimPMTObject.cc? You have it so each function (GetDefaultDeadTime(), GetDefaultIntegrationWindow(), etc) has an if statement to determine how to set the value based on digitizer type. In  src/WCSimPMTObject.cc, each PMT type has these as member functions. There are some disadvantages to this (it makes for a longer file, perhaps), but also has advantages (if I want to put in a new PMT everything I need is in one place and I can just cut and paste an example and fill in my own info). Plus, since we already have this framework for the PMTs it would be nice to have the same framework for the digitizer.

tdealtry:
It wasn't quite trivial (hence why I didn't do it this way in the first place) but it's done now. It's not identical to the WCSimPMTObject.cc method, because it's the WCSimWCDigitizer class that needs to know about the options (rather than an external class). The only negative is the GetVariables() must be called in the derived constructor, but I'll update WCSim/WCSim-Documentation#2 with this information when this pull request is complete.

eosullivan:
Great, this looks more in line with what I was picturing and I think this way will make it easier to add future options. So, please add in the trigger options in this way as well, if possible.

tdealtry:
Done. Just to be clear for what it works well for / developer needs to be careful for:
It works well for a trigger class that has a single algorithm (e.g. the NDigits class). It'll also work for classes with multiple algorithms (e.g. SKIV's WIT + NDigits trigger) called either consecutively (like NDigits2) or in a single algorithm (no example in the pull request).
If you use the same algorithm type multiple times in the trigger (e.g. SK has various NDigits thresholds giving SHE, HE, LE, ... triggers) you need to be careful. But sensible use of enumerations (& suitable additions to WCSimWCDAQMessenger) will make it work well. NDigits2 shows the use of the enumerations, but not the WCSimWCDAQMessenger additions (it's just a barebones example).

There is one more thing I want to do - the default trigger window of 200 ns in SK is set as the maximum straight line travel time for a photon across the detector. I'll make the default query the geometry to find this (I presume there's some methods in DetectorConstruction that'll make this number calculable).

tdealtry:
I've just made a commit so that the defaults are sensible, no matter which detector geometry or dark noise rate are used:
* default trigger window is set as the maximum straight line travel time for a photon across the detector (rounded to nearest 100 ns to match up with what SK use)
* default ndigits threshold is automatically adjusted for noise
So all the .mac-related comments/changes have now been addressed.

eosullivan:
I agree in principle that it might be useful to have the default trigger window be the max straight line travel time for a photon, but we should probably do that in a separate pull request. For now, let's just leave the default hard coded to be what it was prior to this PR and you can implement this new default after this is merged. That way, things won't change with this PR for people who are already using the big tanks.

tdealtry:
Done. I'll submit the other PR once this one is complete.

eosullivan:
Could you also back out the methods you committed to find the max travel time? Looking at the commit log https://github.com/stewartt1982/WCSim/commit/5960da6790a7c5f53bcd64bdccf77cdc4a2a3359 it looks contained to a few files. Keep this on a branch on your fork so you have it handy when you do it in the other PR. We just want to make sure we have all the changes related to this new default trigger window together in the same PR. Also make sure you change novis.mac so it doesn't say the default trigger time is the max travel time.

tdealtry:
Done!

include/WCSimWCDigitizerNew.hh
```diff
@@ -0,0 +1,107 @@
...
+public:
+  
+  WCSimWCDigitizerSKI(G4String name, WCSimDetectorConstruction*, WCSimWCDAQMessenger*);
+  ~WCSimWCDigitizerSKI();
+
+  void DigitizeHits(WCSimWCDigitsCollection* WCHCPMT);
+
+private:
+  int GetDefaultDeadTime()          { return 0; }   ///< SKI digitizer deadtime is 0 ns
+  int GetDefaultIntegrationWindow() { return 400; } ///< SKI digitizer integration window is 400 ns
```
eosullivan: Thinking about this a bit more since our meeting this morning, changing the default digitizer integration time should be it's own separate pull request. We should try and minimize small changes sneaking in with this issue so that people using the code aren't confused about why their output is suddenly different. Make an issue on GitHub, explain why the digitizer integration should be 400 instead of 200. Then, it should be a straightforward, one-line PR to submit and approve after this PR is merged.

tdealtry:
As a follow-up from the meeting yesterday (for legacy, see http://www-sk.icrr.u-tokyo.ac.jp/indico/getFile.py/access?resId=0&materialId=slides&confId=2509 for some final questions / answers on this PR)
* .mac files have been clarified
* comments have been moved to the correct location / deleted as necessary
* WCSim/WCSim-Documentation#2 has been updated
* I've run some 10,50,500 MeV electron particle guns (with & without dark noise), and compared results with those from the [verification plots](https://gist.github.com/tdealtry/0efb539cd2ab425900e8). All distributions plotted are identical, for both the new and old classes (after a fix for the old digitizer).

include/WCSimWCDigitizerNew.hh
tdealtry: Done

src/WCSimWCTrigger.cc
```diff
@@ -0,0 +1,523 @@
...
+
+	  //first smear the charge & time
+	  float peSmeared = (*WCDCPMT)[i]->GetPe(ip);
+	  float Q = (peSmeared > 0.5) ? peSmeared : 0.5;
+	  G4double digihittime = -triggertime
+	    + WCSimWCTriggerBase::offset
+	    + digit_time
+	    + PMT->HitTimeSmearing(Q);
+	  if(digihittime < 0)
+	    continue;
```
tdealtry: This continue statement will cause problems in the future; it was put here as this is what is done in the old digitizer. However if you think of the following situation (ignoring the smearing)
* pretrigger window -5 microseconds (which is the minimum for all SKIV trigger types, except SLE and relic)
* offset -1 microseconds (which is ~ the case for SKIV)
* this means the earliest digits (i.e. -5 microseconds before the trigger) will have a digit time of ~ -4 microseconds and will never get saved
* using the defaults there won't be any problems with unsaved digits (offset -950 ns, pretrigger window -400 ns)

So a question for @eosullivan - shall I remove this continue statement now or in a future PR?

src/WCSimWCTrigger.cc
```diff
@@ -0,0 +1,523 @@
...
+	  //hit in event window
+	  //add it to DigitsCollection
+
+	  //first smear the charge & time
+	  float peSmeared = (*WCDCPMT)[i]->GetPe(ip);
+	  float Q = (peSmeared > 0.5) ? peSmeared : 0.5;
+	  G4double digihittime = -triggertime
+	    + WCSimWCTriggerBase::offset
+	    + digit_time
+	    + PMT->HitTimeSmearing(Q);
```
tdealtry: We spoke about this yesterday - the fact that HitTimeSmearing() is done here (i.e. after the trigger decision is made) causes digits to migrate in/out of the 200 ns trigger decision window, meaning there are false triggers and triggers with the incorrect trigger time.
Looking into the code, it seems* that this hit time smearing describes the PMT time response (i.e. has nothing to do with the digitizer or trigger), and so should be added to WCSimWCPMT (and possibly WCSimWCAddDarkNoise too). This'll effect both the old and new digitizer/trigger classes, and so I'd recommend leaving this as is for now and sorting this out in a future PR. @eosullivan what do you think?
*WCSimPMTObject isn't documented, so I had to get what the function means from the [wiki](https://github.com/WCSim/WCSim/wiki/Tutorial:-How-to-add-your-own-photdetector). But poor documentation is another issue (#99)

eosullivan: Yes, I agree with you (both about the fix to this issue and about the need for better documentation for developers). Make an issue on the issue tracker so we remember to do this and make a pull request with the change after this pull request is merged in.

src/WCSimWCTrigger.cc
eosullivan: Ok, let's leave this for a future PR as well since it was also a problem in the old digitizer routine and doesn't really fit the mandate of this pull request. Make an issue on the tracker so we remember. I will have to think about why this continue was put in to begin with. I wonder if there is some situation where it is needed. We can discuss it more when the issue page gets created.

eosullivan:
All of my comments have been addressed and this looks ready to merge in to me. Do you agree @tdealtry and @stewartt1982?

tdealtry:
Ready to go on my end.

stewartt1982:
I'm fine with going ahead with the merge at this time.

eosullivan:
I updated the PR description to reflect some changes made during the review process.

PR #90, opened 2015-07-16
Uncommenting out the PMTCollEff on line and making it so that collection efficiency is on by default

eosullivan:
This pull request turns the PMT collection efficiency on in novis.mac and vis.mac and sets the default mode to use PMT collection efficiency in WCSimDetectorConstruction.cc.  

Currently, when you want to run the box and line (B&L) or hybrid photodector (HPD) PMTs, you have to both specify a detector configuration with B&L/HPD PMTs *and* uncomment a line in the macro which turns on collection efficiency. I think this extra step of enabling the collection efficiency in the macro is causing confusion and there is no need for it to be off by default since we are properly implementing this efficiency factor for all PMTs (100% for normal PMT, 95% for B&L, 98% for HPDs). Furthermore, since B&L PMTs are becoming the likely candidate for Hyper-K and turning on the collection efficiency is essential to having these PMTs work properly, someone should have to actively want to turn off collection efficiency for that to happen in the code. 

Using verificationhitchargetime.C, I have checked that this pull request should produce output which is identical to the code available on the current development branch for the default Super-K case. I have also run an event in the SuperK_20inchBandL_14perCent configuration and it was identical to the code available on the current development branch.

### DIFF
--- a/novis.mac
+++ b/novis.mac
@@ -60,7 +60,7 @@
 #/WCSim/PMTQEMethod     SensitiveDetector_Only
 
 #turn on or off the collection efficiency
-#/WCSim/PMTCollEff on
+/WCSim/PMTCollEff on
 
 # command to choose save or not save the pi0 info 07/03/10 (XQ)
 /WCSim/SavePi0 false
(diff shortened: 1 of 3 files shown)


cwwalter:
So.. since you are changing the default to be on, shouldn't we have the macro be 'on' but commented out?  That is the pattern for the other options.

eosullivan:
The standard seems to be to have it uncommented. For example, SavePi0 is false by default, but `/WCSim/SavePi0 false` is uncommented. Similarly, there is the line `/WCSim/PMTQEMethod Stacking_Only`, but this is the default. I think the only place where the standard is to comment out the default is for the detector configuration option.

cwwalter:
OK.

Then go ahead an merge this version.

Issue #91, opened 2015-07-16
Removing old DUSEL detector configurations


eosullivan:
In WCSimDetectorConfigs.cc, half of the current configurations are various DUSEL-related detectors. Is anyone using these for anything? I would like to remove them because each time I add a new option that must be specified in the WCSimDetectorConfigs file (a new way of choosing PMTs for example), I must add it to all configurations and then test each one to ensure I haven't introduced an error. With the upcoming addition of an OD to our detectors, perhaps now is the time to delete these defunct options so we only have the options we need/actively use? If anyone desperately wanted to recover the old configurations then they could look through a previous tagged version and fish out the details. Any complaints?

cwwalter:
Yes, please remove the old ones.  They are also one of the reason the .mac files is long and people miss the /construct line.

PR #96, opened 2015-07-20
Feature/issue91

eosullivan:
Removing old DUSEL configurations from the macros, as well as from the code. This is related to issue #91.


cwwalter:
That's a lot of stuff!

This all looks good.  Assuming things still run because nothing was missed this looks good to merge.

eosullivan:
Tested it and it still compiles and runs with novis.mac and vis.mac.

PR #97, opened 2015-07-21
Visualization

ntp4:
The visualization code allows users to inspect the geometry of WC detectors easier. A new visualizer called RayTracer was added and can be accessed by using the visRayTracer.mac macro file instead of the original vis.mac. The majority of the changes from the original files come in the form of visual attributes being added to cylinder and PMT volumes for the RayTracer visualizer.


jobOptions.mac
```diff
@@ -15,4 +15,6 @@
 /WCSim/physics/list WCSim # included for backward-compatibility.  Requires jobOptions2.mac
 
 
-
+## command to set visualization parameters in WCSimConstructPMT and WCSimConstructCylinder
+#/WCSim/WCVisChoice OGLSX
+/WCSim/WCVisChoice RayTracer
```
eosullivan: You should set this to OGLSX by default.

vis.mac
```diff
@@ -78,17 +78,19 @@
 #/DarkRate/SetConvert 1.119 #For HPDs
 #/DarkRate/SetConvert 1.126 #For Box and Line PMTs
 
-/vis/scene/create
-/vis/open OGLSX
+##Comment out lines the create, open OGLSX, zoom, viewpointThetaPhi, and upVector lines below if using RayTracer
+
+##/vis/scene/create
+##/vis/open OGLSX
```
eosullivan: Make sure vis.mac is still running OGLSX by default.

vis.mac
```diff
@@ -114,14 +110,30 @@
...
-/mygen/generator normal
-/gun/particle mu+
-/gun/energy 1000 MeV
-/gun/direction 0 1 0
-/gun/position 0 0 0 
+#/mygen/generator normal
+#/gun/particle mu+
+#/gun/energy 1000 MeV
+#/gun/direction 0 1 0
+#/gun/position 0 0 0
```
eosullivan: This /mygen and /gun stuff shouldn't be commented out.

vis.mac
```diff
@@ -78,11 +78,13 @@
...
 #/DarkRate/SetConvert 1.126 #For Box and Line PMTs
 
+##Comment out lines the create, open OGLSX, zoom, viewpointThetaPhi, and upVector lines below if using RayTracer
+
 /vis/scene/create
 /vis/open OGLSX
 # You can also set the size and position of the window if you like
 #/vis/open OGLSX 1000x1000-0+0
-/vis/ogl/set/printSize 1000 1000
+#/vis/ogl/set/printSize 1000 1000
```
eosullivan: Looking at the diff, this line should be uncommented if we want it to be the same as before this pull request.

eosullivan:
I would like to propose that this becomes its own macro (separate from vis.mac). In order to get Raytracer to work, you have to comment out almost every line of vis.mac and uncomment out the Raytracer lines. It would be cleaner and easier to run if it exists as a separate .mac file.

jobOptions.mac
```diff
@@ -13,6 +13,3 @@
 #/WCSim/physics/list QGSP_BIC_HP #preferred for energies below 5 GeV
 #/WCSim/physics/list LBE # preferred for low-background experiments.  
 /WCSim/physics/list WCSim # included for backward-compatibility.  Requires jobOptions2.mac
-
-
-
```
eosullivan: jobOptions.mac shows up in the diff of this pull request because there is some missing white space. Cut and paste the old jobOptions.mac here so it doesn't show up in the diff.

vis.mac
```diff
@@ -78,6 +78,8 @@
 #/DarkRate/SetConvert 1.119 #For HPDs
 #/DarkRate/SetConvert 1.126 #For Box and Line PMTs
 
+##Comment out lines the create, open OGLSX, zoom, viewpointThetaPhi, and upVector lines below if using RayTracer
```
eosullivan: This comment is left over from the last commit. Remove it and make sure vis.mac doesn't show up in the diff since I don't think anything is changed here anymore.

visRayTracer.mac
```diff
@@ -0,0 +1,101 @@
...
+/vis/rayTracer/eyePosition -15 0 -15
+/vis/rayTracer/trace g4RayTracer.jpeg
+
+/vis/rayTracer/eyePosition 70 1 70
+/vis/rayTracer/trace g4RayTracer1.jpeg
+
+/vis/rayTracer/eyePosition 70 1 20
+/vis/rayTracer/trace g4RayTracer2.jpeg
+
+#exit
```
eosullivan: Most of the lines here are unnecessary to run raytracer. I think the only lines that are needed are 1-9 and 84-99. You could also leave in lines 11,12, and 16 to show people how to change the detector configuration, but it isn't required to list them all here. Remove all of the other lines.

visRayTracer.mac
```diff
@@ -0,0 +1,33 @@
...
+/hits/verbose 0
+
+##select RayTracer Visualization
+/WCSim/WCVisChoice RayTracer
+
+## select the geometry
+# Default config if you do nothing is currently SuperK
+#/WCSim/WCgeom SuperK
+
+# Some other SuperK options:
```
eosullivan: This comment on line 15 should be removed.

Also, there should be the /WCSim/Construct line here. Make sure you put a comment beside it that says that this line is needed for raytracer to work (people are used to uncommenting this only if they want to switch the configuration to something other than SK, but in this case it is needed for all detector configs).

vis.mac
```diff
@@ -81,8 +81,8 @@
 
 # Vis Settings for SK
 /vis/viewer/zoom 1.2
-#/vis/viewer/set/viewpointThetaPhi 45 45 deg
-/vis/viewer/set/viewpointThetaPhi 110 0 deg
+/vis/viewer/set/viewpointThetaPhi 45 45 deg
+#/vis/viewer/set/viewpointThetaPhi 110 0 deg
```
eosullivan: vis.mac is still showing up in the diff, I think because you have the wrong /vis/viewer/set/viewpointThetaPhi uncommented. Make sure the 45 45 one is commented and the 110 0 is uncommented (opposite to what you currently have here).

visRayTracer.mac
```diff
@@ -0,0 +1,32 @@
+# Visualization and setup macro 
+# This file is run by default if you just type WCSim and nothing else.
```
eosullivan: Replace the comments on lines 1 and 2 with relevant ones for visRayTracer.mac. This one isn't run by default in WCSim. Put something like, "This macro is used to visualize geometries using Raytracer. The output of this macro is three .jpeg files visualizing the geometry from three different viewpoints."

visRayTracer.mac
```diff
@@ -0,0 +1,32 @@
...
+
+/WCSim/Construct # This must be uncommented to use RayTracer
+
+## Errors may arise if using RayTracer and attempting to view the detector from a distance further than 40 meters
+## If this occurs, attempt changing the y-direction
+
+/vis/open RayTracer
+/vis/viewer/set/upVector 0 0 1
+
+/vis/rayTracer/eyePosition -15 0 -15
```
eosullivan: Put a brief comment beside these eyePosition lines saying inside the ID looking toward the top corner, etc

eosullivan:
I checked out this pull request today and played around with visualizing different configurations with raytracer. It's working well and is really easy to use! Once you've made these last few small changes that I indicated on the line notes, this can get merged in.

eosullivan:
Ok, looks good. I am merging this in. Thanks for your work this summer, Nathan!

PR #98, opened 2015-07-22
Adding a correction factor to compensate for the 73% collection efficiency of the R3600

YujiOkajima:
In order to set the collection efficiency of R3600 to 73%, the effective QE in MC process was changed to 1/0.73 times bigger.
from the uploaded slide on the URL below, the modification seems to be correctly introduced.

https://dl.dropboxusercontent.com/u/61208795/Collection_efficiency_modification.pdf

The figure below shows the total number of charge and hit comparison between before and after modification.
comparison_r3600_version

The figure below shows the total number of charge and hit comparison between geometry with R3600  and Box and Line PMT after CE modification.

10 MeV
compare_r3600_boxline_10mev

500 MeV
compare_r3600_boxline_500mev

From these figure, the validation has been finished.

### DIFF
--- a/include/WCSimPMTObject.hh
+++ b/include/WCSimPMTObject.hh
@@ -47,8 +47,6 @@ public:
   G4float  GetmaxQE();
   float    HitTimeSmearing(float);
   G4double GetPMTGlassThickness();
-protected:
-  G4float* GetCollectionEfficiencyArray();
 };
 
 class PMT8inch : public WCSimPMTObject
--- a/include/WCSimPMTObject.hh
+++ b/include/WCSimPMTObject.hh
@@ -47,6 +47,8 @@ public:
   G4float  GetmaxQE();
   float    HitTimeSmearing(float);
   G4double GetPMTGlassThickness();
+protected:
+  G4float* GetCollectionEfficiencyArray();
 };
 
 class PMT8inch : public WCSimPMTObject
--- a/src/WCSimPMTObject.cc
+++ b/src/WCSimPMTObject.cc
@@ -1343,7 +1343,7 @@ G4float* BoxandLine20inchHQE::GetQEWavelength(){
 }
 
 G4float* BoxandLine20inchHQE::GetQE(){
-  G4float correctionFactor = 1./0.73;//Correction factor added in July 2015 to scale the output of B&L PDs to 2.3 times the 20" PMTS based on Hamamatsu measurement. This was done in Pull Request #98 and will be removed once a more permanent solution is found.
+  G4float correctionFactor = 1./0.73;//Correction factor added in July 2015 to scale the output of B&L PDs to 2.32 times the 20" PMTS based on Hamamatsu measurement. This was done in Pull Request #98 and will be removed once a more permanent solution is found.
   static G4float QE[20] =
     { 0.00*correctionFactor, .0008*correctionFactor, .1255*correctionFactor, .254962*correctionFactor, .2930*correctionFactor, .3127*correctionFactor, .3130*correctionFactor, .2994*correctionFactor, .2791*correctionFactor, .2491*correctionFactor,
       .2070*correctionFactor,  .1758*correctionFactor, .1384*correctionFactor, .0779*correctionFactor, .0473*correctionFactor, .0288*correctionFactor, .0149*correctionFactor, .0062*correctionFactor, .0002*correctionFactor, .0001*correctionFactor};  
@@ -1351,7 +1351,7 @@ G4float* BoxandLine20inchHQE::GetQE(){
   return QE;
 }
 G4float BoxandLine20inchHQE::GetmaxQE(){
-  G4float correctionFactor = 1./0.73;//Correction factor added in July 2015 to scale the output of B&L PDs to 2.3 times the 20" PMTS based on Hamamatsu measurement. This was done in Pull Request #98 and will be removed once a more permanent solution is found.
+  G4float correctionFactor = 1./0.73;//Correction factor added in July 2015 to scale the output of B&L PDs to 2.32 times the 20" PMTS based on Hamamatsu measurement. This was done in Pull Request #98 and will be removed once a more permanent solution is found.
   const G4float maxQE = 0.315*correctionFactor;
   return maxQE;
 }
@@ -1516,7 +1516,7 @@ G4float* BoxandLine12inchHQE::GetQEWavelength(){
 }
 
 G4float* BoxandLine12inchHQE::GetQE(){
-  G4float correctionFactor = 1./0.73;//Correction factor added in July 2015 to scale the output of B&L PDs to 2.3 times the 20" PMTS based on Hamamatsu measurement. This was done in Pull Request #98 and will be removed once a more permanent solution is found.
+  G4float correctionFactor = 1./0.73;//Correction factor added in July 2015 to scale the output of B&L PDs to 2.32 times the 20" PMTS based on Hamamatsu measurement. This was done in Pull Request #98 and will be removed once a more permanent solution is found.
   static G4float QE[20] =
     { 0.00*correctionFactor, .0008*correctionFactor, .1255*correctionFactor, .254962*correctionFactor, .2930*correctionFactor, .3127*correctionFactor, .3130*correctionFactor, .2994*correctionFactor, .2791*correctionFactor, .2491*correctionFactor,
       .2070*correctionFactor,  .1758*correctionFactor, .1384*correctionFactor, .0779*correctionFactor, .0473*correctionFactor, .0288*correctionFactor, .0149*correctionFactor, .0062*correctionFactor, .0002*correctionFactor, .0001*correctionFactor};  
@@ -1524,7 +1524,7 @@ G4float* BoxandLine12inchHQE::GetQE(){
   return QE;
 }
 G4float BoxandLine12inchHQE::GetmaxQE(){
-  G4float correctionFactor = 1./0.73;//Correction factor added in July 2015 to scale the output of B&L PDs to 2.3 times the 20" PMTS based on Hamamatsu measurement. This was done in Pull Request #98 and will be removed once a more permanent solution is found.
+  G4float correctionFactor = 1./0.73;//Correction factor added in July 2015 to scale the output of B&L PDs to 2.32 times the 20" PMTS based on Hamamatsu measurement. This was done in Pull Request #98 and will be removed once a more permanent solution is found.
   const G4float maxQE = 0.315*correctionFactor;
   return maxQE;
 }
--- a/src/WCSimPMTObject.cc
+++ b/src/WCSimPMTObject.cc
@@ -213,19 +213,11 @@ G4float* PMT20inch::GetQEWavelength(){
 
 G4float* PMT20inch::GetQE(){  
   static G4float QE[20] = { 0.00, .0139, .0854, .169, .203, .206, .211, .202,.188, .167, .140, .116, .0806, .0432, .0265, .0146, .00756, .00508, .00158, 0.00};
-
-
   return QE;
 }
 G4float PMT20inch::GetmaxQE(){
   const G4float maxQE = 0.211;
   return maxQE;
-}
-
-G4float* PMT20inch::GetCollectionEfficiencyArray(){
-  static G4float CE[10] = { 100., 100., 100., 100., 100., 100., 100., 100., 100., 100.};//temporary value. It will be changed in future.
-  //  static G4float CE[10] = { 73., 73., 73., 73., 73., 73., 73., 73., 73., 73.};
-  return CE;
 }
 
 
@@ -1352,7 +1344,6 @@ G4float* BoxandLine20inchHQE::GetQEWavelength(){
 
 G4float* BoxandLine20inchHQE::GetQE(){
   G4float correctionFactor = 1./0.73;//Correction factor added in July 2015 to scale the output of B&L PDs to 1.9 times the 20" PMTS based on Hamamatsu measurement. This was done in Pull Request #98 and will be removed once a more permanent solution is found.
-  // set R3600 C.E. to 73% and not to change the R3600 output.
   static G4float QE[20] =
     { 0.00*correctionFactor, .0008*correctionFactor, .1255*correctionFactor, .254962*correctionFactor, .2930*correctionFactor, .3127*correctionFactor, .3130*correctionFactor, .2994*correctionFactor, .2791*correctionFactor, .2491*correctionFactor,
       .2070*correctionFactor,  .1758*correctionFactor, .1384*correctionFactor, .0779*correctionFactor, .0473*correctionFactor, .0288*correctionFactor, .0149*correctionFactor, .0062*correctionFactor, .0002*correctionFactor, .0001*correctionFactor};  
@@ -1360,7 +1351,7 @@ G4float* BoxandLine20inchHQE::GetQE(){
   return QE;
 }
 G4float BoxandLine20inchHQE::GetmaxQE(){
-  G4float correctionFactor = 1./0.73;//Correction factor added in July 2015 to set R3600 C.E. to 73% and not to change the R3600 output.
+  G4float correctionFactor = 1./0.73;//Correction factor added in July 2015 to scale the output of B&L PDs to 1.9 times the 20" PMTS based on Hamamatsu measurement. This was done in Pull Request #98 and will be removed once a more permanent solution is found.
   const G4float maxQE = 0.315*correctionFactor;
   return maxQE;
 }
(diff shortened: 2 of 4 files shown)


eosullivan:
Thanks for doing this verification, Okajima-san! I like that you attach slides with detailed information, but could you also put the key plots directly in this pull request? I am worried that if you ever clean your Dropbox folder, all of this information will be lost.

eosullivan:
Because we are applying this new 1/0.73 factor to all PDs through the QE, we need to think about more than just the 20" PMTs and the B&Ls. I assume this 73% CE should also be applied to the other PMTs (the 8", 10", 12" PMTs)? Also, do we have an expectation for the relative yield between R3600 and the HPDs? Does the output match there now that we have also increased this factor?

src/WCSimStackingAction.cc
```diff
@@ -23,13 +23,13 @@ G4ClassificationOfNewTrack WCSimStackingAction::ClassifyNewTrack
   G4String WCIDCollectionName = DetConstruct->GetIDCollectionName();
   G4ClassificationOfNewTrack classification    = fWaiting;
   G4ParticleDefinition*      particleType      = aTrack->GetDefinition();
-  
+  G4float correctionfactor = 1/0.73; //Collection factor for 20inchPMT CE setting
```
eosullivan: The comment on the correctionfactor should be clearer so that we remember why we did this. Something like "Correction factor added in July 2015 to account for the measured relative pe number differences between R3600 and box and line PDs."

eosullivan:
@YujiOkajima, would it be possible to fix our problem by increasing the CE of the B&L rather than applying this factor to the ratio? Then, at least this temporary fix is only in one place. So, something like:
```C++
G4float* BoxandLine20inchHQE::GetCollectionEfficiencyArray(){
static G4float CE[10] = { 95., 95., 95., 95., 95., 95., 95., 95., 95., 95.};
G4float correctionFactor = 1/0.73; //Correction factor added in July 2015 to account for the measured relative pe number differences between R3600 and box and line PDs. Remove once this is fixed.
return CE*correctionFactor;
}
```

YujiOkajima:
@eosullivan 
The 73% CE of R3600 is based on Hamamatsu evaluation value, independent value from this correction.
And it will effect not only Box and Line PMT, but also any other photodetectors.

And CE is too high to correct by 1/0.73.
When CE of Box & Line PMT was multiplied by 1/0.73, the CE will be higher than 1.

cwwalter:
This should be a temporary change only until we fix it correctly.  So, if we can we want to make a change in just one file.  We don't want to make many changes and we only want to do this for the B&L so we can start the real studies right away.

We know that the CE will be unphysical (greater than 1) but will we get the correct results for B&L?

YujiOkajima:
When CE is set to > 1, the all photoelectron emitted from photocathode (QE process) will pass the process about collection efficiency, but the number of photoelectron will not be increase.
The QE and CE process are independent from each other.
Therefore the maximum limit of CE is 1.
Even if we change only one file, we should not correct CE value, but QE value.

I will try to modify this code to change only one file.

YujiOkajima:
I modified the code and validated with previous requested version.
10 MeV electron was generated in Super-K geometry with R3600 and Box and Line PMT (randomly distributed in FV and isotropic direction)
and compared by using verification_HitChargeTime.C

R3600 (coverage: 40%)
compare_10mev_events

Box and Line PMT (coverage: 20%)
compare_10mev_events_bl

In the both geometries, there is no difference in the number of hit and p.e. between the first events of MC with previous and current requested version WCSim.

eosullivan:
I think this second commit looks worse than the first commit if our goal is to minimize the number of times this correctionFactor shows up. It's nice that it's contained to a single file (WCSimPMTObject), but it's in a lot of places in this file. My preference is to use the first way you did it because I think it will be easier to remove it from there once we have a permanent solution worked out. What do you think @cwwalter?

src/WCSimPMTObject.cc
```diff
@@ -212,14 +212,24 @@ G4float* PMT20inch::GetQEWavelength(){
 }
 
 G4float* PMT20inch::GetQE(){  
-  static G4float QE[20] = { 0.00, .0139, .0854, .169, .203, .206, .211, .202,.188, .167, .140, .116, .0806, .0432, .0265, .0146, .00756, .00508, .00158, 0.00};
+  G4float correctionFactor = 1./0.73;//Correction factor added in July 2015 to set R3600 C.E. to 73% and not to change the R3600 output.
+  static G4float QE[20] = { 0.00*correctionFactor, .0139*correctionFactor, .0854*correctionFactor, .169*correctionFactor, .203*correctionFactor, .206*correctionFactor, .211*correctionFactor, .202*correctionFactor,.188*correctionFactor, .167*correctionFactor, .140*correctionFactor, .116*correctionFactor, .0806*correctionFactor, .0432*correctionFactor, .0265*correctionFactor, .0146*correctionFactor, .00756*correctionFactor, .00508*correctionFactor, .00158*correctionFactor, 0.00*correctionFactor};
+  //  static G4float QE[20] = { 0.00, .0139, .0854, .169, .203, .206, .211, .202,.188, .167, .140, .116, .0806, .0432, .0265, .0146, .00756, .00508, .00158, 0.00};
+
+
   return QE;
```
eosullivan: If we are going to do it this way, you should just return QE*correctionFactor rather than putting it in the array.

YujiOkajima:
The GetQE returns the pointer of QE array.
Therefore I think QE*correctionFactor can not be done.

eosullivan:
Ok, I think Chris and I have reached an agreement about what we think is the best thing to do. Keep what you have in this most recent pull request, but make the following modifications:
- Don't put the 73% collection efficiency into the 20" PMT. We will make this change later, once we have a permanent solution. 
- Because of this, also don't include the correctionFactor in the 20" PMT and remove this new GetCollectionEfficiencyArray() method. 
- Change the comment for the correctionFactor so it is more descriptive. It's ok if it's long. Like this: //Correction factor added in July 2015 to scale the output of B&L PDs to 1.9 times the 20" PMTs based on Hamamatsu measurements. This was done in Pull Request #98 and will be removed once a more permanent solution is found. 
- Make it so that StackingAction doesn't show up in the diff of this PR. There is no change there, but there is an extra space so it is showing up as a modified file. Either remove that extra space or replace the file with the old version so we can keep our history clean.

Thanks for doing this work, Okajima-san!

eosullivan:
Also, please rerun the verification as a check once you have made these changes.

YujiOkajima:
Fixed the WCSim code and rerun the verification with electrontest.mac.
And confirmed the R3600 output was completely the same as that of current version WCSim.
compare_r3600_new

include/WCSimPMTObject.hh
```diff
@@ -47,6 +47,8 @@ public:
   G4float  GetmaxQE();
   float    HitTimeSmearing(float);
   G4double GetPMTGlassThickness();
+protected:
+  G4float* GetCollectionEfficiencyArray();
```
eosullivan: Remove lines 50 and 51. We should aim to make minimal changes and there is no need to add GetCollectionEfficiencyArray(); to the 20" PMTs in this pull request.

src/WCSimPMTObject.cc
```diff
@@ -213,6 +213,8 @@ G4float* PMT20inch::GetQEWavelength(){
 
 G4float* PMT20inch::GetQE(){  
   static G4float QE[20] = { 0.00, .0139, .0854, .169, .203, .206, .211, .202,.188, .167, .140, .116, .0806, .0432, .0265, .0146, .00756, .00508, .00158, 0.00};
+
+
```
eosullivan: Remove extra white space in lines 216 and 217.

src/WCSimPMTObject.cc
```diff
@@ -220,6 +222,12 @@ G4float PMT20inch::GetmaxQE(){
   return maxQE;
 }
 
+G4float* PMT20inch::GetCollectionEfficiencyArray(){
+  static G4float CE[10] = { 100., 100., 100., 100., 100., 100., 100., 100., 100., 100.};//temporary value. It will be changed in future.
+  //  static G4float CE[10] = { 73., 73., 73., 73., 73., 73., 73., 73., 73., 73.};
+  return CE;
+}
+
```
eosullivan: Remove lines 225-230. We should aim to make minimal changes and is no need to add GetCollectionEfficiencyArray(); to the 20" PMTs in this pull request.

src/WCSimPMTObject.cc
```diff
@@ -1343,13 +1351,17 @@ G4float* BoxandLine20inchHQE::GetQEWavelength(){
 }
 
 G4float* BoxandLine20inchHQE::GetQE(){
+  G4float correctionFactor = 1./0.73;//Correction factor added in July 2015 to scale the output of B&L PDs to 1.9 times the 20" PMTS based on Hamamatsu measurement. This was done in Pull Request #98 and will be removed once a more permanent solution is found.
+  // set R3600 C.E. to 73% and not to change the R3600 output.
```
eosullivan: Remove this comment. It doesn't apply now that we aren't changing R3600 to 73%.

src/WCSimPMTObject.cc
```diff
@@ -1343,13 +1351,17 @@ G4float* BoxandLine20inchHQE::GetQEWavelength(){
...
-    { 0.00, .0008, .1255, .254962, .2930, .3127, .3130, .2994, .2791, .2491,
-      .2070,  .1758, .1384, .0779, .0473, .0288, .0149, .0062, .0002, .0001};  
+    { 0.00*correctionFactor, .0008*correctionFactor, .1255*correctionFactor, .254962*correctionFactor, .2930*correctionFactor, .3127*correctionFactor, .3130*correctionFactor, .2994*correctionFactor, .2791*correctionFactor, .2491*correctionFactor,
+      .2070*correctionFactor,  .1758*correctionFactor, .1384*correctionFactor, .0779*correctionFactor, .0473*correctionFactor, .0288*correctionFactor, .0149*correctionFactor, .0062*correctionFactor, .0002*correctionFactor, .0001*correctionFactor};  
+
   return QE;
 }
 G4float BoxandLine20inchHQE::GetmaxQE(){
-  const G4float maxQE = 0.315;
+  G4float correctionFactor = 1./0.73;//Correction factor added in July 2015 to set R3600 C.E. to 73% and not to change the R3600 output.
```
eosullivan: Remove this comment. It doesn't apply now that we aren't changing R3600 to 73%. Replace it with the comment on line 1354 //Correction factor added in July 2015 to scale the output of B&L PDs to 1.9 times the 20" PMTS based on Hamamatsu measurement. This was done in Pull Request #98 and will be removed once a more permanent solution is found.

src/WCSimPMTObject.cc
```diff
@@ -1513,13 +1525,16 @@ G4float* BoxandLine12inchHQE::GetQEWavelength(){
...
-    { 0.00, .0008, .1255, .254962, .2930, .3127, .3130, .2994, .2791, .2491,
-      .2070,  .1758, .1384, .0779, .0473, .0288, .0149, .0062, .0002, .0001};  
+    { 0.00*correctionFactor, .0008*correctionFactor, .1255*correctionFactor, .254962*correctionFactor, .2930*correctionFactor, .3127*correctionFactor, .3130*correctionFactor, .2994*correctionFactor, .2791*correctionFactor, .2491*correctionFactor,
+      .2070*correctionFactor,  .1758*correctionFactor, .1384*correctionFactor, .0779*correctionFactor, .0473*correctionFactor, .0288*correctionFactor, .0149*correctionFactor, .0062*correctionFactor, .0002*correctionFactor, .0001*correctionFactor};  
+
   return QE;
 }
 G4float BoxandLine12inchHQE::GetmaxQE(){
-  const G4float maxQE = 0.315;
+  G4float correctionFactor = 1./0.73;//Correction factor added in July 2015 to set R3600 C.E. to 73% and not to change the R3600 output.
```
eosullivan: Remove this comment. It doesn't apply now that we aren't changing R3600 to 73%. Replace it with the comment on line 1354 //Correction factor added in July 2015 to scale the output of B&L PDs to 1.9 times the 20" PMTS based on Hamamatsu measurement. This was done in Pull Request #98 and will be removed once a more permanent solution is found.

eosullivan:
Hi Okajima-san. I made a few more in-line comments on this pull request. I am being extra picky with the commenting of this pull request because this correctionFactor will probably be confusing for new people (or even just people who weren't at the software meeting last week). I also don't want the collection efficiency to be explicitly set to 100% for the 20" PMT, mostly because there is no need to change the collection efficiency of the 20" PMT in this pull request, but also because it doesn't make sense to do it this way when the collection efficiency is set to 100% by default.

eosullivan:
Also, thanks for doing the verification that there is no difference in the 20" PMT after this PR. Could you also do a quick check to make sure we are still seeing the appropriate increase for the B&L PDs that we would expect (that 1.9x number) using this method of increasing the QE?

src/WCSimPMTObject.cc
```diff
@@ -1513,13 +1525,16 @@ G4float* BoxandLine12inchHQE::GetQEWavelength(){
 }
 
 G4float* BoxandLine12inchHQE::GetQE(){
+  G4float correctionFactor = 1./0.73;//Correction factor added in July 2015 to set R3600 C.E. to 73% and not to change the R3600 output.
```
eosullivan: Remove this comment on line 1528. It doesn't apply now that we aren't changing R3600 to 73%. Replace it with the comment on line 1354 //Correction factor added in July 2015 to scale the output of B&L PDs to 1.9 times the 20" PMTS based on Hamamatsu measurement. This was done in Pull Request #98 and will be removed once a more permanent solution is found.

eosullivan:
Ok, I think the code looks ready to go. Once @YujiOkajima confirms that we still see this 1.9x difference between 20" PMT and 20" B&L, then it should be ready to merge.

YujiOkajima:
Modified WCSim code and checked the MC output.
10MeV electrons were generated uniformly in FV on isotropic direction in Super-K with R3600 and Box and Line PMT and compared their MC output.
compare_output_20150728
From this figure and table, the number of p.e. before digitization ratio of B&L PMT to R3600 is almost the same as expected value from parameters in WCSim.

On the other hands, the number of digitized hit ratio from WCSim is 2.19.
I modified Nishimura-san's slide following to WCSim setting.
hit_eff_hamamatsu
Because the QE ratio of HQE photocathode to normal QE photocathode in WCSim is ~1.5, 
the expected number of digitized hit ratio from Hamamatsu measurement is 2.0-2.1  (WCSim: 2.19)
So WCSim output agreed with the Hamamatsu data within 5%-10% relatively.

eosullivan:
Why is the QE ratio in WCSim different than the one in Nishimura-san's slide? I don't understand why the QExCExHE is now 2.10 rather than 1.92.

YujiOkajima:
Maybe it comes from the difference of source.
In WCSim code, QE of R3600 is the same as that of SKDetSim and QE of Box & Line PMT is based on the average QE of 8 high QE R3600 in EGADS.

Nishimura-san set the QE of Box & Line PMT to 30%, and that of R3600 to 22%.
I heared the QE of Box & Line PMT (30%) is based on the Hamamatsu evaluation with high statistics.
And the source of R3600 QE (22%) is not known.

Because of the QE difference, the ratio is changed to 1.92 x (1.5(as QE ratio in WCSim))/(30%/22%)=2.1

I described this QE difference in local Hyper-K meeting and concluded that modification of QE is not necessary for now.

eosullivan:
What's the Hamamatsu measured ratio between the B&L and R3600? Is it 1.70? If we are getting a ratio of 2.90 from WCSim and the measured value is 1.70, then we aren't doing so well. Should we change the correction factor we are using to make the data match the simulation?

YujiOkajima:
the measured ratio of Hamamatsu is 1.78(1.70).
And the averaged QE of measured PMT is 28.9%(High QE Box and Line PMT) and 22.9%(Normal QE R3600).
So when we correct the QE with multiplying the ratio by 1.189 (1.5/(28.9%/22.9%))  , the expected ratio becomes 2.1(2.0).
It seems to agree with MC output of WCSim (ratio: 2.19)
Therefore I think the correction factor is working well.

eosullivan:
Thanks for your patient explanations Okajima. I think one of my biggest problems is not understanding where the numbers are coming from. 

You say that the QE difference between B&L and PMT in WCSim is currently a factor of 1.5 (before this pull request)? I just checked and this is approximately true for max QE. So, if I want to calculate what is currently in WCSim (before this pull request), I will take the ratio = Eff_B&L(QE,CE,HE)/Eff_PMT(QE,CE,HE) = 1.5x0.95x0.856/1x1x0.717=1.70.

Then, in the above slide from Hamamatsu, it says the ratio of total measured efficiency = 1.70, which I deduce from the y-axis of the graph to also be the ratio of QExCExHE.

So, without making any of the changes in this pull request, doesn't our output match the measurement exactly?

YujiOkajima:
@eosullivan 
As I described in today's meeting, the quantum efficiency of High QE Box & Line PMT in Hyper-K is possible to be different from the QE of the PMTs used for Hamamatsu 1p.e. counting measurement.
For example, QE of Box & Line PMT is expected to be 30%.

eosullivan:
Ok, we had an email discussion and I think I understand now. I will summarize below so I don't get confused when I revisit this later.

It seems like we can ignore the plot of the Hamamatsu measurement and their measured ratio of 1.7. Based on Hamamatsu simulations, the collection efficiency for the R3600 is 73%. We have factors in WCSim that suppress the pe output and so makes the R3600 simulations correct, but since we now include other photosensors (ex. box and line), we need to correct by the new photosensors by the 1/0.73 factor.
this correction factor of 1/0.73 to WCSim to compensate for the tuning
factors in the code that are applied to all tubes (and thus artificially
decreasing the pe output of the B&L). If we look at what pe ratio (QExCExHE)
we should expect for B&L to R3600 in WCSim, we should get 2.32 as per your
calculation below:

(0.315*0.95*0.856/0.75*0.73)/(0.211*0.73*0.717/0.75*0.73)=2.32

eosullivan:
Ok, so the code looks pretty ready to go. Can you update the comments to say that we expect a ratio of 2.32 instead of 1.9? 

I will also update the title of this pull request since we aren't updating the collection efficiency of the R3600 in this PR.

eosullivan:
@YujiOkajima, not sure if you saw my post in the flurry of emails yesterday, but once you update your comments we can merge in your pull request. This is the last thing we are waiting on for this tag, so the sooner you can get to it the better. The comment should be:
//Correction factor added in July 2015 to scale the post-digitization pe output of B&L PDs to 2.32 times the 20" PMTs based on Hamamatsu simulations. This was done in Pull Request #98 and will be removed once a more permanent solution is found.

YujiOkajima:
Sorry for delayed modification.
Now I fixed the comment.

YujiOkajima:
@eosullivan 
I fixed the comment based on your comment

Issue #84, opened 2015-06-14
Move dark rate and convert rate from the macro into WCPMTObject


eosullivan:
This shouldn't have to be set manually by the macro. Since it is a property of the PMt, it should be automatically done when the PD type is changed. This will also need to be done for the OD where the dark rate will be different than the ID.

cwwalter:
We should find out if the dark rate is really just the intrinsic dark rate of the PMT or if it also somehow reflects the radioactive activity in the tank.

eosullivan:
Well, it is currently set differently in the macro for different PD types, so there must be at least some component that is tied to the PD.

tfeusels:
@eosullivan: Is this something that needs help or is are you waiting for proper values for each PD type? SK PMT is 4.2 kHz, 20in B&L seems to converge to 10 kHz. Actually has this been discussed which values will be used as defaults for the reconstruction tuning and physics studies? 

Ideally, we should add the radioactive activity in the tank as separate noise source but I believe this is not even done in skdetsim. Maybe separate issue to open up?

eosullivan:
I intended to do it myself but haven't had any time to even start it, so if you are volunteering I can reassign this to you. Let me know.

My original idea was pretty simple: we are currently changing both the dark rate and the convert rate in the macro depending on which photosensor we are using and I would like it if it automatically changed with the photosensor type to reduce the probability of user error. This could be easily achieved if we move these variables to be a property in WCSimPMTObject. We could also leave in the option to set the value in the macro if someone wanted to increase the dark rate because of their particular configuration. 

The defaults for reconstruction and physics studies are the values in the macros (4.2 kHz, 8.4 kHz, or 3.0 kHz and the corresponding conversion rate depending on the PD type). 

As for the issue of incorporating the radioactivity in the tank, I think it's a further enhancement to what we are doing and therefore a separate issue (though perhaps @cwwalter disagrees).

Also, the dark rate stuff is changing around in pull request #89. I think if it's a small change of setting the default dark rate to be a value in WCSimPMTObject then it would be simple to coordinate with this pull request, though we would need to make sure @tdealtry was happy with it and how it fits in with his new changes.

tfeusels:
I was also thinking of changing it the way you proposed. Simple first: adding it in WCSimPMTObject, while keeping the possibility to change it globally in the macro.

[...]

PR #101, opened 2015-07-27
Feature/issue84

tfeusels:
This is the quick implementation of DarkRate for each PMT type as it is a property of the PMT to address issue #84. Most DarkRate values in WCSimPMTObject are motivated in the code, but some can use some external references (Photosensor group should check?). 
In WCSimWCDigitizer, the DarkRate is set from the PMT type in the initializer *before* the call to WCSimDarkRateMessenger, to allow the user to overwrite the PMT type defaults. Although this should only be done for specific tests and studies and probably discouraged as we should trust the actual PMT DarkRate values.
I ran the code through gdb to see whether the values are properly set. I can do additional tests if requested but that might take some time due to travel.


tfeusels:
@tdealtry : Regarding pull request #89. It depends in which order everything is pulled into the develop branch. I can easily adapt or you could probably easily update my pull request whichever goes first I guess. Let me know what the best approach is.

eosullivan:
This is great. I especially like that you went through the literature and past HK talks to motivate the dark rate numbers. 

There are two things going on in this pull request: the method for how we set the dark rate is being changed and the dark rate numbers themselves are being changed. These are both important changes and it would be cleaner if they were done separately in two different pull requests.

As you said in your PR description, the dark rate numbers will need to be verified by the photosensor group. While we are approving the first pull request with the new dark rate method, you should circulate the proposed new numbers and where you found them around the software and photosensor group for comment via email.

tfeusels:
It wasn't very clear from the (no)vis.mac where those numbers go in. I used the numbers where I thought they applied and update the numbers where I didn't find any motivation or where I know where to find the motivation. 
I agree with your comments, but hope both changes can stay in this pull request while I inquire the photosensors'group feedback on the actual numbers. So, once they are approved/adjusted, I'll update the pull request.

eosullivan:
The default dark rate in the macros are 8.4 kHz for 20 inch HPDs/B&Ls, 3.0 kHz for 12 inch HPDs/B&Ls, and 4.2 kHz for other PMTs (which essentially just applied to the 20 inch PMTs). If you end up using these values and just filling in new values for the other photosensors, then it makes sense to be in one pull request. If you have to update any of the B&L, HPD, or 20" PMT values, then it should be a separate pull request.

tdealtry:
@tfeusels Looks like it'll be an easy modification to the new dark noise class, and it seems like this request will be merged into develop first. So I'll add it in when the time comes (and hopefully remember to ask you to check it)

tdealtry:
I've just had a thought - what about the dark rate conversion factor? This is also a property of the PMT, so shouldn't defaults be set in WCSimPMTObject too?
Maybe the idea is to get rid of this factor in the big WCSim tune, but until then I think automatically setting an appropriate default reduces the probability of errors when running.

eosullivan:
Yes, the conversion factor also be set in WCSimPMTObject in the same way as the dark rate. 

Based on the email conversation with the photosensor group, it seems the best option is to set the dark rate to 8.4 kHz for 20 inch HPDs/B&Ls, 3.0 kHz for 12 inch HPDs/B&Ls, and 4.2 kHz for other PMTs. Any update to the dark rate numbers will be a future project.

tfeusels:
Should I include the conversion factor in this pull request? The Dark Rate values are updated.
Comments should stay in as a reminder.

tfeusels:
Actually, I prefer *not* to add this dark rate conversion factor as I hope this should go away at some point (soon). Can someone remind of the reason for this additional factor and whether the actual solution is already in progress?

Let me know if there are additional requests for this pull request so I can scratch this off my list.

eosullivan:
This conversion factor takes into account that we apply the dark rate after digitization (while in real life, it happens prior to digitization). @tdealtry should correct me if I'm wrong, but I believe this conversion factor is still needed even after the most recent pull request (#89), but will no longer be needed once the full implementation of the DAQ/trigger scheme is finished since the dark rate will then be applied prior to digitization.

tdealtry:
Hi @eosullivan. In both the old code & the new code, dark noise is added prior to digitization, and both use the conversion factor (and both give the same rate; plots coming soon). So the original reason for the conversion factor has changed & perhaps it isn't needed at all anymore. Input from the photodetector group as to what the e.g. 4.2 kHZ number actually means (whether this is a measurement pre-digitization (with some hard charge threshold) or post-digitization (with the probability-based threshold)) will be required.

tfeusels:
Hi @tdealtry, from the recent e-mail chain discussion with the photosensor wg:
- "The default dark rate values of R3600 (SK PMT) described in skdetsim input card files is
4.2kHz for SK1 and 5.7kHz for SK4, respectively. Actually, these are for the values before
simulating the electronics threshold effect, since a skdetsim input card file requires such
values before cut by a threshold. The original (=measured) dark rate values with a 0.25pe
threshold were about 3.4kHz for SK1 and 4.5kHz for SK4. 
I guess that the current value (4.2kHz) in the default WCSim input card was taken from
the skdetsim input card as it was. Since a dark rate value in a WCSim input card is one
after the electronics threshold cut, it may be reasonable to use 3.4kHz, instead of 4.2kHz."

Could you briefly say what the remaining reason for the conversion factor is? Ideally the user should only care about setting the Dark Rate in its macro and/or implemented photosensor. I still find it strange to make that an actual PMT property. Of course, the macro option stays so if this pull request is closed now, the only thing that changed now is that all photosensor have intrinsic dark rates, which can always be altered through the macro.

Is the correction factor a strong requirement for this pull request?

tdealtry:
@eosullivan should comment, because now I've got no idea what the purpose of this number is. Both the old and new dark noise / digitizer / trigger classes add the dark noise before the digitisation. So, reading that email, I'd say that the conversion factor should be removed from both the old code and the new code. The fact that the old code has a conversion factor however means that either there was some miscommunication between the photosensor group & WCSim folks, or there is another reason for the conversion factor that I'm not aware of.

cwwalter:
You can ask Miura-san (I think @MakotoMiura is not in the WCSim organization?)  to comment on this.  He understands the technical details which relate to calibration.

eosullivan:
Ok, so some history: the conversion rate was put into WCSim by Okajima-san in March 2014. The commit message says: "Add Conversion factor to change the dark noise rate before digitization to after digitization". But looking through the code, @tdealtry is right. The dark noise is currently applied after the digitized gates are made, but before the digitizer effects are applied. So perhaps @YujiOkajima (or, as Chris suggested, @MakotoMiura) could comment on the reasoning behind why this factor was implemented.

YujiOkajima:
@tdealtry 
I thought the 4.2 kHz dark rate is the value after threshold cut.
Because the dark rate of new detectors are not decided and possible to be changed by future measurement, I think the dark rate should be change but should not be removed.

eosullivan:
Ok, the conclusion from the software call (not sure if you were there @tfeusels?) was that the photodetector group will be providing the dark rate value for various PDs after the threshold cuts. So, we would like to keep the conversion factor so that we can put in the rate they give us and transparently convert it to the rate we need to use in WCSim.

tfeusels:
Sorry, but I couldn't join the call. 
A few (practical) questions:
- The PMT Dark Rate is the rate of the pulses above some chosen, analog threshold (typically 0.25 pe), right? This is what we measure for each PMT in the photosensor wg. WCSim simulates random analog pulses according to the dark rate without threshold (so dark rate * correction factor)? Is that correct?
- Is there no way (separate issue maybe) to avoid the correction factor (which I consider a fudge factor) and directly simulate random pulses above a given threshold (given by the measurement conditions of the dark rate)? @tdealtry?
- If the answer is no, then it means there is no way around the fudge factor and I will add it as an empirical PMT property. If the answer is yes, then I think we should continue using the macro defined conversion factor until the improved implementation is included but not add it to the PMT properties.

Based on the whole discussion, it would be easiest to let me know how crucial the correction factor as PMT property is for this pull request. Nothing will change (as it's a macro parameter now) if we continue with the current pull request code. 

Thanks for the clarification on what I can do to finalize this pull request.

eosullivan:
Yes, I think your first bullet point agrees with my understanding of why the conversion factor is needed. 
 
Maybe let's back up a bit and I will explain my motivation for creating this issue to begin with. As you know, currently in WCSim if you want to simulate a detector with B&L PDs, for example, you need to change the detector configuration, uncomment out the appropriate dark rate, and uncomment out the appropriate conversion factor in the macro file. I found that users were sometimes missing a step in this process, often not changing the dark noise or conversion factor (or sometimes both). Since the dark rate is directly tied to PD type, at least in the way we have it implemented currently, there is no reason to have to do all three of these steps and I want a way for this to be done automatically so that the user only needs to specify the detector configuration and the dark rate will be implemented correctly. 

So, I think for now, we should move the dark rate and the conversion factor into the PMTObject as part of this pull request. I really like the discussion this PR has generated and I'm open to any new ideas about how to improve the treatment of dark noise in WCSim, but changing how we do things should be future work to be done in another pull request.

tdealtry:
Just to throw the awkward question out there - what happens when there is a new digitizer that has a threshold that is different to 0.25 p.e.? Presumably we'll need a different conversion factor.
Regarding simulating only noise above some (e.g. 0.25 p.e.) threshold, I'm not sure it'll be possible. A stumbling block I can see is that dark noise can occur on PMTs that have photon hit(s) in the same digitizer integration window. If we only allow dark noise charge of >0.25 p.e. to be added, this will change the probability of accepting photon+dark noise digits (i.e. can get single photon + charge of 0.1 p.e. from dark noise digits, which is accepted more readily than just a single photon digit). Plus, it's my understanding is that the digitizer threshold in WCSim isn't a hard cutoff, but probability based.

eosullivan:
If we end up with a digitizer that has a different threshold than the one used to measure the dark noise for a given PD, we could eventually make this conversion factor a property of the digitizer, or at least make it digitizer dependent somehow. Something to keep in mind for the future.

eosullivan:
@tfeusels, where are we with this PR? I would like to start clearing out these pending pull requests. I think as soon as you put the conversion factor into the PMTObject file, we should be ready to go.

YujiOkajima:
@tfeusels 
I checked the number of hit in SK geometry with SKDetSim when the dark rate was set to 4.2kHz in card and 0MeV gamma was set to incident particle (Threshold cut was not applied).
As a result, the Nhit (after digitization) in 1350 ns timing window was 62.49 hit.
The below figure shows the Nhit distribution of 10000events
darknoise_4 2khz_hit

The below figure shows the hit timing distribution of all hits in 10000events.

darknoise_4 2khz_time

There are 11146 tubes in SK-1 detector.
So, number of hit when the dark rate of PMTs is f kHz,
the expected dark rate is 11146 x f kHz x 1350 ns = 15.05f.
From MC result, the dark rate was calcurated to 4.15kHz ~ 4.2kHz.
Therefore the setting value of dark rate in SKDetSim card was the dark rate after digitization.

eosullivan:
@tfeusels, what is the status of this pull request? I think all that was left was to move the SetConvert rate into WCSimWCPMT. It would be nice to get this pull request cleared out. You will also need to update your branch after this latest pull request #89 merge.

tfeusels:
@eosullivan: I finally had some time to finalize this pull request. The conversion factors are added.
Two notes:
- Please check the comments I wrote as I think the "after digitization" is invalid, based on our discussion here.
- The SetDefaultValue in DarkRateMessenger.cc is removed for DarkRate and ConversionFactor as this would otherwise overwrite the proper defaults from WCPMTObject. Now the defaults from the PMT are correct.

src/WCSimDarkRateMessenger.cc
```diff
@@ -39,19 +39,18 @@ void WCSimDarkRateMessenger::Initialize()
...
   SetFrequency->SetUnitCategory("Frequency");
   SetFrequency->SetDefaultUnit("kHz");
   SetFrequency->SetUnitCandidates("Hz kHz MHz GHz");
-  SetNewValue(SetFrequency, G4UIcommand::ConvertToString(defaultFrequency, "kHz"));
+ 
 
   double defaultConvRate = 1;
   SetConversionRate = new G4UIcmdWithADouble("/DarkRate/SetConvert",this);
-  SetConversionRate->SetGuidance("Caribrate the frequency of dark noise after digitization");
+  SetConversionRate->SetGuidance("Calibrate the frequency of dark noise after digitization"); // TF: shouldn't this be "before threshold of 0.25 pe" instead of after digitization (cfr. Pull Request issue84 discussion)??
```
eosullivan: Sure, this SetGuidance text could be clearer. Feel free to change it.

src/WCSimPMTObject.cc
```diff
@@ -220,7 +219,27 @@ G4float PMT20inch::GetmaxQE(){
...
+G4float PMT20inch::GetDarkRate(){
+  /* From e-mail discussion with A.Konaka and S.Nakayama:
+   * SK-I: 4.2 kHz 
+   * SK-IV:5.7 kHz, both before electronics threshold in skdetsim
+   * Measured DN with 0.25 pe threshold:
+   * SK-I: 3.4 kHz  (2003 SK-NIM: 3 kHz. A.Konaka: "2kHz with hot PMTs removed?") 
+   * SK-IV: 4.5 kHz (higher due to FRP)
+   * ToDo: investigate after updating electronics routing, whether to change value to 3.4 kHz
+   */
+
```
eosullivan: I think this isn't a problem anymore? SKDETSIM also uses 4.2 kHz, and from Okajima-san's tests I feel convinced that the number is after the 0.25 pe threshold. If you agree, this comment should be removed.

src/WCSimPMTObject.cc
```diff
@@ -220,7 +219,27 @@ G4float PMT20inch::GetmaxQE(){
...
+  const G4float rate = 4.2*CLHEP::kilohertz;   //SKI value set in SKDETSim. 
+  return rate;
+}
 
+// Convert dark noise frequency to one before applying threshold of 0.25 pe, as that is what
+// will be simulated (WCSimWCDigitizer.cc ~L.160
```
eosullivan: Remove the line number in the comment as this will be constantly changing.

src/WCSimPMTObject.cc
```diff
@@ -373,7 +392,15 @@ G4float  PMT8inch::GetmaxQE(){
   return maxQE;
 }
 
+G4float PMT8inch::GetDarkRate(){
+  const G4float rate = 4.2*CLHEP::kilohertz;   //??
```
eosullivan: Remove the //??

eosullivan:
Ok, looks good. Just some comments on your comments, which I made inline. You should run some basic verification to make sure that the output is not changed by this PR.

src/WCSimPMTObject.cc
tfeusels: If SKDETSIM has 4.2kHZ after the 0.25 pe threshold as Okajima-san showed, then this is inconsistent with the e-mail discussion that I summarized in the comments, as it should have been before (3-3.4 kHz in the 2003 SK-NIM). Current behaviour is consistent with SKDETSIM, but I would prefer to leave the comments in as I'm not sure whether this is the correct behaviour for both. Please correct if I didn't understand/summarize the e-mail discussion properly.
Changing this number is for a separate pull request, but this comments makes sure we don't forget about it.

tfeusels:
I'll run the electrontest.mac with a few different settings to show the behaviour in the next few days.
The comments are adjusted except for one (see my reply).

src/WCSimPMTObject.cc
eosullivan: Ok, so if this is still an open issue perhaps the best thing would be to use the GitHub issue tracker rather than put this in the comments. The issue tracker allows new people to see the discussion and progress on the issue. We also, in general, don't want to be describing our open issues in the comments of the code.

tfeusels: That makes sense. I'll move the ToDo in that comment to a new issue.
The rest of the comment serves as reference/documentation of where the numbers
come from, which is the discussion we had over e-mail. Unless there is a better place
to document measurements/numbers used in the code.

eosullivan:
Ok, I'll hold off on merging this until the electrontest.mac plots are produced.

eosullivan:
The validation tests on this pull request still haven't been completed and we will be making the new release this week. Will you have time to do this? Otherwise, I will remove this issue from the milestone.

tfeusels:
I finally had some time to finish these simple verification tests. In attachment you'll find:
- current devel with electronTest.mac: comparing default and one where I added DarkNoise in the Macro (as in novis.mac), ie. followlng lines:
# default dark noise frequency is 0 kHz
/DarkRate/SetDarkRate 4.2 kHz #This is the value for SKI set in SKDETSIM.

# command to multiply the dark rate.
# Convert dark noise frequency before digitization to after digitization by setting suitable factor
/DarkRate/SetConvert 1.367  #For Normal PMT

Then in the second pdf I compared the electronTest_wDarkNoise (from the macro), with the default electronTest.mac from the issue84 branch. As this pull request assures the nominal DarkNoise and conversion factor are PMT properties and are set in PMTObject this should give exactly the same result, as illustrated by the plots.

Two notes: - running with dark noise seriously slows down the processing. Any thoughts on how to speed that up? @eosullivan @tdealtry ? Or unavoidable?
                  - The default behaviour if one does NOT specify the dark noise, as in electronTest.mac is now: PMTs WITH DarkNoise (ie. the natural behaviour). I added this explicitly in the electronTest.mac. I also updated the behaviour explanation in vis.mac and novis.mac. 

One should now clean up setting the dark noise in novis.mac and vis.mac. This would make the macros more readable. The macros can overwrite the nominal values if the user wants.

devel_wdn_vs_issue84_default_electrontest
devel_w_vs_wodarknoise_electrontest

eosullivan:
Ok, this looks good. Merging in.

PR #108, opened 2015-08-12
Modify comment about CE modification

YujiOkajima:
Because the hit efficiency evaluation of R3600 was wrong and evaluate again.
(R3600 hit efficiency was changed from 71.7% to 73.2%)
As a result, the expected output ratio of Box & Line PMT/R3600 (comparison of QE x CE x (hit efficiency) value) was fixed from 2.32 to 2.27.
This factor was mentioned in WCSimPMTObject.cc, and it was fixed on this pull request.

### DIFF
--- a/src/WCSimPMTObject.cc
+++ b/src/WCSimPMTObject.cc
@@ -1343,7 +1343,7 @@ G4float* BoxandLine20inchHQE::GetQEWavelength(){
 }
 
 G4float* BoxandLine20inchHQE::GetQE(){
-  G4float correctionFactor = 1./0.73;//Correction factor added in July 2015 to scale the output of B&L PDs to 2.32 times the 20" PMTS based on Hamamatsu simulation. This was done in Pull Request #98 and will be removed once a more permanent solution is found.
+  G4float correctionFactor = 1./0.73;//Correction factor added in July 2015 to scale the output of B&L PDs to 2.27 times the 20" PMTS based on Hamamatsu simulation. This was done in Pull Request #98 and will be removed once a more permanent solution is found.
   static G4float QE[20] =
     { 0.00*correctionFactor, .0008*correctionFactor, .1255*correctionFactor, .254962*correctionFactor, .2930*correctionFactor, .3127*correctionFactor, .3130*correctionFactor, .2994*correctionFactor, .2791*correctionFactor, .2491*correctionFactor,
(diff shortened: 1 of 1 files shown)


eosullivan:
Can you tell us more about the details of the hit efficiency calculation? How do you get this 73.2% number?

YujiOkajima:
@eosullivan 
I'm sorry for late response.
The hit efficiency was calculated based on the 1 p.e. distribution table of R3600 in WCSim and the threshold cut probability which is introduced in include/WCSimWCDigitizer.hh (function "Threshold").
In WCSim MC process, the charge of digitized p.e. is set based on 2 random number.
Based on 1st random number, the digitized charge against 1 p.e. is set to discrete value.
In order to set the charge with continuous value, 0 - 1/22.83 p.e. is added to the charge based on the 2nd random number.
Because I forgot to consider the 2nd random number, the charge which is set as argument of threshold function was smaller than expected value by 0.5/22.83 p.e.

On the other hand, because the Box & Line PMT hit efficiency was evaluated based on the fitted curve of 1 p.e. distribution, this error was not effected to Box & Line PMT hit efficiency.

eosullivan:
Do you use a calculation to determine this number? If so, could you explicitly show the calculation you did to get the 73.2% and 85.6% for the R3600 and B&L?

cwwalter:
Also for the title. You aren't just changing the "comment" this is changing the code itself.

eosullivan:
No, I don't think the code is changing, just the calculation that determines what value we will compare our output to. We are always using the 1/0.73 factor in the code and then just using this calculation to see how the simulation compares with our expectation of the ratio between the B&L and R3600.

cwwalter:
Ah Sorry.. I looked at this quickly in the airport and just noticed that it was a line with code that had changed, not that the part that had changed was an inline comment.

YujiOkajima:
@eosullivan , @cwwalter 
Sorry for delayed responce.
I have calculated by using the 1p.e. distribution table and threshold function in WCSim.
like the function below.
latex-image-1
The threshold function is defined in WCSimWCDigitizer.hh
I append the table used for hit efficiency calculation.
"New PD" sheet shows the calculation of Box & Line PMT and HPD hit efficiency.
In this sheet, the probability is set by using fitted function of 1 p.e. distribution used for 
making 1p.e. distribution table in WCSim.

"R3600_beforefix" is the calculation of hit efficiency before fixing the charge of each bin.
"R3600_fixed" is the fixed version.

http://www-sk.icrr.u-tokyo.ac.jp/~okajima/data/BandL&HPD&R3600_hiteff_analysis.xlsx

eosullivan:
Ok, this looks reasonable to me. I will merge this in tomorrow morning if there are no further comments before then.

PR #115, opened 2015-10-06
Remove the old digitizer (WCSimWCDigitizer)

### LINKED ISSUE #113: Remove the /DAQ/Construct line in vis.mac and novis.mac
This line was a temporary solution introduced to accommodate having both the old and the new digitizer schemes available. Will be removed before the next tag when the old digitization scheme is removed.

tdealtry:
* The files for the old dark noise + digitizer + trigger class WCSimWCDigitizer have been removed
* SKI_SKDETSIM is no longer a valid option for /DAQ/Digitizer or /DAQ/Trigger
* The enumeration kTriggerNHitsSKDETSIM has been removed
* The WCSimWCAddDarkNoise constructor using WCSimWCDigitizer has been removed
* The WCSimWCDigitizerNew files (containing WCSimWCDigitizerBase and WCSimWCDigitizerSK) has been renamed WCSimWCDigitizer
* The /DAQ/Construct option has been removed; instances of the digitizer and trigger classes are now created at the first call of WCSimEventAction::BeginOfEventAction(). This fixes #113 
* Various methods have been made private / non-virtual


tdealtry:
Just a follow up - I've run some electron particle guns & there is no change (everything is identical) in the SKI + NDigits outputs between the #89 merge and this.

eosullivan:
Ok, so now that issue #120 is closed, we should work to remove the old digitizer scheme. There are many pending pull requests related to changes to the new DAQ stuff. Do you have some preference as to the order in which these are merged? If not, I will probably merge in this one first (after I have reviewed it) since I would like to remove some of the things we had to do to accommodate the parallel digitzer scheme. Then we can focus on the PRs that deal with the improvements to the digitizer, ensuring that we continue to match with the skdetsim values at each step. It looks like #127 can be merged in whenever. There will probably be some conflicts that you will need to resolve as we start to merge in pull requests, regardless of the order we do this in.

tdealtry:
This one first would probably be easier to fix merge conflicts (the others are mainly minor code changes, some of which will probably lead to long discussions). The caveat is that #122 should be done first, or not done at all (I guess it depends if you want an intermediate tag with the options for the old class clarified). And yes, #127 can go in whenever.

tdealtry:
I've just done a merge locally with the current develop, and everything has messed up. This seems to mainly be due to the delete of WCSimWCDigitizer and rename of WCSimWCDigitizerNew to the old name. So don't review the code quite yet - I'll comment again when I fix the merge.
This'll mean that all my other PRs should be left until after this (except maybe #127 which is a more minor change). The other PRs from @tfeusels and @stewartt1982 should be simpler merges (if I keep the filename WCSimWCDigitizerNew for a while longer)

tdealtry:
I was just reading the diff wrong. No problems & ready to be looked at.

eosullivan:
Ok, I've reviewed this pull request and it looks good to go in as is. This removes the old digitizer scheme and also removes the /DAQ/Construct line. One small comment is that the program is now running a bit verbosely, but this can be something we change down the road once we have multiple digitizers and it becomes clear what interesting information users need printed to the screen.

Issue #112, opened 2015-09-30
Make the default digitizer and trigger options more realistic


tdealtry:
Currently, the default values of some of the digitizer / trigger options are not ideal
* The default value of the digitizer integration window for the SKI digitizer is 200 ns. The actual integration window used in SKI electronics is 400 ns.
* The default value of the trigger decision window is 200 ns. In SK, this value is chosen as it is the time for a Cherenkov photon to transverse the longest diagonal distance in the tank (the actual value is 220 ns; some rounding occurs). In order for people to use sensible values of the trigger decision window in other tanks, the default should depend on tank size.

tdealtry:
I just wanted to leave a comment on the automatic dark noise rate adjustment - currently this is calculated as:
`n_pmts * trigger_decision_window_seconds * dark_rate_Hz`
Using a particle gun with 50000 0 MeV electrons (so no Cherenkov photons are created) in the SK geometry (11146 PMTs) with a 200 ns trigger decision window, we see:

Dark rate | 4.2 kHz | 8.4 kHz
-------------|------------|------------
Expected occupancy | 9.4 | 18.7
Expected occupancy * 1.367 | 12.8 | 25.6
Actual occupancy | 9.4±1.4 | 18.8±1.9

Therefore the 1.367 conversion factor is not required in the average occupancy calculation and the code is correct as it is currently.

tdealtry:
I've had a look through SKDETSIM (an old version: Generated on Thu Jan 21 20:35:20 JST 2010 from branch tags with tag skdetsim-v11p79 by okumura.) for the defaults, mainly in the files:
* dsparam.h (variable definitions)
* dsinip.F (variable defaults)

From what I can make out, SKDETSIM uses the following numbers by default for simulating SKI:
* Digitizer integration window (SIGWDS, SGGATE) 300 ns*
* Digitiser deadtime (DEADDS) 900 ns*
* Digit offset (OFFSET0) [540,960] ns depends on run mode. The 950 ns we currently use is fine?
* Trigger decision window (TWINDS) 200 ns
* Trigger threshold (TRIGDS) 30 digitised hits. le_sk1_example.card uses 15. The 25 we currently use is fine?
* Posttrigger - Pretrigger window (GATEDS) 1300 ns. The 950 - (-400) = 1350 ns we currently use is fine?
* Pretrigger or postrigger window. I can't work this out.

[...]

PR #116, opened 2015-10-06
Make the default SKI digitizer integration window more realistic

tdealtry:
Change the default SKI digitizer integration window from 200 ns to the value of the actual integration window used in SKI electronics: 400 ns.
This is a partial fix of #112


eosullivan:
Have you done any tests to show how this change affects on the physics output of the code? After making this change, are we still able to match SKDETSIM? Also, was there any resolution to your finding that the digitizer integration window is 300 ns in SKDETSIM?

tdealtry:
Comparing 200 ns and 400 ns [dealtry_2016_01_21_wcsim.pdf](https://github.com/WCSim/WCSim/files/99514/dealtry_2016_01_21_wcsim.pdf)
* The nhit & ndigit distributions look fine
* The digit time distributions look ok, but the breakdown (noise-only vs Cherenkov-only vs mix) is different - there is migration to the 'mix' category (as expected)
* There's a small difference in the trigger times at low energy, on the same order as we saw before (a few 10s ns)

I'll chase people up about SKDETSIM in Kashiwa. My current best guess that SKDETSIM uses 300 ns is exactly that - a guess. To reiterate, I currently have no access to the complete SKDETSIM source code, I can't run it, and the files I do have are from 2010.

cwwalter:
> To reiterate, I currently have no access to the complete SKDETSIM source code, I can't run it, and the files I do have are from 2010.

Just as a reminder you can get a suketto account at Kamioka as a HK member.

tdealtry:
> Just as a reminder you can get a suketto account at Kamioka as a HK member.

Yep, I applied last week. Just waiting for approval.

eosullivan:
I'm going to hold off on merging this pull request until after the HK meeting. This will give you time to look through SKDETSIM and to ask some SKDETSIM experts at the meeting, and is not so crucial to be in this next release.

tdealtry:
I'm going to close this PR. 200 ns is the value that's been used for the past 5 years for HK studies, so should continue.
We will of course have a more realistic HK digitiser simulation at some point in the future. And requests from SK collaborators  to make the digitiser more realistic are also welcomed

PR #117, opened 2015-10-06
Make the default NDigits trigger decision window more realistic

tdealtry:
The default value of the trigger decision window is 200 ns. In SK, this value is chosen as it is the time for a Cherenkov photon to transverse the longest diagonal distance in the tank (the actual value is 220 ns; some rounding occurs). In order for people to use sensible values of the trigger decision window in other tanks, the default should depend on tank size.
This commit sets the default trigger decision window using the maximum photon travel time (assuming a cylindrical detector) rounded to the nearest 100 ns. Therefore, for the SuperK geometry 200 ns is used, and for the Cylinder_60x74_20inchBandL_40perCent geometry 400 ns is used.
This is a partial fix of #112


eosullivan:
The idea of having a variable trigger window makes me a bit nervous. In real life you set a constant trigger window, which may or may not be the maximum photon travel time in the detector. What do you think of this: we could leave the default trigger window set to the SKI value and move this option of setting the trigger window to the maximum photon travel time to be an option in the macro?

Issue #114, opened 2015-10-05
Make the NDigits2 trigger option stable


eosullivan:
This is a nice example of how to add new derived trigger classes into WCSimWCTrigger.cc, and is a good example of using multiple triggers. However, this experimental feature occasionally throws the error: 

``WCSim: src/WCSimWCTrigger.cc:371: void WCSimWCTriggerBase::FillDigitsCollection(WCSimWCDigitsCollection*, bool, TriggerType_t): Assertion `triggered_composition.size()' failed.
``
This trigger class must be stable before the next release of WCSim. I will nominally assign @tdealtry, but if there is someone else who is more appropriate, please let me know.

tdealtry:
#118 fixes the most obvious causes of this. I'll set some high stats jobs running at various energies to check if there are any remaining issues.

tdealtry:
I've run some high/medium/low energy electron particle guns with 5000 events in each. No crashes were seen, and distributions look fine. So I'm fairly confident that #118 fixes everything.

tdealtry:
I think we can close this issue now. The error that was first reported has now been fixed; another issue can be made if there are any future problems.

eosullivan:
Ah, GitHub was doing so well with automatically closing issues. I'll close this now.

PR #118, opened 2015-10-06
WCSimWCDigi::RemoveDigitizedGate()

tdealtry:
This fixes some bugs in WCSimWCDigi::RemoveDigitizedGate() concerning the digitized composition vector of pairs, which is the cause of #114


eosullivan:
This seems reasonable and fixes the NDigits2 trigger option without affecting the output of the default NDigits trigger. I'll go ahead and merge this in.

PR #119, opened 2015-10-07
Allow digits with negative times to be kept by the trigger

### LINKED ISSUE #111: Sensible handling of digits with negative times
There is a continue statement in WCSimWCTriggerBase::FillDigitsCollection() which rejects digits with negative times. For reference, digit time is defined as:
* - trigger time + time of first hit within digit + offset + smearing

This continue statement will cause problems in the future; it was put here as this is what is done in the old digitizer. However if you think of the following situation (ignoring the smearing)
* pretrigger window -5 microseconds (which is the minimum for all SKIV trigger types, except SLE and relic)
* offset -1 microseconds (which is ~ the case for SKIV)
* this means the earliest digits (i.e. -5 microseconds before the trigger) will have a digit time of ~ -4 microseconds and will never get saved (whereas in real life, they are saved)

Using the defaults there won't be any problems with unsaved digits (offset -950 ns, pretrigger window -400 ns), but the more general case should be handled correctly.

tdealtry:
Just removed the `continue` statement that caused digits with negative times to not be saved. This fixes #111 
To check that it doesn't cause any problems, I ran some electron particle guns with the 950 ns offset variable set to -950, 0, +950 ns. These samples have ~all/some/no digits with negative times. NDigit / p.e. / raw hit time style plots are identical between the 3 configurations, and digit time plots are also unaffected (once the n * 950 ns differences are taken into account).


eosullivan:
Ok, this seems reasonable. I'm having a bit of a hard time picturing the verification you did, though. Could you attach these plots here?

tdealtry:
Currently, the post-trigger digit time is defined as:
```
G4double digihittime = -triggertime
            + WCSimWCTriggerBase::offset
            + digit_time
            + PMT->HitTimeSmearing(Q);
```
I changed the value of WCSimWCTriggerBase::offset from +950 ns to 0 ns and -950 ns, as a simple way to force negative digit times, as so:
compare

To confirm that things are identical, I looped over all bins, applied an (n * 950) ns offset, and made sure the bin contents agree. They do.

The following are a selection of plots comparing the ±950 ns options, showing that everything else is the same.
Raw hit time (part of the input)
[compare_h1hittime_+.root+_1.pdf](https://github.com/WCSim/WCSim/files/97291/compare_h1hittime_.root._1.pdf)
NHits (sorry about the poor quality, check the KS prob!), trigger time (part of the output)
[compare_h1nhits_+.root+_1.pdf](https://github.com/WCSim/WCSim/files/97292/compare_h1nhits_.root._1.pdf)
[compare_h1triggertime_+.root+_1.pdf](https://github.com/WCSim/WCSim/files/97296/compare_h1triggertime_.root._1.pdf)

eosullivan:
Ok, so we need to put this in for SK-IV and looking at this validation it seems like negative hit times won't do bad things to our simulation. I'll merge this in.

PR #121, opened 2015-10-12
Feature/hit time smearing

### LINKED ISSUE #110: Move hit time smearing to the correct place
The function [HitTimeSmearing()](https://github.com/WCSim/WCSim/blob/develop/src/WCSimPMTObject.cc#L81) is currently called in WCSimWCTriggerBase::FillDigitsCollection() (i.e. after the trigger decision is made). This causes digits to migrate in/out of the 200 ns trigger decision window, meaning there are false triggers and triggers with the incorrect trigger time. In the most extreme case, this will cause events with less digits than the trigger threshold (see #48).
Looking into the code, it seems* that this hit time smearing describes the PMT time response (i.e. isn't a digitizer or trigger effect, but actually effects the input to the digitizer), and so should be added to WCSimWCPMT (and possibly WCSimWCAddDarkNoise too, but my first thought is that this would be double-counting the observed random fluctuations).
*WCSimPMTObject isn't documented, so I had to get what the function means from the [wiki](https://github.com/WCSim/WCSim/wiki/Tutorial:-How-to-add-your-own-photdetector). PMT experts should comment on what exactly this function means.

tdealtry:
Fixes #110
Note that this is a change from applying [HitTimeSmearing()](https://github.com/WCSim/WCSim/blob/develop/src/WCSimPMTObject.cc#L81) to a digit, to applying the function to raw photon hits.
This pull request does not apply time smearing to raw dark noise hits - the idea of the dark noise routine is to evenly distribute single p.e. hits over all PMTs over a specified time window, and so by applying the smearing to dark noise hits it would either be:
* incorrectly dealing with edge-of-time-window effects (e.g. the pre/post subevent windows)
* computationally wasteful - needing to use a wider window to add dark noise to, just to deal with edge effects

Looking at the effect of the change on electron particle guns (both with old/new digitizer+trigger code):

* raw hit time distributions are smeared. I'll leave it up to WCSim gatekeepers whether this is desired behaviour (currently the time the Cherenkov photon hits the PMT is saved; this change effectively saves the time the PMT makes a signal). If you'd like to keep the old behaviour, it will be possible by e.g. adding extra fields to WCSimWCDigi (e.g. a new raw hit time variable)

hittime-1

* Trigger time is different (this has a knock-on effect on digit time)
  * ~5-10 ns earlier at 100-500 MeV (this is to be expected - instead of the sudden turn-on of digits (see the dashed line above) there are digits at earlier times (see solid line above), therefore the trigger time will be earlier)
  * ~5 ns later at 5 MeV (the peak of the hit time distribution is shifted slightly later (see solid line above) therefore, at low energy where the trigger is running close to its threshold, the trigger will be later
  * these are small changes compared with the window saved (~1 us)

* NDigits & total p.e. are essentially unchanged

* The old(new) trigger is slightly more(less) efficient at 5 MeV (~1% effect)


eosullivan:
I have no problem with the time of the raw hits being smeared, since I think this is more realistic to what is actually happening. The person who might take issue with a change in the raw hit collection is Yano-san. @takatomiyano, can you comment on this? Perhaps this can be motivation for you to incorporate your low-energy trigger scheme into WCSim so we can track things like this?

eosullivan:
After looking at this again (and after a discussion with Chris) I realize that @takatomiyano probably uses the raw data structure coming out of WCSimWCSD which is explicitly written to the ROOT file, rather than this intermediate structure that we introduced last year which comes out of the PMT. So, what you are doing seems right and is the way it should be. We'll talk to Yano-San at the meeting to make sure he knows about this change, but it is probably fine to merge this in. Can you resolve the conflicts?

tdealtry:
Conflicts resolved.

cvilelasbu:
Hi @tdealtry and @eosullivan,

Sorry to come back to such an old PR but I noticed this change introduces significant a effect in the digitized hit times, particularly with the old "PMTXXinch" type of PMTs used in the SuperK (and NuPRISM) geometry.

In the following plots, I shot 500 MeV muons pointing straight at the top end-cap at around 5 m to the wall.

 - With the SuperK (WCSim default) geometry:
   - The digitized hit time distribution for version v1.6.0 and for v1.6.0 with the changes in this pull request reversed:
sk_thit
   - The 2D distribution of the hit time and charge for the nominal v1.6.0
sk_thitvsq_v160
   - And with PR 121 reversed:
sk_thitvsq_nopr121
   - Difference between the two:
sk_thitvsq_significance

 - For the Hyper-K 40% geometry, the difference is much smaller:
hk40percent_thit
hk40percent_thitvsq_v160
hk40percent_thitvsq_nopr121
hk40percent_thitvsq_significance

From the above, it looks like the changes in this PR are moving digitized hits to earlier times.

As far as I understand it, before this PR the hit time was given by the earliest raw pe and then smeared with `PMT->HitTimeSmearing(Q)`, where `Q` was the *total* integrated charge in the digit.

With this PR, the hit time smearing was moved upstream in the digitizer code and is now applied to individual pe's, using the charges of *individual* pe's as the `Q` in `PMT->HitTimeSmearing(Q)`.

As the sigma in the smearing function goes as `C + D/Q` (`C` and `D` are constants), after this PR, the multi-pe hits are shifted to earlier times, as each of these pe's will be smeared by a wide Gaussian function, and after that the earliest time will be picked. Before this PR, the earliest (unsmeared) pe would be used and its time smeared by a much narrower Gaussian.

So, the new way of smearing the hit times seems conceptually very different from what was done before and introduces significant differences in the observable quantities. Is this the desired behaviour? Should we be applying the same function to individual pe's that we were applying to the "overall" hit?

tdealtry:
Thanks for making these plots @cvilelasbu I knew there was a change to the digitised times, but you are correct in noticing that it is too large (the change would have worked if it didn't depend on Q...). I guess the quick fix is to revert this PR, so that the function we have and know works is applied at the correct stage.

In the longer term, asking the photosensors group for a PMT-level (i.e. single p.e.) time smearing (if such a thing can be de-convoluted from the electronics smearing).

tdealtry:
Actually... the short term solution shouldn't be to revert this exactly (i.e. in `WCSimWCTrigger`) but apply the smearing at the digitizer stage (i.e. in `WCSimWCDigitizer`)

eosullivan:
Ok, so let's move the smearing to WCSimWCDigitizer for now. @tdealtry, this should still fix #110 if done this way, right?

tdealtry:
Yeah it will do.
However re-reading #110, I understand why I moved from `WCSimWCTrigger`->`WCSimWCPMT`, instead of `WCSimWCTrigger`->`WCSimWCDigitizer`. The only source of documentation on the `HitTimeSmearing()` I could find is in the [wiki](https://github.com/WCSim/WCSim/wiki/Tutorial:-How-to-add-your-own-photdetector), and suggests it's a PMT rather than PMT+digitizer effect. So, if this documentation is correct, the function call should be moved to `WCSimWCPMT` whenever the next WCSim retune happens.

PR #122, opened 2015-10-27
Option clarification for WCSimWCDigitizer

tdealtry:
Some new DAQ options have no effect on the old digitizer WCSimWCDigitizer, and this wasn't documented. This commit does two things:
* Makes the trigger decision window option have an effect on WCSimWCDigitizer
* Makes it clear in the .mac files that the generic digitizer options (deadtime & integration window) have no effect on WCSimWCDigitizer

### DIFF
--- a/src/WCSimWCDigitizer.cc
+++ b/src/WCSimWCDigitizer.cc
@@ -45,8 +45,8 @@ WCSimWCDigitizer::WCSimWCDigitizer(G4String name,
   //in this class use this synonym for clarity
   nhitsThreshold  = ndigitsThreshold;
   if(nhitsWindowBins % 5) {
-    G4cerr << "Option /DAQ/TriggerNDigits/Window must be exactly divisble by 5 when running with SKI_SKDETSIM" << endl;
-      exit(-1);
+    G4cerr << "Option /DAQ/TriggerNDigits/Window must be exactly divisble by 5 when running with SKI_SKDETSIM" << G4endl;
+    exit(-1);
   }
   nhitsWindowBins = ndigitsWindow / 5;
 }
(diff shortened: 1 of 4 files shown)


eosullivan:
I agree with the extra commenting, especially since the old digitizer scheme is sticking around a bit longer than we had planned. I am a bit confused about the need for adding the nhitsWindowBins variable, though. You have only added it to the old digitizer scheme (WCSimWCDigitizer.cc) and it seems like it just gives the ability to adapt the gate widths in the digitizer. Will we ever actually want to adapt the gates in this old scheme?

tdealtry:
I was trying to compare the old and new code for HK, and got burnt by the fact that the trigger decision window is fixed to 40 * 5 ns bins in WCSimWCDigitizer (so was comparing old & new with different trigger decision windows, making the results meaningless). Whether anyone but me is interested is another question. But it can't hurt.

src/WCSimWCDigitizer.cc
```diff
@@ -43,7 +43,12 @@ WCSimWCDigitizer::WCSimWCDigitizer(G4String name,
   //the nhits threshold value is read into the base class from /DAQ/TriggerNDigits/Threshold
   // into the variable ndigitsThreshold
   //in this class use this synonym for clarity
-  nhitsThreshold = ndigitsThreshold;
+  nhitsThreshold  = ndigitsThreshold;
+  if(nhitsWindowBins % 5) {
```
eosullivan: This if statement isn't working. Should be a !=0 or something?

tdealtry:
Sorry about that - I was using the wrong variable. It should work now

eosullivan:
Ok, I'll close this now that #115 was merged. This was a pretty minor change and in the interest of time I decided to fast track #115 instead.

PR #123, opened 2015-10-27
Random particle gun vertex positioning in other detectors

tdealtry:
The method of randomly distributing particle gun particles across many locations/directions (MakeKin.py) assumes the SuperK geometry. This commit adds a new option to MakeKin.py where the detector can be specified, and the vertex positioning uses this to calculate the radius/height of the volume to distribute vertices in.
Currently only setup for SuperK, and the HK cylinders, but new detectors can be added in a single line.


eosullivan:
Great, it's nice to have a way to make simple vector files in our new configurations. A small suggestion: I am worried that to change a default you must change both the optdefault and the number that is explicitly set in the addoption command as default=#. Could you use the optdefault variable to actually plug in the default number?

tdealtry:
I meant to do that, but missed it. Thanks for spotting it!

eosullivan:
Ok, looks good. A nice extension to this tool.

PR #124, opened 2015-11-03
Restrict number of digits per PMT per trigger to <= 1

### LINKED ISSUE #120: Make the NDigits and total p.e. distributions at high energy in the new digitizer/trigger scheme agree with the old digitizer/trigger
The combination of the new SKI digitizer and NDigits trigger disagrees with the old WCSim SKI_SKDETSIM combined digitizer+trigger in the NDigits and total p.e. distributions for high energy electron particle guns. In the past, the WCSim SKI_SKDETSIM was compared* with the SKI mode of skdetsim, and good agreement was seen at high energy. Therefore it is important for WCSim's new classes to agree with the old class, in order to perform high energy physics studies.
*@cwwalter it'd be useful to know which configurations were tested (e.g. particle gun particle types, energies, etc.), which distributions were plotted, and how close the agreement was. Could you post a link to slides/documents?

nhits-1
NDigits is calculated as the number of digitized hits in the first subevent in the output file, and so is after the entire digitization + trigger process.
pe-1
Total p.e. is calculated by summing Q of the digitized hits in the first subevent in the output file.

The above plots are for 5 GeV electrons without dark noise. See #89 [verification plots](https://gist.github.com/tdealtry/0efb539cd2ab425900e8) for more configurations.

It is useful to explain how digits are created & saved
* New code
  1. Find the first raw hit on a PMT in the simulation. Integrate hits on this PMT for hit time + 200 ns.
  2. Apply the efficiency factor & threshold function. Create a new digit if threshold is passed
  3. Reject hits in the deadtime window (0 ns)
  4. Find the next hit. Integrate for hit time + 200 ns. Go to iv.
  5. Only digits that pass an NDigit trigger are saved
* Old code
  1. Trigger on raw hits
  1. Find the first raw hit on a PMT in the subevent window. Integrate hits on this PMT for hit time + 200 ns.
  2. Apply the efficiency factor & threshold function. Create a new digit if threshold is passed
  3. Digitizer is now dead until next subevent window
  3. All digits created are saved
  

The expected cause of this is the number of digits (digitized hits) allowed per PMT per subevent window - in the old code it is 0 or 1, whereas in the new code >1 is allowed.

There are a few potential solutions
* Restrict the new code so that it allows only only 0 or 1 digit per PMT per subevent window.
* Change the digitizer deadtime in the new code ([verification plots appendix A](https://gist.github.com/tdealtry/0efb539cd2ab425900e8) shows that moving from 0 ns to 500 ns, this works for electron guns)
* Change the digitizer integration window in the new code (#116) - a longer integration window will cause a reduction in NDigits, but not total p.e. (unless combined with digitizer deadtime)

I've tried having a look at the skdetsim code to see what values are used, and how many digits per PMT per subevent window are allowed, but have been unable to work this out (I can't run skdetsim, I don't have all the source files, and I don't know fortran). So it'd be useful if someone who knows skdetsim could take a look & report on how it does things, so that the best solution(s) can be found.

This should be solved before #115 is merged

tdealtry:
This intends to fix #120. It adds the option to restrict the number of digits per PMT per trigger to <= 1, which is how SKI works. It is the default in both the NDigits and NDigits2 classes, and has no effect on the SKI_SKDETSIM class. I've added it as an .mac option (rather than hardcoded) so that people can study the effect of this on any trigger.

The following plots are for a 5 GeV electron particle gun, and show that turning the restriction on (red) is more consistent with the SKI_SKDETSIM option (black).

hits5000_42-1

time25000_42-1

### DIFF
--- a/include/WCSimWCDAQMessenger.hh
+++ b/include/WCSimWCDAQMessenger.hh
@@ -41,6 +41,7 @@ private:
   G4String            StoreTriggerChoice;
   G4UIcmdWithABool*   MultiDigitsPerTrigger;
   G4bool              StoreMultiDigitsPerTrigger;
+  G4bool              MultiDigitsPerTriggerSet;
 
   G4UIcmdWithoutParameter* DAQConstruct; //TODO remove this
 
@@ -73,6 +74,7 @@ private:
   G4int                 StoreNDigitsPostWindow;
 
   G4String initialiseString;
+  G4bool   initialised;
 };
 
 #endif
--- a/include/WCSimWCDAQMessenger.hh
+++ b/include/WCSimWCDAQMessenger.hh
@@ -39,8 +39,8 @@ private:
   G4String            StoreDigitizerChoice;
   G4UIcmdWithAString* TriggerChoice;
   G4String            StoreTriggerChoice;
-  G4UIcmdWithABool*   RestrictDigitsPerTrigger;
-  G4bool              StoreRestrictDigitsPerTrigger;
+  G4UIcmdWithABool*   MultiDigitsPerTrigger;
+  G4bool              StoreMultiDigitsPerTrigger;
 
   G4UIcmdWithoutParameter* DAQConstruct; //TODO remove this
 
--- a/include/WCSimWCDAQMessenger.hh
+++ b/include/WCSimWCDAQMessenger.hh
@@ -39,6 +39,8 @@ private:
   G4String            StoreDigitizerChoice;
   G4UIcmdWithAString* TriggerChoice;
   G4String            StoreTriggerChoice;
+  G4UIcmdWithABool*   RestrictDigitsPerTrigger;
+  G4bool              StoreRestrictDigitsPerTrigger;
 
   G4UIcmdWithoutParameter* DAQConstruct; //TODO remove this
 
--- a/include/WCSimWCTrigger.hh
+++ b/include/WCSimWCTrigger.hh
@@ -60,8 +60,8 @@ public:
   // Trigger algorithm option set methods
   //
 
-  ///Set whether to restrict the number of digits per PMT per trigger to <= 1
-  void SetRestrictDigitsPerTrigger(G4bool restrict) { restrictDigitsPerTrigger = restrict; }
+  ///Set whether to allow the number of digits per PMT per trigger to go > 1
+  void SetMultiDigitsPerTrigger(G4bool allow_multi) { multiDigitsPerTrigger = allow_multi; }
 
   // NDigits options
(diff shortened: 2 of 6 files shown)


eosullivan:
Somehow the restrictdigitspertrigger naming scheme makes me think it is doing something special, rather than just doing what the digitizer actually does. Maybe something like multidigitspertrigger, which would be set to false by default, may be clearer. 

We now have many trigger and digitizer options. I wonder if it might be worth making a separate macro with all these options in there and having novis/vis.mac include just the /DAQ/Digitizer and 
/DAQ/Trigger options, but none of the special options where you can change the timing windows, etc.

eosullivan:
Also, the verification looks nice. The old digitizer now agrees with the new digitizer scheme for high energy. I assume that you see no effect at low energy after this change?

tdealtry:
Sorry I forgot to mention - the only effect I've seen on the distributions I've looked at are at high energy (>~1 GeV), on the distributions shown above (and related plots e.g. total p.e.). Some digits are being cut out, and they're the later ones in the event.

I'll go through & swap the RestrictDigitsPerTrigger to MultiDigitsPerTrigger.

In terms of what option you want in the separate DAQ.mac file, could you please clarify? I've helpfully copied all the current options below - if you could write the ones you want in the separate file in your next message, that'd be great. Then I can put in a new PR.
/DAQ/Digitizer
/DAQ/Trigger
/DAQ/DigitizerOpt/DeadTime
/DAQ/DigitizerOpt/IntegrationWindow
/DAQ/RestrictDigitsPerTrigger
/DAQ/TriggerSaveFailures/Mode
/DAQ/TriggerSaveFailures/TriggerTime
/DAQ/TriggerSaveFailures/PreTriggerWindow
/DAQ/TriggerSaveFailures/PostTriggerWindow
/DAQ/TriggerNDigits/Threshold
/DAQ/TriggerNDigits/Window
/DAQ/TriggerNDigits/AdjustForNoise
/DAQ/TriggerNDigits/PreTriggerWindow
/DAQ/TriggerNDigits/PostTriggerWindow
/DAQ/Construct

eosullivan:
For completeness, I will attach the slides @tdealtry  showed at the software meeting. This contains the low energy comparison, as well as some comparison with data. (p.s. You can now attach PDFs to GitHub comments!).
[dealtry_2015_11_04_hk_software.pdf](https://github.com/WCSim/WCSim/files/30514/dealtry_2015_11_04_hk_software.pdf)

I think this PR looks ready to be merged in. @tdealtry ,do you agree?

tdealtry:
Finally pdfs! Great!

Yeah it's ready to go

PR #125, opened 2015-11-09
Feature/c make lists rootcint

tfeusels:
The "make rootcint" step that creates the WCSimRootDict.cc file and then creates the shared object libWCSimRoot.so and properly links it, was not yet possible with the current CMakeLists.txt.
The CMakeLists.txt has been updated to also compile the rootcint libs.
Now one can do in their build directory:
cmake ../src
make clean
make (will first make the rootcint libs: WCSimRootDict.cc in the src dir, and libWCSimRoot.co in the build dir, then perform the standard compilation and linking).

Fully tested and already frequently used.

Also updated the required CMAKE to 2.8 (as 2.6 is really quite old). The ROOT required in the comments might have been only necessary for debugging (using PARSE_ARGUMENTS), but recommended to use the latest one though.


tdealtry:
@tfeusels Could you document how to build with cmake? The obvious place to put it would be README.md

tfeusels:
@tdealtry I just updated the documentation now. Try it out and let me know if it's clear.
@eosullivan Let me know if this is good to go with the updated README now.

tfeusels:
@eosullivan : is this good to go?

eosullivan:
Ok, this looks good. I couldn't fully test it since my Geant and ROOT aren't compiled with cmake and so I seem to be getting errors related to that, but I trust that you have tested it and it works. 

Maybe there are just a couple of modifications before this goes in:
* You should update the macros in CMakeLists.txt (set(WCSIM_SCRIPTS) currently lists novis.mac and vis.mac, but now these are WCSim.mac - sorry, I probably could have coordinated the ordering of the pull requests a bit better).
* The recommended version of Geant is still 4.9.4 (and will hopefully soon be 4.10). Update the README to refer to 4.9.4 instead of 4.9.6.

eosullivan:
And since #133 just got merged in, could you add the enumerations library to the cmake file while you're in there?

tfeusels:
Thanks for the comments. I updated the branch with your changes. 
However, compiling with CMake for Geant4.9.4 works, but is more experimental as it was just introduced then. I explained it better in the README. This is rather a suggested way of building
WCSim once we're beyond Geant4.9.4. 
I assume the new recommended Geant4 version will be 4.10.2 as well? I did compile all Geant4's for when WCSim will be ready and can update documentation accordingly if problems arise during cmake builds.

Issue #126, opened 2015-11-10
Make a separate macro with all the new DAQ options


eosullivan:
We have lots of new options relating to the DAQ that can be set via the macro. Since vis.mac and novis.mac are getting pretty long, I think we should consider making more specialized macros that show all of the available macro options for a particular topic. We could try this first with the DAQ options.

eosullivan:
Cutting and pasting from the comment section of PR #124, the number of options are: 
/DAQ/Digitizer
/DAQ/Trigger
/DAQ/DigitizerOpt/DeadTime
/DAQ/DigitizerOpt/IntegrationWindow
/DAQ/RestrictDigitsPerTrigger
/DAQ/TriggerSaveFailures/Mode
/DAQ/TriggerSaveFailures/TriggerTime
/DAQ/TriggerSaveFailures/PreTriggerWindow
/DAQ/TriggerSaveFailures/PostTriggerWindow
/DAQ/TriggerNDigits/Threshold
/DAQ/TriggerNDigits/Window
/DAQ/TriggerNDigits/AdjustForNoise
/DAQ/TriggerNDigits/PreTriggerWindow
/DAQ/TriggerNDigits/PostTriggerWindow
/DAQ/Construct

So, perhaps we can keep:
/DAQ/Digitizer
/DAQ/Trigger
/DAQ/Construct

in vis.mac and novis.mac to just show the bare functionality of the DAQ and then include the entire list above in the newly created DAQ macro.

tdealtry:
I think we can close this issue now

PR #127, opened 2015-11-10
daq.mac - the place for describing the DAQ options

tdealtry:
Move most of the DAQ options (thresholds, time windows, etc.) into a new file - daq.mac. This file is then called with novis.mac and vis.mac
This is to solve #126


tdealtry:
Just to note that I've not put everything in daq.mac (list of things not in are below), since @eosullivan requested to keep them in novis.mac and vis.mac. I didn't add them to daq.mac too, as I thought having them in two different locations gives the strong possibility of running issues.
/DAQ/Digitizer
/DAQ/Trigger
/DAQ/Construct

eosullivan:
Looks great. This is the first step to having a more basic novis/vis.mac and instead enabling advanced options from their own separate macro.

PR #128, opened 2015-11-10
Fix bug in WCSimWCAddDarkNoise::FindDarkNoiseRanges to prevent seg fa…

stewartt1982:
Fixes a bug in WCSimWCAddDarkNoise::FindDarkNoiseRanges (when using /DarkRate/SetDarkMode 1) which causes a segmentation fault to occur when no hits are recorded in the PMTs.  Code detects this occurrence and results in no dark noise hits being added (expected behaviour).

### DIFF
--- a/src/WCSimWCAddDarkNoise.cc
+++ b/src/WCSimWCAddDarkNoise.cc
@@ -213,6 +213,19 @@ void WCSimWCAddDarkNoise::FindDarkNoiseRanges(WCSimWCDigitsCollection* WCHCPMT, 
   //for the algorithm below to work
   sort(ranges.begin(),ranges.end());
 
+  //check if the vector range has any entries
+  //If no entries this indicates that no digits were
+  //found in the WCSimWCDigitsCollection, which can cause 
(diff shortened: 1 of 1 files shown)


eosullivan:
I'm confused what this means. Does this mean if no hit is recorded in *any* PMT in an event, then no dark noise will be added?

stewartt1982:
The dark noise routine at the moment can be run in two ways.  One mode adds noise to a user specified time window (eg. the use can specify 100ns - 1000ns).  In the other mode it adds noise to a user specified time window around hits coming from real photons.  If there are no hits, no noise is added.  Somewhat analogous to the situation before pull request 89, where if there was no trigger found no noise was added.

tdealtry:
Just to be clear "If there are no hits, no noise is added." is only for mode 1, where hits are added around real photon hits.
Mode 0 adds noise to a fixed time window, whatever the photon hit distribution (i.e. if there are any, or none)

eosullivan:
Ok, great. Thanks for the bug fix!

PR #129, opened 2015-11-10
Fix/sens detector

tfeusels:
When running the current develop or latest tagged release 1.4.0, segfaults will happen when
you specify the same detectorGeom as the default (and change some specifications in the macro)
and do Construct. This segfault happens when simulating more than one event and the same tube as event one is getting hit. Normally the PMTHitMap should have been cleared, but because the SDManager has double registered the same sensitive detector (glassFacePMT) with the same path, confusion arises, another PMTHitMap is not getting cleared, the code tries to find the hitCollection for that tube, which is not there for event2, and segfaults. 
=> Registering the same sensitive detector with SDManager, with the same path creates serious problems. Warning "sensitive detector already registered" signals this.

This segfault started after the change in ConstructPMT from June3. 
However, with the older code, segfault happen when switching the geometry wrt to the default one. "sensitive detector not found" because "aWCPMT" already existed from the defaultGeometry, but the pathname should change.

Current fix works in all cases: aWCPMT is only defined ones and the sensitive detector is always set to to the proper pointer. If it finds one with the same pathname, it just properly sets the pointer, otherwise it makes a new sensitive detector with the right path. Code has been thoroughly tested and fixes this crucial bug.


eosullivan:
Ok, seems reasonable. I don't think there are currently any options in the macro that go with the SK geometry and require the /WCSim/Construct line to be implemented, but this could become a problem as new options are included (or if we ever change the default geometry). Thanks for fixing this.

PR #130, opened 2015-11-10
Fix/visualize hits

tfeusels:
There were two places remaining in the source code that still had old check wrt glassFacePMT and weren't adapted to the name change. This resulted in hits not being shown by the Geant4 visualization if one does /vis/scene/add/hits.
Checking for substrings solves these.


eosullivan:
Great, I did notice this when I was trying to put together some figures for the HK design report recently, but I didn't have time to track the bug down. It would be nice to make it more flexible at some point and not require the SD to contain the string "glassFaceWCPMT", but this can be changed later.

src/WCSimConstructGeometryTables.cc
```diff
@@ -69,6 +69,7 @@ void WCSimDetectorConstruction::GetWCGeom
     }
 
     if ((aPV->GetName() == "WCCapBlackSheet") || (aPV->GetName() == "glassFaceWCPMT")){
```
eosullivan: I'm getting an error when I compile this related to an extra {. I think it's because you meant to remove this if statement and replace it with the one below.

tfeusels:
Sorry about that. Compiles now.

eosullivan:
Ok, this looks ready to go!

PR #133, opened 2015-11-16
Add WCSimEnumerations to the WCSimRoot.so library

tdealtry:
It was brought to my attention (thanks @brichards64) that ROOT macros using WCSimEnumerations weren't running correctly in non-compiled mode. This PR fixes this by modifying the makefile and LinkDef.h. This also required splitting up the WCSimEnumerations.hh source file - method implementations are now in WCSimEnumerations.cc


tfeusels:
@tdealtry If pull request #125 is merged first. Could you also update L35 and L39 in the CMakeLists.txt correspondingly (the header in L35 and the .cc in L39)? This would make sure compilation through cmake still gives the same behaviour. Thanks!

tdealtry:
Just to clarify what the issue is...
* When running in non-compiled mode, macros will run correctly. However, if you want to use one of the enumerations (e.g. to only select events that have passed a certain trigger type) this won't work with the current code. This PR adds the relevant information to `libWCSimRoot.so` for it to work.
* When running in compiled mode, you can use the line `#include "WCSimEnumerations.hh"` and it works with the current code.

Issue #81, opened 2015-06-08
Restructure how our macros are organized


eosullivan:
We currently have two default macros in WCSim: novis.mac and vis.mac. These macros contain all possible detector configuration options, as well as any options which can be controlled with the detector messenger. We should think about restructuring to have a simple default macro file, which can then include other macro files to perform more complicated functions. This would mean having a larger catalogue of macros (for example, having a new macro for every new detector configuration) and either having a large macro with all the possible commands (but which is not the default macro) or utilizing the wiki to explain all the available commands. This could be a potential fix for issue #79 and is related to issue #80.

tdealtry:
#142 looks good. Can I suggest in a future PR having a "macros" (or similar name) directory where all the macros (probably other than WCSim.mac) can live. There are already a lot of macros (and presumably lots more will come into existence) - having a specialist directory will keep the root WCSim directory much clearer & cleaner.

eosullivan:
Yes, agreed. In the future, a separate directory for the macros is probably the way to go.

PR #135, opened 2015-11-18
Document the G4 General Particle Source option in the *vis.mac files

tdealtry:
I wasn't aware of the /gps/ options, and I know of at least one person who was using it incorrectly, therefore some documentation in *vis.mac is handy.

We may also want to change (or overload) the word "laser" with "gps" to make it clearer.


tfeusels:
I totally agree!
It would also be convenient to change the name from "normal" to "singleParticle" then and from "muline" to "nuance". This would create some more clarity for the user to use gps and gun. 
Could this be merged after the name changes?
@tdealtry, can you explain for later reference what the incorrect use was?

tdealtry:
The obvious issues, that are simple to fix with good documentation:
* Using the /gps/ commands with the 'normal' mode
* Mix-and-matching /gun/ and /gps/ commands

tfeusels:
Thanks. We just ran in the first one too because of the name "laser".

cvilelasbu:
Yes, I never understood why the gps option was called "laser". The only other thing that I found in the code when I first came across this feature that is affected by using "laser" vs "normal" is the following piece of logic in ```src/WCSimPrimaryGeneratorAction.cc``` lines 227 - 265:
```
else if (useNormalEvt)
...
G4double E         = std::sqrt((P.dot(P))+(m*m));
...
else if (useLaserEvt)
...
G4double E         = std::sqrt((P.dot(P)));
```
This seems redundant to me, but maybe this should be fixed too if we call the gps option something more useful than "laser".

cwwalter:
> Yes, I never understood why the gps option was called "laser". The only other thing that I found in the 
> code when I first came across this feature that is affected by using "laser" vs "normal" is the following 
> piece of logic in src/WCSimPrimaryGeneratorAction.cc lines 227 - 265:

The reason it was called laser is this is how Tarek implemented the laser simulalation for comparison with calibration data for tuning the water parameters.  It's all it was used for.

cwwalter:
BTW that is why that code is there.  The mass of the photon is zero.

tdealtry:
@eosullivan, would you like the proposed changes to be added to this PR, or made into a new one:
* normal -> singleParticle
* muline -> nuance
* laser -> gps

I guess initially overloaded meanings would be the best way to go, and put some kind of deprecated warning in the documentation. What do you think? Also, let me know if there are suggestions of better terminology.

Regarding the code @cvilelasbu mentioned, it seems like a
```c++
G4double m       =anEvent->GetPrimaryVertex()->GetPrimary()->GetMass();
```
line would work, even for massless particles, so I'd be in favour of 
```c++
else if (useLaserEvt)
...
G4double E         = std::sqrt((P.dot(P))+(m*m));
```
being used. But let me know if you prefer 'laser' to be kept explicitly massless.

I just ran some tests with multiple gps electrons per event, and all true particles are saved in the output file in positions 2 to 2+N (which is good), and the 0th particle has the wrong mass/energy (as expected from the current code). But I'm not sure if we want to do anything special about the '0th' (i.e. beam) particle in this situation? Currently it chooses just a random gps particle (i.e. any in the range [0,N-1]) as the 'beam'. Should we enforce it to choose gps 0? Do some averaging? Or just document that 'the 0th particle for gps events with multiple initial particles should not be used'?

cwwalter:
Dear All,

I'm away for Christmas this week and Erin and I haven't been at Duke together for a few weeks to talk about all of these suggestions.  So we will get to them when we can see each other.  But just a few notes now while I remember (and then no more until I get back).

> - normal -> singleParticle

If you want to rename this I would call it 'gun' since this is what it is.

> - muline -> nuance

muline is not nuance format.  It is a very reduced format (which is basically almost never used anymore).  It would not accept real nuance files like the vecfile option (which is in nuance format).

> - laser -> gps

If you want to make this change you need to document how to make lasers and we should also find the macros for calibration (which we will need soon and uses this).  You need to specify the laser properties.  It is more involved then just setting a single particle.

Happy Holidays!

eosullivan:
Following on from Chris' suggestions, I wonder if we should just add in the gps options as a separate thing from the existing laser option. We'll be redoing the tuning parameters in the near future to remove all of these fudge factors that are in the code, so we don't want to lose that capability entirely (or make it hard to revert back).

tdealtry:
Ok, so to summarise the discussion so far, we want to modify the "/mygen/generator" options to make them clearer:
* normal -> gun
* muline -> vecfile
* new option: gps (essentially identical to the laser option, but with the energy calculated by including particle mass)

We also need to think of what the 0th particle in the case of multi-particle gps should be.

@eosullivan would you like these changes in this PR, or in a new one?

JostMigenda:
@tdealtry Just noticed that vis.mac gets re-added by this PR; apparently one of the merges didn’t work as intended. Could you please remove that file?

I’d vote for renaming the `/mygen/generator` options in this PR, rather than a separate one, since all the discussion on why/how to rename is contained here.

tdealtry:
Good spot @JostMigenda. Sorted!

@eosullivan, shall I add the renaming & new 'gps' option to this pull request?

In terms of what to do about the 0th particle, I think this can be done properly when #171 is merged. Although I don't understand all the detail about this yet.

JostMigenda:
Regarding the 0th particle: If I understand you correctly, the unwanted behaviour is already present in the current WCSim code (prior to this PR) and not directly related to the renaming? In that case, I would propose splitting that into its own PR.

tdealtry:
Yeah I agree that the 0th particle stuff is own pull request (to be submitted after #171 is merged). I should of been more explicit.

eosullivan:
Sure, make this all one PR. But, to be clear, I think muline should just stay as muline? Reading over @cwwalter's comment again, I don't think he was suggesting you change muline -> vecfile, just saying that you can't use the nuance format for muline like you could for vecfile. At some point in the future, we could depreciate muline. For now, maybe just add a comment that says it's a reduced format that is hardly used anymore.  

So, this PR would change normal->gun and add in a documented gps option (and put back the positioning of these lines to a reasonable part of the macro).

tdealtry:
Done & tested

Just to clarify what the current 0th particle is...
* A random entry in the list of gps particles
* In the `laser` case, takes the momentum as the 0th energy & momentum
* In the `gps` case, takes the energy as the 0th energy & momentum

I think #171 brings in the facilities to make this something more sensible than a random entry.

eosullivan:
Could you turn this problem regarding the 0th particle with multi-particle gps into an issue on the Issues page? Then we won't forget to revisit this after #171 gets merged.

WCSim.mac
```diff
@@ -107,15 +101,39 @@
...
+#/mygen/generator gps
+#/gps/particle e-
+#/gps/energy 500 MeV
+#/gps/direction 1 0 0
+#/gps/position 0 0 0
+
+# Or you can use the laser option
+# This is equivalent to the gps command, except that the gps particle energies are saved ignoring their mass
+# It is used for laser calibration simulation
+#/mygen/generator laser
```
eosullivan: This is an improvement over the current WCSim.mac, which doesn't mention the laser option at all, but for completeness maybe we should include the commands to run laser like we have for the gps and gun above?

tdealtry: Sure, I can put this in, but what example should I add? The single electron particle gun doesn't make sense, due to: `gps particle energies are saved ignoring their mass`. That's why I put the `This is equivalent to the gps command` statement there - it may need rewording but it means that the `/gps/` commands above work just the same.
I guess I could do a single 500 MeV gamma? Or 1000 optical photons?
What do you think?
When #81 is addressed, I suppose there could be an `inputs.mac` that has a more comprehensive range than the simple examples shown in `WCSim.mac`

eosullivan:
Sure, I guess running electrons makes no sense here. Let's do optical photons since that is how I imagine we will use this for tuning.

tdealtry:
Ready to go!

PR #137, opened 2015-12-13
Fix Geometry overlaps

tfeusels:
The issue: the default SuperK geometry and the latest HyperK 40% 60m tank have geometry overlaps.
The latter has 14 overlaps of the PMT volume with the BorderCells. Overlaps confuse the G4Navigator and result in wrong photon tracking and yields.
The main issue was tracked down to the height of the BorderCells in the BorderRing. PMTs placed in the corner overlapped with edge of the (Extra)SolidBarrelCell. 
To conserve the distance between PMTs, the PMTs should only be placed in the barrelCellHeight region, hence the SolidBarrelCells were enlarged. To make sure the puzzle still fits in the WCCapAssembly Cylinder, the WCCap Z position had to be lowered, as well as the WCCapBlackSheet Z. The Cap PMTs are anyway placed in the BarrelPMTOffset region on top of the BlackSheet. The extra "PMTExpose" part of the WCCap was unnecessary. 
Attachments:
- Bugs are illustrated by: 
  overlapnearborderbug_superkgeom and (the WCCapEdgeLimit bug from SuperK geom):
  overlapbug2
- The bugfix is illustrated by: 
  overlapbugsolved
  and
  overlapbug2_solved_superk
- The detailed technical drawing of the fixes (not all to scale) is [OverlapBug.pdf](https://github.com/WCSim/WCSim/files/60297/OverlapBug.pdf)

Commit details:
- Add flags for checking overlaps for all placements in ConstructCylinder. Optionally switch off for faster running.
- Add flags for optionally building only parts of the cylinder. Useful for geometry overlap debugging.
- Fix overlaps from PMT with BorderCell for j = 0 by only placing them in the borderCellHeight z region. Had to extend the BorderAnnulusZ. Solves all overlaps for the 40% HyperK tank.
- Increasing the BorderRing height requires the WCCap to be decreased in height so that they both still fit together in the CapAssembly. This
  results in a cap that is now BarrelPMTOffset high instead of BarrelPMTOffset + PMTExpose.
- Also CapBlacksheet had to be adjusted according to the above change in Z.
- SuperK geometry overlap fixes: WCCapEdgeLimit is now consistent with the other geometries, ie. WCIDRadius - WCPMTRadius.
  WCBarrelPMTOffset for SuperK geometry has to be larger or equal than the PMTExpose, cfr. Tutorial:-How-to-make-your-own-detector. Now consistent to be
  PMTRadius as used in other geometries. Probably the definition of WCBarrelPMTOffset has changed since defined as 0.0715 m (< 0.18 m PMTExpose).

=> After these fixes, NO more Overlaps were found for SuperK geometry, HyperK 14% with 60m tank and HyperK 40% with 60m tank.


src/WCSimConstructCylinder.cc
```diff
@@ -90,6 +90,18 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructCylinder()
   if (WCAddGd)
   {water = "Doped Water";}
 
+  // Optionally switch on the checkOverlaps. Once the geometry is debugged, switch off for 
+  // faster running. Checking ALL overlaps is crucial. No overlaps are allowed, otherwise
+  // G4Navigator will be confused and result in wrong photon tracking and resulting yields.
+  // ToDo: get these options from .mac
+  checkOverlaps = true;   
+  checkOverlapsPMT = false;
```
eosullivan: Should the checkOverlapsPMT be true by default? The places where it shows up are true in the current code.

src/WCSimDetectorConfigs.cc
```diff
@@ -29,13 +29,13 @@ void WCSimDetectorConstruction::SetSuperKGeometry()
   WCPMTRadius = PMT->GetRadius();
   WCIDDiameter          = 33.6815*m; //16.900*2*cos(2*pi*rad/75)*m; //inner detector diameter
   WCIDHeight            = 36.200*m; //"" "" height
-  WCBarrelPMTOffset     = 0.0715*m; //offset from vertical
+  WCBarrelPMTOffset     = WCPMTRadius; //offset from vertical
```
eosullivan: For most of the detector configurations, setting the WCBarrelPMTOffset = WCPMTRadius (or even to the expose height of the PMT?) and WCCapEdgeLimit = WCIDDiameter/2.0 - WCPMTRadius is a good approximation. But for this SK configuration, I think these numbers were taken from the SK NIM paper and are supposed to represent what the actual detector looks like. I just did a small test where I set theWCBarrelPMTOffset = 0.0715*m and  WCCapEdgeLimit = 16.9*m in the SetSuperKGeometry() and there didn't seem to be any overlap warnings, so perhaps it's best to leave these values as they were.

tfeusels: That surprises me, because the drawings shown in this pull request illustrate that there is an overlap when I was running the default SuperK geometry and why the change is necessary.
Did you run with both checkOverlaps == true?

src/WCSimConstructCylinder.cc
tfeusels: It should when people want to make new geometries to clearly signal any issues, while for current geometries, they should be off or it will generate a lot of output and will seriously slow down running WCSim. Essentially the WCCap has a lot of overlap checks that only need to be done when making a new detector geometry.

src/WCSimDetectorConfigs.cc
eosullivan: Ah, ok. I changed the wrong configuration when I did my test. So, yes, there are overlaps when the WCBarrelPMTOffset and WCCapEdgeLimit are set to the original values. I will try and see if I can track down these original numbers. Setting WCBarrelPMTOffset = WCPMTRadius and WCCapEdgeLimit = WCIDDiameter/2.0 - WCPMTRadius is an easy fix, and probably won't make an impact on the physics. However,  everything in this configuration is supposed to be set to the actual physical values in the SK detector (and we don't have any overlaps in the actual detector!) so it would be nice to know why these original values aren't working (are these not actually the physical values in the tank or are they not being implemented properly in the code?)

tfeusels: I did not check whether the number of PMTs are different though, which could change the impact on physics. According to my figure, those two PMTs are not allowed to be there, but I'm not sure whether the PMTs on the Cap are reordened or whether there will be less PMTs with the geometry overlap fix.
I agree that this is strange and should also be understood where the numbers come from. I thought a different definition of WCBarrelPMTOffset could cause this? 
Also, isn't there a database with the exact xyz-location of the SK PMTs, which will then be the ideal comparison with how they are programmed in WCSim?

eosullivan:
I have tracked down the first commit that sees these overlap warnings, and it's from 2012 and was put in to separate the ConstructPMT method from the WCSimConstructWC file (details are here: https://github.com/WCSim/WCSim/commit/33863b43431a9e95001ddb1a3a426528fed65815). I suspect some offset wasn't properly copied over to the new ConstructPMT.cc file, so if you want to understand what change is causing the problem (and what the proper fix would be) I would start there. The tuning from Alex looks like it was done after this problem was introduced, so when we find the source of the overlaps we'll have to do some tests to see whether we can fix it without messing up the existing tuning parameters, or if we should wait until we redo the tuning later this year.

tfeusels:
We can split this pull request in two to go forwards: while someone (else) looks for the proper underlying fix for the SuperK overlaps (which I'll put in again and make into a separate issue), the other issue causing overlaps for HyperK geometries can be merged in. 
Anyway, current SuperK geometry (even with tuning) is buggy near the cap edges for those few PMTs so I'm not sure how much you can trust the tunings for those PMT responses.

eosullivan:
The problems with Super-K and Hyper-K are probably not so decoupled. If we fix whatever happened in https://github.com/WCSim/WCSim/commit/33863b43431a9e95001ddb1a3a426528fed65815 it theoretically could fix both the Super-K and Hyper-K problems. I also don't want to arbitrarily modify the WCSimWCCylinder file to fix the Hyper-K geometry since this is also used by the Super-K geometry.

tfeusels:
#33863b4 essentially moves the PMT geometry construction code to a separate file as far as I can see. As I showed in the drawing, it's the edge construction of the cylinder itself which is buggy. So I don't see how those two bugs are related. The only way you can squeeze a PMT in that corner is if the PMT is not constructed a cylinder LogicalVolume, but as  its actual shape. Maybe that changed? 
Except for being able to place daughters in the mother volume in ConstructPMT, I don't understand why that cylinder mother volume is required. We could replace the mother by the actual PMT glass outer boundary as mother and maybe then it will fit in the current corner of the Cylinder, but this will highly depend on the chosen number of PMTs in a cell, so is maybe not a general solution. Otherwise, I believe one has to use my proposed solution if one sticks to cylinders for PMTs.

eosullivan:
Ok, it took me some time to convince myself that this won't adversely affect any of the other detector configurations, and that it only affects logical volume placement, but that does seem to be the case. Looking through the code, the placement of the PMTs are not relative to the logical volume placement and comparing the geofile.txt output from your branch with the main version of WCSim, I see no difference in the placement of PMTs. There are event-by-event differences when I run the code from this pull request and compare it to the current develop branch, and even when I run many events I can't really convince myself that the hits distribution is statistically consistent (see figure below: cylinder w/ 40% photocoverage, your PR is in red, current code in black). This change is small however (~1% change to the mean, at most), and is perhaps related to photon tracking in the corners now being done correctly. 
pr137_comp

eosullivan:
Long story short, go ahead and resubmit this pull request without the changes to the SK geometry (we are going to be tagging out this next release tomorrow, so if you do it by then it will make it into this next release). Make an issue on the tracker about the SK overlap problem.

tfeusels:
Just remember that taking out the SuperK bugfix results in a G4Exception because SuperK is build first as DefautlGeometry. The WCCapEdgeLimit bug in SuperK, illustrated above, is not the reason and can be post-poned to another pull request if you want. The WCBarrelPMTOffset on the other hand has to be WCPMTRadius or one gets this error:
Checking overlaps for volume WCExtraTowerBlackSheet ... OK! 

-------- EEEE ------- G4Exception-START -------- EEEE -------
*** G4Exception : GeomSolids0002
      issued by : G4Polyhedra::Create()
Illegal input parameters - WCCapMainPart
        R/Z segments cross !
*** Fatal Error In Argument *** core dump ***
-------- EEEE -------- G4Exception-END --------- EEEE -------


*** G4Exception: Aborting execution ***

eosullivan:
@tfeusels, will you be in Japan this week for the Hyper-K meeting? If so, let's talk about this there.

tfeusels:
@eosullivan, unfortunately not. 

WCBarrelPMTOffset is used for the construction of the Cap geometry and has to be large enough (see OverlapBug.pdf attached above). Therefore for the SuperK geometry it has to be increased as well. I don't think this will change the SuperK PMT position and should stay with this pull request. 
The WCCapEdgeLimit, however, is used by the Cap filling and placement. Fixing this for SuperK will avoid the overlap, but (has to) change the placement of the Cap PMTs. As you want to investigate the SuperK geometry in more detail for a different fix, I moved that part of this pull request to issue #144.

eosullivan:
Right, I guess if the WCBarrelPMTOffset is too small, the PMTs on the cap will poke out of the logical volume of the cap (which is what's happening with the SK geometry). 

This pull request does change the placement of the barrel PMTs for the SK geometry, though. The PMTPosition vector has barrelCellHeight in it, where barrelCellHeight = (WCIDHeight-2.*WCBarrelPMTOffset)/WCBarrelNRings. So, by increasing WCBarrelPMTOffset, you are squishing all of the barrel PMTs a bit closer together and leaving a bigger gap at the top and bottom of the barrel where there are no PMTs. You can see this if you compare the geofile.txt (the file which shows all the PMT positions that gets generated each time you run WCSim). We have made some effort to make the SK geometry realistic to the actual detector and the space at the top and bottom of the barrel shouldn't be changed unless we know what this value should be. 

So, for this pull request to be merged in, it either needs to fix the overlaps without changing the PMT placements, or it needs to fix the SK geometry with the actual values in the detector.

tfeusels:
What about using the current numerical value in barrelCellHeight, ie. replacing the WCBarrelPMToffset variable? That way both the BarrelPMToffset can be changed, fixing the logical volumes, and we keep the PMT positions? 

It would also be helpful if the real, measured SK geometry (and PMT positions) is available somewhere for HyperK members.

eosullivan:
Ok, I'm going to hold off on making the next release for a few more days. I would like to think through this solution a bit more and get your PR #84 in (and perhaps #125 too).

eosullivan:
I'm not following how barrelCellHeight would be determined without using BarrelPMToffset. The main purpose of BarrelPMToffset is to tell the simulation how much room you want on the top and bottom of your cylinder where no PMTs will be built, which has to be accounted for when you want to figure out the height of each PMT cell. Are you suggesting adding in an extra variable somewhere?

spradlin:
@tdealtry The geometry fixes proposed in this discussion were incorporated into the geometry fixes implemented for https://github.com/WCSim/WCSim/pull/347.  This PR can be closed without merging.

Issue #80, opened 2015-06-08
Merge novis.mac and vis.mac to make a single default executable


eosullivan:
We currently have two macros (novis.mac and vis.mac) which differ only in that one of them has the visualization turned on. We should make a single, simple macro where we can turn on and off the visualization using a command. Thanks to @fdilodov for suggesting this.

eosullivan:
Solved by #142

PR #142, opened 2016-01-25
Feature/issue80

eosullivan:
We currently have two main macros in our git distribution: novis.mac and vis.mac. When a new feature gets added that has a macro option, we need to update both these macros to reflect the new available option. This pull request consolidates novis.mac and vis.mac into a single file called WCSim.mac (this solves #80 and is the first step to fixing issue #81). 

**Changes in the Pull Request**
* The visualizer options that were in vis.mac have been moved to a file called  visOGLSX.mac. 
* The pre-existing visRayTracer.mac has been modified to just contain the RayTracer options rather than act as a stand-alone macro.
* novis.mac has been renamed WCSim.mac and lines to turn on the visualizer have been added. In order to turn on the visualizer,  /control/execute visOGLSX.mac or /control/execute visRayTracer.mac can be uncommented in the WCSim.mac file.


eosullivan:
I just did a quick round of verification runs. Running WCSim.mac gives the same output as novis.mac. If I turn on the OGLSX options the visualization looks the same as running with vis.mac in the old code. Running with the RayTracer options outputs the same figures as visRayTracer.mac did before.

Issue #141, opened 2016-01-21
Incorrect detector design in WCSim


YujiOkajima:
Current design of cylinder detector for Hyper-Kamiokande in WCSim ("Cylinder_60x74_20inchBandL_14perCent" and "Cylinder_60x74_20inchBandL_40perCent") has inner detector with 74 m diameter and 60 m height. However the candidates of detector design for Hyper-Kamiokande has inner detector with 70.8 m diameter and 54.8 m height.
The whole water tank volume seems to be taken to inner detector volume when the design is 
introduced in WCSim.
Therefore it is necessary to fix the geometry.

eosullivan:
Ok, thanks for bringing this up. Where are you getting these numbers from? Is there a technical drawing somewhere? Are these numbers set or are they still changing?

YujiOkajima:
The diameter and height has been already written in design report (Figure. 35).
So I think this design will not be changed.

JostMigenda:
I changed the dimensions in pull request #151

PR #151, opened 2016-02-11
update HyperK configurations to reflect current Design Report

JostMigenda:
Changed ID dimensions according to Fig. 36 in Design Report February 7th, 2016 (https://wiki.hyperk.org/DesignReport/Drafts)
This should fix issue #141


JostMigenda:
Should we also rename the Cylinder_60x74 configurations to “HyperK” (and rename “SetHyperKGeometry” to “SetPrevHyperKGeometry”, or similar)? Or is it still to early for that?

YujiOkajima:
Because the interval of PMTs in 40% tank has been already written in design report,
I think it will be better to make geometry based on the design.
The interval between PMTs in requested geometry becomes 71.3 cm, and it is set to 70.3 cm in design report (figure.37).

JostMigenda:
Thank you for pointing this out!

I just noticed that page 78 of the design report (section D 4.2) gives an interval of 70.7cm (70.84 horizontally) between PMTs. I will write to the TOTF to ask for clarification and report back as soon as I receive an answer.

eosullivan:
Yes, this should be part of a larger campaign to rename things so it's clear the Hyper-K geometry is the cylindrical design that is reflected in the design report. Perhaps a good name for the current SetHyperKGeometry would be SetEggShapeHyperKGeometry or something.

JostMigenda:
This seems like a good way to get an overview of how WCSim is structured, so I’ll try to tackle that in the next few days.

JostMigenda:
As far as I can tell, the renaming campaign is now completed; I went with “EggShapedHyperK”, thanks @eosullivan for the suggestion!

No update yet regarding the interval between PMTs.

eosullivan:
Ok, let's leave off changing the PMT spacing for now (this is probably a minor effect anyways). Could you check in the geofile.txt file to make sure that the PMT spacing is close to 71 cm? 

Also, a couple of minor things to change before we merge this in:
- The naming of Cylinder_60x74_20inchBandL_40perCent doesn't really make sense if it's not a 60x74 cylinder anymore. I hate having a bunch of configurations that are defunct, but maybe for now let's leave the 60x74 cylinder as it is so users can still go back and use it if they want to compare. Let's make this new configuration in the WCSimDetectorConstruction::SetHyperKGeometry() part (in the same way as we specify the SK configuration in the SetSuperK() part). 
- Add the SetHyperKGeometry() option to WCSim.mac

JostMigenda:
The PMT spacing in geofile.txt is 70.7 cm. (And those two minor things are now done.)

WCSim.mac
```diff
@@ -17,6 +17,9 @@
 #/WCSim/WCgeom SuperK_12inchBandL_15perCent # Note: the actual coverage is 14.59%
 #/WCSim/WCgeom SuperK_20inchBandL_14perCent # Note: the actual coverage is 13.51%
 
+# HyperK (one cylindrical tank with a height of 54.8 m, a diameter of 70.8 m and 40% coverage)
```
eosullivan: You should include in the comment that the 20" B&L photosensors are used.

src/WCSimDetectorConfigs.cc
```diff
@@ -148,8 +148,8 @@ void WCSimDetectorConstruction::Cylinder_60x74_20inchBandL_14perCent()
   WCPMTName           = PMT->GetPMTName();
   WCPMTExposeHeight   = PMT->GetExposeHeight();
   WCPMTRadius         = PMT->GetRadius();
-  WCIDDiameter          = 74.0*m;
-  WCIDHeight            = 60.0*m;
+  WCIDDiameter          = 74*m;
```
eosullivan: Can you include the .0 in WCIDDiameter and WCIDHeight for these two cylinder configurations? I want them to disappear from the diff.

eosullivan:
Great, I'll merge this in! @cvilelasbu, did you have a chance to test this modified configuration with fiTQun?

cvilelasbu:
Hi @eosullivan, I had a quick look. Other than a slight worsening of the momentum resolution and bias I didn't see anything glaringly wrong. The differences are at the < 1% level, but given that we are still trying to iron out some reconstruction effects in the nominal geometry I would not recommend reconstructing events generated with the new geometry description using the fiTQun tunes for the old one.

PR #153, opened 2016-02-23
Feature sk4trigger

jeimber:
SKIV-like trigger implementation as part of #13

### DIFF
--- a/include/WCSimEnumerations.hh
+++ b/include/WCSimEnumerations.hh
@@ -8,6 +8,10 @@ typedef enum ETriggerType {
   kTriggerUndefined = -1,
   kTriggerNDigits,
   kTriggerNDigitsTest,
+  kTriggerNDigitsSK4_SHE,
+  kTriggerNDigitsSK4_HE,
+  kTriggerNDigitsSK4_LE,
+  kTriggerNDigitsSK4_SLE,
   kTriggerFailure // this should always be the last entry (for looping)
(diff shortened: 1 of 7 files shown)


tdealtry:
Hi @jeimber this looks good.
A somewhat substantial comment - I think it'd be beneficial to use WCSimWCTriggerBase::FillDigitsCollection(), instead of making a copy with only a couple of minor changes. From a quick scan through, it looks like:
* "Offset" should be a .mac parameter (with the default set to 950 ns for all SK electronics)
* The way to set the time range for each trigger should be standardised. Because of the overlapping-ness of the 4 SKIV triggers, the old method can't be used. Therefore other triggers should fill the new vectors TriggerStartTimes and TriggerEndTimes (which should be reset correctly, i.e. in ReInitialize())

jeimber:
Hi @tdealtry 
I agree with your comments.
It would be best to reduce the amount of repeated code where possible (now that i have a better idea of the modifications necessary) and handle the vectors in a clearer fashion.

eosullivan:
@jeimber, where are we with this pull request? It seemed from how it was left off in this comment thread that there was still some work needed to address the issues brought up by @tdealtry.

eosullivan:
@jeimber, since this PR is very behind and still needs work, I'm going to close it for now. It will still be accessible in the closed pull request tabs if we need to look at it (or reopen it) in the future.

PR #160, opened 2016-03-30
Fix the PMT-specific default dark noise values

### LINKED ISSUE #159: Default dark noise options don't work for non-SK geometries
Unless I'm missing something obvious, #101 only works if you don't change from the default geometry (i.e. SK). This is because the dark noise options are read before the geometry is actually constructed.
There's (at least) a couple of ways this could be fixed:
 1. Get the default PMT rate the first time the AddDarkNoise() method is called, only if the messenger hasn't overwritten a (new) unphysical default value.
 1. Create the WCSimWCAddDarkNoise class instance in the same way as the trigger/digitizer options (i.e. in WCSimEventAction::CreateDAQInstances()). Will need to store the DarkNoiseMessenger in the EventAction, and do the (somewhat laborious) job of passing of DarkNoise options from EventAction to AddDarkNoise.

Can someone please confirm the bug? And let me know which solution preference you have? 1 is much simpler to implement, and so is my preference.

tdealtry:
Fixes #159 
I also modified the .mac file - it will pick up the PMT-specific defaults by default now. I can revert the last commit if you don't want it @eosullivan


eosullivan:
Sure, it seems reasonable to include these macro changes in the pull request so that it is clear you no longer need to set the values in the macro when changing photosensor types (and so you aren't explictly specifying the 20" PMT dark rate in the defualt macro, which would overwrite the changes we are trying to make here). Could you do some basic verification to show that this PR automatically grabs the right dark rate values from the photosensor? Maybe a nice quick test would be to simulate HK and show that the right dark rate values for the B&L tubes are now being set automatically without needing to set them in the macro. And also confirm that setting the value in the macro still changes the dark rate since we don't want to lose that functionality.

tdealtry:
I've added a few .mac scripts (for a variety of combinations of detector/user-set/PMT-default) & a quick exit - give `$WCSIMDIR/testing/run.sh` a go, let me know when you're satisfied it's working as expected, and I'll revert these two commits

eosullivan:
Ok, so I ran some verification tests. First I used your macros, which cout the dark rate and the convert rate to the screen, to confirm that the PR seemed to be doing what it said it should do. Everything looked fine. 

I wanted to also see it a bit more visually to make sure the final product (in this case, simple outputs like the number of hits) was acting as expected. I'll include these test plots here for posterity.  

Test 1: Compare the 40% photocoverage cylinder with no dark rate set in the macro for the develop branch vs. no dark rate set in the macro for this PR. The hits in the develop branch are less than in hits in this PR, which is consistent with the dark noise in the develop branch being lower than that in this PR (which we would expect from the B&L dark rate being set as the PMT dark rate). 

test_defaults

Test 2: Compare the 40% photocoverage cylinder with the proper B&L dark rate and convert rate set in the macro for the develop branch vs. no dark rate set in the macro with this PR. We get identical results, which suggests that this PR now properly sets the default B&L dark rate parameters. 

test_proper

Test 3: Compare the 40% photocoverage cylinder with no dark rate set in the macro for this PR vs. the dark rate set to 3 Hz (lower than the default) in the marco for this PR. We set the number of hits decreasing as we set the dark rate lower in the macro, which suggests that we are still able to set the dark rate manually in the macro. 

test_macrotest

Test 4: Run the SK configuration with no dark rate set in the macro for this PR vs. with no dark rate set in the macro for the develop branch. I only ran ten events, but I confirmed that I saw no change (as expected, since the SK configuration dark rate was not affected by this bug). 

So, long story short, I am satisfied that this is doing as it should. Revert the last two commits and we'll get this merged in.

tdealtry:
Great, thanks. It's ready to go.

PR #163, opened 2016-05-17
 /mygen/vecfile not found

tdealtry:
When /mygen/vecfile points to a file that doesn't exist, or isn't called when using muline mode, there should be a fatal error (rather than a warning that can get lost in the output)


eosullivan:
Good idea to include this error message so it is clear when the vector file is called incorrectly in the macro. I just did a quick check to see that this does what it should, and it seems to be working. I'll go ahead and merge this in.

PR #164, opened 2016-05-24
G4cout, G4cerr, G4endl

tdealtry:
It's recommended to use the Geant4 versions of iostream methods, and I noticed a few instances where the std library versions were used. So I used grepped search/replace to:
* Swap the remaining instances of cout, cerr, endl for G4cout, G4cerr, G4endl
* Remove the iostream includes


eosullivan:
Great, thanks for the clean up!

Issue #138, opened 2016-01-18
Clean up WCSimEventAction comparing HitCollection and DigiCollection


tfeusels:
L.750 to L. 834 of WCSimEventAction.cc is completely deprecated: This part loops over both HitCollection WCHC and WCDigitsCollection WCDC_hits (hits with dark noise added) to be able to add parentIDs to the hits in WCDC_hits and to set the parentID of the pure noise hits to -1. However, since the recent major DAQ upgrade, parentID information is included and properly filled in the DigitsCollection WCDC_hits. A simple cout illustrates this. That part can therefore be removed and replaced by a simple loop over WCDC_hits, filling CherenkovHits of WCSimRootEvent.

PR #165, opened 2016-05-26
Fix saving of raw hit information

tdealtry:
When looking at fixing #138, I found bugs in the current implementation of saving raw hit information (specifically WCSimRootHitTime's)
1. The current method matches times between the hits saved in "WCRawPMTSignalCollection" (a digits collection, containing photon & noise hits) and a hits collection (containing photon hits only). As of #121, the digits collection times are smeared, but the hits collection times are not. This causes failures in the time matching, resulting in wrong times & parent IDs being saved.
2. The floating point comparison of times was not done correctly

The first commit of the pull request does as much as can be done with the current method - fixes 2.

The second commit implements the suggestion of #138 and uses only the digits collection. This required changes to WCSimWCDigi so that parent information is handled correctly when sorted by time (third commit)


tdealtry:
Thinking about this some more, I realise that the hit times in the output file are now PMT-smeared. This was not the case before #121. Please let me know whether this is acceptable, or if you'd like the pre-smeared times in the output. (Or perhaps both?)

eosullivan:
Is there some sensible way to set up the times such that they are not smeared? Here was the original intent of the collections. There would be a hits collection that would record the PMTs that were hit and the true time, an intermediate collection would apply the effects from the PMT, and a final collection that would include all electronics effects. Then, the default write out to the ROOT file would be the true information (after simulation) and the final collection with all electronics effects included. If we think there is some useful information that is in the intermediate collection that we can't get from the final collection, we should write it out separately rather than removing the hits collection information.

tdealtry:
I'll set up the intermediate collection to have timing information of both pre- and post-smearing, and save the pre-smearing into the output file. It'll then be easy to add the post-smearing times to the output file (if required).

tdealtry:
Done.
To check it's doing as expected, I used `verification_HitsChargeTime.C`. Since this pull request fixes known bugs, it can't be directly compared to the head. Therefore I compared `36b1e4d931a89bf0d4484b0726c4c04e46b837b2` (i.e. the version with the bugs fixed, saving post-smeared hit times) with `9f310614c41142ee7db39f653517b8c5dd7988f7` (i.e. the version with the new fields in WCSimWCDigi, but modified slightly to save the post- instead of pre-smeared hit times)
Exact agreement as expected.

The pull request now fixes the bugs I found, solves #138, and saves the pre-smeared hit times (going back to the behaviour before #121)

eosullivan:
I merged in a couple of pull requests today and it's lead to a conflict on this branch (I think it was just #164 that you pushed awhile ago). Could you update this pull request and then I'll do a final check and merge it in.

tdealtry:
Conflicts resolved!

PR #166, opened 2016-05-26
Add operator==() to WCSimRootEvent etc.

tdealtry:
This allows an exact comparison between information in two WCSim output files, to check for unwanted changes.
Also added an example root macro that uses these.


eosullivan:
Could you summarize the motivation behind adding this function? I'm not sure if I understand how it improves on things by looking at the diff. Also, I wonder if it might be confusing to have an operator called == since that is something that already has a standard usage.

tdealtry:
I wrote it because the current standard way of checking consistency, `verification_HitsChargeTime.C`, is incomplete - only a small subset of variables are checked. These new methods go through all the variables in order.
You might question the verbosity, but here operator==() is a standard comparison operator, returning `true` if equal and `false` if not. But I can rename it `IsEqual()`, `Compare()`, or similar. Let me know.

eosullivan:
I see. It seems like a neat feature, but maybe just needs a bit of renaming. Perhaps your macro could be called something like complete_comparison.C or full_comparison.C or something to that effect. Also, please include a descriptive header that says what the macro does. Finally, I would favour a renaming of the operator. Probably CompareAllVariables() or something like that to describe it a bit more fully.

tdealtry:
Done. Let me know if you'd like any other modifications.
I realise that there is the potential of problems similar to #182 here. I can modify by changing `std::cout` -> `G4cout` (if 1/2 are the solution) or move the `ComparisonPassed*()` functions to the `Utilities` files (if 4 is the solution).

tdealtry:
Since 3) is the solution to #182, this is ready to go as is

eosullivan:
Here, the complete_comparison.C script doesn't follow the pattern of the other scripts in verification-test-scripts and sample-root-scripts in that it isn't run inline in ROOT, but is compiled. Anything preventing it from being modified to run inline?

tdealtry:
Compiling code is good practice - clearer/better errors & the code runs faster.
But I realise that it didn't actually work with `.L complete_comparison.C+` in standard ROOT*.
I've put in a fix to make it work with `.L complete_comparison.C`, like the other macros.

*I work with a useful ROOT wrapper that automatically loads the relevant libraries and sets the paths for includes etc. This means you don't have to mess about with having to run from a specific path, worry about loading the libraries all the time, or have trouble when using multiple functions in a file. I'll put in a pull request for it (it's actually part of #88)

PR #167, opened 2016-06-03
Feature/gpu

federiconova:
- added a NoTrigger trigger, which stores all events without applying selection
- added a radioactive generator, to simulate radioactive decays in the detector water and / or in the PMT glass
- added the possibility to simulate several primary particles in each event (this allows to simulate pile up of independent radioactive decays)
- finally, added few scripts in sample-root-scripts/ to produce plots from the output

comparison


eosullivan:
Thanks for your work, @federiconova! Just a couple of things before I can start to review this. First, I noticed that some old code, like vis.mac, is included here even though it's been removed in the current development branch. Second, we usually try and keep our pull requests to be code that addresses single issues. This helps me/anyone else who is looking over the pull request because it is easier to review and catch any mistakes when the code is all dealing with a single topic. It also helps the other developers and users because they can see the changes go in incrementally and if there is a bug that gets introduced, it's a bit easier to track down.

So, I would suggest the following: check out a clean copy of WCSim (this will also help clean up your commit history) and split this pull request into a few smaller pull requests. I would do three pull requests: one for the addition of a NoTrigger trigger, one for the addition of the radioactive generator, and one for the addition for several primary particles to be simulated in an event. Then, the scripts to analyze the output could either be it's own (fourth) pull request or could be included as a part of one of the above three pull requests if it is specific to one of those issues. Does that make sense?

Issue #169, opened 2016-06-07
Fix verification_HitsChargeTime.C


JostMigenda:
Three remarks on the script `verification_HitsChargeTime.C`:

1. This one is a fairly straightforward bug:
In [line 227](https://github.com/WCSim/WCSim/blob/3b873649315c8f98018754e6a47704bc5c11ff48/verification-test-scripts/verification_HitsChargeTime.C#L227), the `for` loop should do `nevents2` iterations (which is the number of events in the 2nd root file), instead of `nevents` iterations (which is the number of events in the 1st file). In particular, if the second root file contains fewer events than the first one, this will cause an interpreter error.

2. I’m not quite sure about this one:
The `for` loop starting on line 246 iterates over the subevents of a given superevent. It calculates the average hit time and charge (in line 267/268), but they are added to the histograms only _after_ that `for` loop. _I.e._ if an event has multiple subevents, only the average hit time and charge of the _final_ subevent are added to the histograms.
This should probably be fixed by moving the curly bracket from line 269 to after line 271 — unless I’m missing something?

3. This one just seems weird to me:
Line 261:
`int tubeNumber     =  (wcsimrootcherenkovdigihit->GetT(),wcsimrootcherenkovdigihit->GetTubeId());`
This code gets the hit time, [discards it(?)](http://www.cplusplus.com/doc/tutorial/operators/#comma) and then assigns the TubeID to a variable which is not actually used anywhere. Does this code have some sort of useful side effect that I’m missing? I couldn’t find any hints what this might have been used for in the commit history.

I’ll submit a pull request for 1. shortly. Regarding 2. and 3., I’d like to wait for comments before I change/remove the code.
(Also, note that 2. and 3. apply equally to lines 144 and 160.)

JostMigenda:
One more thing to note regarding 2. is that the loop variable `index` is never actually used inside the loop, so each iteration adds up the charge/time of the first sub-event. It seems that this is done right in `sample_readfile.C`, so I fixed this accordingly in my PR.

JostMigenda:
Regarding 3., after thinking about it a few more times, it seems most likely to me that this is simply an artefact and was copied over from another sample root file. I have removed both lines in my PR.

JostMigenda:
Closed after merging #170

PR #170, opened 2016-06-07
iteration in verification_HitsChargeTime.C

JostMigenda:
This fixes part 1 of #169

### DIFF
--- a/verification-test-scripts/verification_HitsChargeTime.C
+++ b/verification-test-scripts/verification_HitsChargeTime.C
@@ -143,6 +143,7 @@ void verification_HitsChargeTime(char *filename=NULL, char *filename2, bool verb
     
     for (int index = 0 ; index < wcsimrootsuperevent->GetNumberOfEvents(); index++) 
       { 
+	wcsimrootevent = wcsimrootsuperevent->GetTrigger(index);
 	int ncherenkovdigihits = wcsimrootevent->GetNcherenkovdigihits();
 	h1->Fill(ncherenkovdigihits);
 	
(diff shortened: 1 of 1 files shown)


tdealtry:
Hi @JostMigenda Surely min(nevent,nevent2) is the safest way to loop?

JostMigenda:
Hi @tdealtry! `nevents2` is the number of events in the second root file, so I don’t see how this is in any way unsafe.

There’s an argument to be made for comparing the first `min(nevents,nevents2)` events of both files, since the whole point of that script is to verify that to different data files contain identical events.
On the other hand, if a user compares two data files with a different number of events, that probably means that they mixed up two files (or simply mistyped), just like I did when I discovered this problem. In that case, it would be best to show this problem to the user. (We could even make it more obvious by returning early if `nevents` ≠ `nevents2`.)

tdealtry:
Sorry, my mistake. I forgot how the comparison was done - it's a loop over file1 then a separate loop over file 2 (not a loop over both files simultaneously as I assumed).
Some warning message (or early exit) when the number of entries in the two files are different would be a good idea.

JostMigenda:
e004f2c...d6a921a are mostly to make the code easier to read/understand and remove unnecessary code. The early return will be added in the next commit.

eosullivan:
I don't like that we've removed some functionality of the script by removing a default filename. You could always specify your own filename, but it would be nice if it could be automatically set up to work with the electrontest.mac as specified in the README in the verification-test-scripts/ directory. If this that README is out of date (I think it is) then we should change that. Also, I think we should put out a warning message if the file sizes don't match, but not exit the program. Put it somewhere readable, like where the ks stats are output. I can see some reasons why people might use files of different sizes, like if they only care if the first event matches, for example.

JostMigenda:
I didn’t remove the default filenames – just moved them up to line 6, so we don’t have to check ourselves whether the user provides their own filenames; instead, this now gets done for us. Running the script without explicitly giving filenames still works for me, same as before.

Checking whether the first event matches is a realistic use case, you’re right. How about moving the early return to after the comparison of the first events (but before looping through all events)? I can’t think of a use case for comparing root files with a different number of events, so I would avoid doing that.

eosullivan:
Ok, an early return after checking the first event sounds reasonable, along with a descriptive exit message that says that only the first events are being compared since the file sizes do not match.

PR #172, opened 2016-06-08
Feature/radioactivity

federiconova:
this feature adds the possibility to simulate radioactive decays in detector elements (water and PMT glass). Admissible isotopes are Tl208, Bi214 and K40, but this list can be easily extended
comparison


eosullivan:
Closed with #275

Issue #175, opened 2016-06-20
(too) many ways to get the number of hits


JostMigenda:
“[There are lots of ways to get the number of hits](https://github.com/WCSim/WCSim/blob/2f7ca11b89dc893f115932186184a94614e43ac2/sample-root-scripts/read_PMT.C#L48)” and even after removing the duplicate (PR will come shortly) we have several remaining:

1. `GetNumTubesHit()`
2. `GetNcherenkovhits()`
3. `GetCherenkovHits()->GetEntries()`

and

1. `GetNumDigiTubesHit()`
2. `GetNcherenkovdigihits()`
3. `GetCherenkovDigiHits()->GetEntries()`

Can we add comments or documentation to make it clear when to use which of these three? From looking at the code, that is not obvious to me.

tdealtry:
My understanding is that 2/3 are equivalent, showing the actual number of hits. 1 is different, showing the number of unique PMTs that have been hit.
Very simple example - PMT 0 hit twice. Method one gives answer as 1; methods two/three give the answer as 2.
But I think documentation improvements are never a bad idea

JostMigenda:
That’s what I thought from the method names, too. But when I run a simple test (50 MeV electron at origin) and analyse the result with `read_PMT.C`, all three methods return the same number; which is not what I expected.

To add to my confusion, I just discovered that in `doc/DetectorDocumentation.pdf` on p. 18, both `GetNumTubesHit()` and `GetNcherenkovhits()` are described as “number of tubes with a true hit”.

tdealtry:
Looking at `WCSimEventAction::FillRootEvent()`

Hits
* `GetNumTubesHit()` filled with `WCDC_hits->entries()` - the number of hit phototubes
* `GetNcherenkovhits()` incremented with a loop over `WCDC_hits` - will also be the number of hit phototubes
* `GetNcherenkovhittimes()` incremented with a loop over `WCDC_hits`, and a loop over that PMTs' hit times - will be the total number of hits

Digits
* `GetNumDigiTubesHit()` - incremented with a loop over `WCDC`, and a loop over that PMTs' digits within that gate - will be the total number of digits in the gate
* `GetNcherenkovdigihits()` - incremented with a loop over `WCDC`, and a loop over that PMTs' digits within that gate - will be the total number of digits in the gate

(The `GetEntries()` methods are trivial alternatives to the `GetNcherenkov*()` convenience methods)

The names are confusing, so would suggest
[...]

PR #176, opened 2016-06-20
removed duplicate line from read_PMT.C, unified output

JostMigenda:
This just makes the output of a sample script a bit nicer; it has no effect on functionality.
(as mentioned in #175 )

### DIFF
--- a/sample-root-scripts/read_PMT.C
+++ b/sample-root-scripts/read_PMT.C
@@ -49,14 +49,13 @@ void read_PMT(char *filename=NULL) {
   // As you can see, there are lots of ways to get the number of hits.
   cout << "Number of tube hits " << wcsimrootevent->GetNumTubesHit() << endl;
   cout << "Number of Cherenkov tube hits " << wcsimrootevent->GetNcherenkovhits() << endl;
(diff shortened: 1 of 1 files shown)


sample-root-scripts/read_PMT.C
```diff
@@ -49,14 +49,13 @@ void read_PMT(char *filename=NULL) {
   // As you can see, there are lots of ways to get the number of hits.
   cout << "Number of tube hits " << wcsimrootevent->GetNumTubesHit() << endl;
   cout << "Number of Cherenkov tube hits " << wcsimrootevent->GetNcherenkovhits() << endl;
-  cout << "Number of tube hits " << wcsimrootevent->GetCherenkovHits()->GetEntries() << endl;
+  cout << "Number of Cherenkov tube hits " << wcsimrootevent->GetCherenkovHits()->GetEntries() << endl;
```
tdealtry: Dark noise hits are stored in this container too. Probably worth adding this to the printout for clarity (same for line 51)

JostMigenda: I agree (but would like to agree on / implement the changes you suggested in #175 first, before adding better explanations here).

PR #178, opened 2016-06-20
Feature/issue9: Changes to update GEANT version to 4.10.x

snoteptone:
Update of WCSim to be compatible with 4.10.1 (though should work w/ 4.10.2).  Most changes are for CLHEP, though some updates to physics model.  Work done by Sam Short and Francesca Di Lodovico.  Base update can be read here:
[neutroncapturesgd_geant410_samshort_fdl.pdf](https://github.com/WCSim/WCSim/files/323655/neutroncapturesgd_geant410_samshort_fdl.pdf)

Some validation plots with the changes from 4.9.4 to 4.10.1 can be read here:
[wcsim_terri_docdump.pdf](https://github.com/WCSim/WCSim/files/323658/wcsim_terri_docdump.pdf)
(Note: 4.10.2 also has a change to the Cherenkov time step that GEANT uses, which should affect the number of hits and possibly the timing.  Don't know exactly how just yet.)

### DIFF
--- a/src/WCSimPhysicsList.cc
+++ b/src/WCSimPhysicsList.cc
@@ -84,8 +84,9 @@ void WCSimPhysicsList::ConstructProcess()
 #include "G4MuBremsstrahlung.hh"
 #include "G4MuPairProduction.hh"
 #include "G4hIonisation.hh"
-#include "G4MuonMinusCaptureAtRest.hh"
 
+// FDL replaced in 4.9 by G4MuonMinusCapture
+#include "G4MuonMinusCapture.hh"
 //K.Z.: New MultipleScattering classes 
 #include "G4eMultipleScattering.hh"
 #include "G4MuMultipleScattering.hh"
@@ -99,16 +100,16 @@ void WCSimPhysicsList::ConstructEM()
   //replaced by new G4MultipleScattering classes for e+-, mu+-, hadron and ions.
   //K. Zbiri, 12/30/2009
 
-  theParticleIterator->reset();
-  while( (*theParticleIterator)() ){
-    G4ParticleDefinition* particle = theParticleIterator->value();
+  aParticleIterator->reset();
+  while( (*aParticleIterator)() ){
+    G4ParticleDefinition* particle = aParticleIterator->value();
     G4ProcessManager* pmanager = particle->GetProcessManager();
     G4String particleName = particle->GetParticleName();
      
     if (particleName == "gamma") {
     // gamma
       pmanager->AddDiscreteProcess(new G4GammaConversion());
-      pmanager->AddDiscreteProcess(new G4ComptonScattering());      
+      pmanager->AddDiscreteProcess(new G4ComptonScattering());
       pmanager->AddDiscreteProcess(new G4PhotoElectricEffect());
 
     } else if (particleName == "e-") {
@@ -122,7 +123,7 @@ void WCSimPhysicsList::ConstructEM()
       pmanager->AddProcess(theeminusMultipleScattering);
       pmanager->AddProcess(theeminusIonisation);
       pmanager->AddProcess(theeminusBremsstrahlung);
-      //      
+      //
       // set ordering for AlongStepDoIt
       pmanager->SetProcessOrdering(theeminusMultipleScattering,idxAlongStep,1);
       pmanager->SetProcessOrdering(theeminusIonisation,        idxAlongStep,2);
@@ -187,17 +188,17 @@ void WCSimPhysicsList::ConstructEM()
       // MF , stolen from CWW, april 2005
       if (particleName == "mu-")
         {
-          G4VProcess* aG4MuonMinusCaptureAtRest =
-            new G4MuonMinusCaptureAtRest();
-          pmanager->AddProcess(aG4MuonMinusCaptureAtRest);
-          pmanager->SetProcessOrdering(aG4MuonMinusCaptureAtRest,idxAtRest);
+	  // Change to G4MuonMinusCapture - FDL
+	  G4VProcess* aG4MuonMinusCapture = new G4MuonMinusCapture();
+          pmanager->AddProcess(aG4MuonMinusCapture);
+          pmanager->SetProcessOrdering(aG4MuonMinusCapture, idxAtRest);
         }
 
 
     } else if ((!particle->IsShortLived()) &&
 	       (particle->GetPDGCharge() != 0.0)&&
                (particle->GetParticleName() != "chargedgeantino")) {
-      // G4VProcess* aMultipleScattering = new G4MultipleScattering();
+
      G4VProcess* aMultipleScattering = new G4hMultipleScattering();
      G4VProcess* anIonisation        = new G4hIonisation();
      //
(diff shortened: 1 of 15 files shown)


src/WCSimPhysicsList.cc
```diff
@@ -223,16 +224,16 @@ void WCSimPhysicsList::ConstructEM()
 void WCSimPhysicsList::ConstructlArStepLimiter(){
 
 #ifdef GEANT4_7_0
-  theParticleIterator->reset();
-  while( (*theParticleIterator)() ){
+  aParticleIterator->reset();
```
eosullivan: Why has this changed? This seems to be for a set version of Geant.

src/WCSimPhysicsList.cc
```diff
@@ -392,349 +369,341 @@ void WCSimPhysicsList::ConstructGeneral()
...
-	    G4LENeutronInelastic* theInelasticModel = new G4LENeutronInelastic;
-	    theInelasticModel->SetMinEnergy(19*MeV);
-	    theInelasticProcess->RegisterMe(theInelasticModel);
-	  }
-	  else if (bertinihad) {
+	  if (bertinihad) {
 	    G4CascadeInterface* theBertiniModel = new G4CascadeInterface;
-	    theBertiniModel->SetMinEnergy(19*MeV);
+	    theBertiniModel->SetMinEnergy(19*CLHEP::MeV);
+	    theBertiniModel->SetMaxEnergy(5.0*CLHEP::GeV);
```
eosullivan: Is it necessary now in Geant 4.10 to set a maximum?

src/WCSimPhysicsList.cc
```diff
@@ -392,349 +369,341 @@ void WCSimPhysicsList::ConstructGeneral()
...
-// Sets secondary hadronic interaction model 
-// Note: this is currently only implemented for
-// protons and neutrons -- not pions.
-// Gheisha = Original Geant4 default 
-// Bertini = Bertini intra-nuclear cascade model
-// Binary  = Binary intra-nuclear cascade model
 void WCSimPhysicsList::SetSecondaryHad(G4String hadval)
 {
+
   SecondaryHadModel = hadval;
```
eosullivan: The Gheisha model is no longer available in Geant 4.10?

eosullivan:
Great, thanks for putting this pull request in. It looks like you have done extensive verification. Just so I can understand what you were testing, were you comparing your code with Geant 4.9.4 vs. 4.10, or were you comparing the old code with Geant 4.9.4 vs. your code with Geant 4.10? Both comparisons are interesting since the first case tells us what changes we would expect from just changing Geant versions while the second case would tell us if there were any differences in the code that will change our output. On a related note, if I run the code in this pull request with Geant 4.9.4, will it match identically with the current development branch run with Geant 4.9.4?

src/WCSimPhysicsList.cc
snoteptone: From http://geant4.cern.ch/support/ReleaseNotes4.9.6.html 
"Physics lists which depend on the CHIPS package or on the parameterized (Gheisha-like) models (LEP/HEP) are deprecated and will be removed in Geant4 10."

src/WCSimPhysicsList.cc
snoteptone: I don't think this matters too much, but I don't have a copy of 4.7.0 to test against.  I think if the pointer is declared to be aParticleIterator rather than theParticleIterator in the code, then it shouldn't matter.  If not, this can be reverted since, as you already mentioned, it's for a specific (and old!) version of GEANT4.

src/WCSimPhysicsList.cc
snoteptone: No, it is not a requirement at this juncture.  The idea is to have a better practice to understand exactly which model is valid where, especially if things change as GEANT makes new releases.  We're not quite as beholden if they change a threshold in that case (and it may help validations).

src/WCSimPhysicsMessenger.cc
```diff
@@ -14,14 +14,14 @@ WCSimPhysicsMessenger::WCSimPhysicsMessenger(WCSimPhysicsList* WCSimPhys)
   WCSimDir->SetGuidance("Commands to change secondary interaction model for protons");
 
   hadmodelCmd = new G4UIcmdWithAString("/WCSim/physics/secondaries/model",this);
-  hadmodelCmd->SetGuidance("Available options: GHEISHA BERTINI BINARY");
+  hadmodelCmd->SetGuidance("Available options for low energy (< 10 GeV): BERTINI BINARY (default is BERTINI)");
```
snoteptone: This was a suggestion from the G4developers, but we need to check if
the Bertini is better than Binary or vice-versa. The model isn't automatically
used it does need to be defined.

snoteptone:
Hi Erin,
To quote you:
"Great, thanks for putting this pull request in. It looks like you have done extensive verification. Just so I can understand what you were testing, were you comparing your code with Geant 4.9.4 vs. 4.10, or were you comparing the old code with Geant 4.9.4 vs. your code with Geant 4.10? Both comparisons are interesting since the first case tells us what changes we would expect from just changing Geant versions while the second case would tell us if there were any differences in the code that will change our output. On a related note, if I run the code in this pull request with Geant 4.9.4, will it match identically with the current development branch run with Geant 4.9.4?"

I was comparing the old code w/ 4.9.4 and new code w/ 4.10.1, though as you can see in the docdump file, there's not much a difference in that respect.

For the question at the end, I haven't tried that, but I don't think it will due to the model changes and changes in header files by GEANT in the process of going from 4.9 to 4.10, so (assuming it compiles) it won't run.  Hope that clears things up a bit.

src/WCSimPhysicsList.cc
```diff
@@ -392,349 +369,341 @@ void WCSimPhysicsList::ConstructGeneral()
...
+
+	  // Added Bertini Model - FDL
+	  G4CascadeInterface* theBertiniModel = new G4CascadeInterface;
+	  theBertiniModel->SetMaxEnergy(5.0*CLHEP::GeV);
+	  theInelasticProcess->RegisterMe(theBertiniModel);
+
+	  // Added FTFP Model - SS
+	  FTFP_model->SetMinEnergy(4.0*CLHEP::GeV);
+	  theInelasticProcess->RegisterMe(FTFP_model);
 	  pmanager->AddDiscreteProcess(theInelasticProcess);
```
eosullivan: It looks like there is an overlap in energy ranges for the Bertini and FTFP models in the pi+ particles between 4 and 5 GeV. @fdilodov, I see you put in the Bertini part. Do you know anything about this? If the overlap is a problem, which model should be adjusted?

src/WCSimPhysicsList.cc
```diff
@@ -392,349 +369,341 @@ void WCSimPhysicsList::ConstructGeneral()
...
 	  }
 	  else {
-	    G4cout << "No secondary interaction model chosen! Using G4 BINARY." << G4endl;
-	    G4BinaryCascade* theBinaryModel = new G4BinaryCascade();
-	    theInelasticProcess->RegisterMe(theBinaryModel);
+	    G4cout << "No secondary interaction model chosen! Using G4 BERTINI." << G4endl;
+	    G4CascadeInterface* theBertiniModel = new G4CascadeInterface;
+	    theBertiniModel->SetMaxEnergy(5.0*CLHEP::GeV);
+	    theInelasticProcess->RegisterMe(theBertiniModel);
 	  }
```
eosullivan: This code changes the default for the secondary interactions to bertini and FTFP (rather than binary), but it hasn't been changed in jobOptions2.mac. Is the proposal to change the secondary interactions model? If so, we should change it in the macro. Again, @fdilodov, you did some work on this. Do you know if switching the secondary model is the intention here?

eosullivan:
resolved with PR #211

Issue #99, opened 2015-07-27
Add comments in code to generate meaningful doxygen output


eosullivan:
This was brought up by @AlexFinch in pull request #89. Our makefile runs doxygen, but we haven't put doxygen style comments in the code. Ensuring each class and method had a basic description of what it does and what the arguments are would allow us to generate meaningful doxygen documentation.

tdealtry:
I've started to added some doxygen to #89, but there is a bit of a problem with the output - the class methods are listed twice (once seemingly from the .hh file, and the other from the .cc file). There's a couple of obvious possibilities for what's gone wrong:
* I've used the doxygen syntax incorrectly. You can check this at 43b2ac6cf1cda9ae5f7ec3156b12b79728e48f89
* There's an issue with the WCSim_doxygen_config

Perhaps a doxygen expert has some ideas?

PR #180, opened 2016-06-22
extend .gitignore file

JostMigenda:
Run results include `wcsim.root` (from `WCSim.mac`), `wcsimtest.root` (from `verifications-test-scripts/electrontest.mac`) and potentially outputs from a user’s own macros (with arbitrary file names), so let’s just ignore all `.root` files.

Also, once we start using doxygen (see #99), that’s another build results folder to ignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,8 @@ make.rootcint.result
 make.result
 bin/
 tmp/
+doxygen/
 
 #ignore run results
-wcsim.root
+*.root
 geofile.txt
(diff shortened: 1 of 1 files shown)


tdealtry:
Thanks for this - a useful addition. Perhaps while you're doing this, perhaps you want to add a few more things to the ignore list? The following are useful when running macros that ROOT or python compiles.
```
#ignore compiled ROOT scripts
*_C.d
*_C.so
*_cxx.d
*_cxx.so
*_cc.d
*_cc.so

#ignore compiled python scripts
*.pyc
```

JostMigenda:
Added all of these; thanks for the suggestion!

PR #188, opened 2016-08-29
Fix integer digit time bug

cristovaovilela:
Digitized hit times are being floored to `int` in the trigger code, effectively being "rounded" to the previous nanosecond. This bug is present in the latest three tagged versions of WCSim `v1.5.0`, `v1.5.1` and `v1.6.0`.

Below are distributions of digitized hit times `WCSimRootCherenkovDigiHit::GetT()` for 600 MeV electron particle gun MC generated with WCSim `v1.4.0` (before the bug was introduced), the current state of the `develop` branch, and the code in this pull request.

digitversioncomparefull

digitversioncomparedetail

### DIFF
--- a/src/WCSimWCTrigger.cc
+++ b/src/WCSimWCTrigger.cc
@@ -197,7 +197,7 @@ void WCSimWCTriggerBase::AlgNDigits(WCSimWCDigitsCollection* WCDCPMT, bool remov
   int window_end_time   = WCSimWCTriggerBase::LongTime - ndigitsWindow;
   int window_step_size  = 5; //step the search window along this amount if no trigger is found
   float lasthit;
-  std::vector<int> digit_times;
+  std::vector<G4float> digit_times;
   bool first_loop = true;
 
   G4cout << "WCSimWCTriggerBase::AlgNDigits. Number of entries in input digit collection: " << WCDCPMT->entries() << G4endl;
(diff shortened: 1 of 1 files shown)


eosullivan:
Ok, looks like this came in with PR #89. @tdealtry and @stewartt1982, I also see that the timing windows for the trigger are integers (NDigitsPreTriggerWindow, for example) where they used to be floats. Is there functional reason to do this or was this unintentional?

tdealtry:
I think this was a confusion between how digit time & trigger time were handled with the old code - trigger times are rounded (floored to 5 ns). This commit does the correct thing to line it up with what the old code did. But of course, in reality, electronics don't give digit timing to floating point precision, but are limited by clock ticks (e.g. to 2.5 ns). Whether we want to handle this correctly with some appropriate flooring, I'm unsure.
In terms of the user inputs (NDigitsPreTriggerWindow etc.) these could be made floats of course. (The values they were set to were all floating point representations of integers, so I guess that's why I made them int in #89)

eosullivan:
Ok, I see. @tdealtry, do you know what the appropriate precision is in the actual electronics? Is it 5 ns? Or is it 2.5 ns? I am not opposed to moving away from floats in favour of making things more realistic, but then we should aim to round things in the most realistic way.

tdealtry:
I think moving to floats is the appropriate thing to do, but also add a digitizer-specific rounding. According to Section 5.1 of the SK NIM paper, the timing resolution of a hit is 0.4 ns.

eosullivan:
Great, allowing for digitizer-specific rounding is a better idea. Do you have time to implement this @tdealtry? If so, I'll close this pull request. Do you agree @cvilelasbu?

cvilelasbu:
Fine by me.

On September 12, 2016 3:15:52 PM GMT+01:00, eosullivan <notifications@github.com> wrote:
>Great, allowing for digitizer-specific rounding is a better idea. Do
>you have time to implement this @tdealtry? If so, I'll close this pull
>request. Do you agree @cvilelasbu? 
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/WCSim/WCSim/pull/188#issuecomment-246360959

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

tdealtry:
Yeah, I'll be able to get to this within a couple of weeks

eosullivan:
Great, I'll close this out then. Just for an idea of the timeline, I'm holding off on merging anything until  #178 is merged (because I want people to test their PRs against the new Geant version), but once that's in I hope to get back to merging the outstanding PRs here.

Issue #189, opened 2016-10-12
Stopped muon decay in GEANT 4.9.4


mscott201:
Looking at muon neutrino interactions in WCSim as part of the NuPRISM fiTQun validation I noticed that there were a number of events with clear Cherenkov rings that were not associated to any true particle.

Looking in WCSimTrackingAction.cc I found that the true particle trajectories were only saved for "Decay" processes, but that Michel electrons were being produced by both the G4Decay process and the G4MuonMinusCaptureAtRest process.  I've copied some output from the tracking action code below, listing the particle type and its creation process:

track # 2,it is a mu-
track # 49702, it is a e-
    It is from a Decay
track # 49701, it is a anti_nu_e
    It is from a Decay
track # 49700, it is a nu_mu
    It is from a Decay
track # 49699, it is a nu_mu
    It is from a muMinusCaptureAtRest
track # 49698, it is a anti_nu_e
    It is from a muMinusCaptureAtRest
track # 49697, it is a e-
    It is from a muMinusCaptureAtRest

This shows a single muon producing two decay electrons.  Attached is a screenshot of this event, with the muon shown as the solid green line, the electrons as solid blue lines and then the associated neutrinos as the dashed lines.  I was able to display this extra electron by adding `ProcessList.insert("muMinusCaptureAtRest")` to the initialisation of WCSimTrackingAction.cc.
![mu_decay](https://cloud.githubusercontent.com/assets/9436369/19293783/0a9c5c08-8fdc-11e6-83f7-e7007e4e8bd7.png)
Investigating this with Peter Gumplinger, the problem is solved by using GEANT 4.9.6 (or higher).  There were large changes to the muon capture code for v4.9.6 going to the newer version which seems to have fixed this issue.  The second screenshot shows the same event with no changes to WCSim, and the track printout showed electrons coming from the muMinusCaptureAtRest process only (as expected).
![mu_decay_good](https://cloud.githubusercontent.com/assets/9436369/19293878/efb4327a-8fdc-11e6-8ea8-c43c2d71e4b1.png)

If anyone is using decay electrons in their physics studies then they should probably avoid GEANT 4.9.4 I think.  This might also give more impetus for switching the official WCSim GEANT4 version to something more recent.

eosullivan:
Thanks for pointing this out! We are imminently moving to Geant4.10 with PR #178, but this is interesting for people who have been using the current version of the code to note.

mscott201:
[...]

PR #191, opened 2016-10-18
Prevent digits from being saved multiple times in different trigger windows

### LINKED ISSUE #190: Overlapping triggers with duplicate digits
In releases v1.5.0 and later, trigger windows are allowed to overlap and digits in the overlapping interval are saved twice to the output file (once in each trigger).

For example, the following histogram shows the hit times for a muon particle gun event, where hit distributions belonging to different triggers are colour coded and stacked.
overlappingwcsimtriggers

In the "old" trigger code (v1.4.0) this didn't happen, as there was a check for previous triggers in the `DigitzeGate` method:

``` c++
G4double lowerbound;
if (G==0) lowerbound = TriggerTimes[G]+EvtG8Down;
  else {
    lowerbound = ( TriggerTimes[G]+EvtG8Down >= TriggerTimes[G-1] + EvtG8Up ) ?  
TriggerTimes[G]+EvtG8Down : TriggerTimes[G-1] + EvtG8Up;
  }
G4double upperbound = TriggerTimes[G]+EvtG8Up;
```

[See here in v1.4.0](https://github.com/WCSim/WCSim/blob/v1.4.0/src/WCSimWCDigitizer.cc#L353:L358)

As far as I can tell, this adjustment for the previous trigger window end time does not exist in the new trigger trigger code's `FillDigitsCollection` method.

``` c++
float lowerbound = triggertime + GetPreTriggerWindow(triggertype);
float upperbound = triggertime + GetPostTriggerWindow(triggertype);
```

[Here in v1.6.0](https://github.com/WCSim/WCSim/blob/v1.6.0/src/WCSimWCTrigger.cc#L319:L320)

Occasionally, `lowerbound` of trigger `N` will be earlier than `upperbound` of trigger `N-1`, so digits between `lowerbound` of trigger `N` and `upperbound` of trigger `N-1` will be saved twice. Is this behaviour intended?

tdealtry:
Fixes #190

@cvilelasbu, could you make a tricolour plot to confirm it's working as expected

Standard verification. KS tests are 1

electrontest_trigger_overlaps

### DIFF
--- a/include/WCSimWCTrigger.hh
+++ b/include/WCSimWCTrigger.hh
@@ -169,6 +169,27 @@ private:
 
   ///takes all trigger times, then loops over all Digits & fills the output DigitsCollection
   void FillDigitsCollection(WCSimWCDigitsCollection* WCDCPMT, bool remove_hits, TriggerType_t save_triggerType);
+
+  ///sort the Trigger vectors (Time, Type, Info) by Trigger Time
+  void SortTriggersByTime() {
+    int i, j;
+    TriggerType_t index_type;
+    float index_time;
+    std::vector<float> index_info;
+    for (i = 1; i < (int) TriggerTimes.size(); ++i) {
+      index_time = TriggerTimes[i];
+      index_type = TriggerTypes[i];
+      index_info = TriggerInfos[i];
+      for (j = i; j > 0 && TriggerTimes[j-1] > index_time; j--) {
+	TriggerTimes[j] = TriggerTimes[j-1];
+	TriggerTypes[j] = TriggerTypes[j-1];
+	TriggerInfos[j] = TriggerInfos[j-1];
+      }//j
+      TriggerTimes[j] = index_time;
+      TriggerTypes[j] = index_type;
+      TriggerInfos[j] = index_info;
+    }//i
+  }
   
   static const double offset;        ///< Hit time offset (ns)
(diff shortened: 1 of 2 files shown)


eosullivan:
@cvilelasbu, I didn't realize that this was holding up your fiTQun activities. Let's try and get this solved ASAP. Have you made your tri-colour plot to confirm that this PR fixes issue #190?

cvilelasbu:
Hi @eosullivan ,

This PR seems to fix the overlapping triggers issue. See, for example, this event:
cyl40percentmu_pr191_4_tseries
Sorry I didn't get around to doing this check sooner.

However, it's not this issue that is the problem for fiTQun tuning. The problem is issue #189. The current WCSim physics model produces 2 Michels for each muon, so it will need to be updated. I've been working on re-tuning the fiTQun description of the Cherenkov emission profile, where we think there might be some subtle differences compared to Geant3 that might be producing the vertex bias in muon reconstruction. Since this process is quite resource consuming, I would prefer to try it again once there is a definite fix for the two-Michel issue, rather than hacking WCSim myself and then having to re-tune if a different solution is chosen. A few more details (and plots) about issue #189 [here](http://www-sk.icrr.u-tokyo.ac.jp/indico/getFile.py/access?sessionId=0&resId=0&materialId=0&confId=2908).

Cheers,
Cris

PR #192, opened 2016-10-18
Change from int to double for the input timing parameters to daq.mac

tdealtry:
This change is propagated through to use in WCSimWCDigitizer and WCSimWCTrigger
An extension of #188 

It'd be worthwhile for someone with a good eye to go through all the files & check if I've missed any `int -> double` conversions

Standard verification. KS tests are 1

electrontest_daq_option_precision


eosullivan:
It's kind of hard to see with these plots. Does this PR change the timing behaviour back to the desired shape as shown in https://github.com/WCSim/WCSim/pull/188?

tdealtry:
Closing this, since it's unlikely that these parameters (digitiser deadtime/integration window & trigger decision/readout windows) will ever need more than ns precision. They can of course be changed at a later date if required

Issue #132, opened 2015-11-16
Add a geometry identifier to WCSimRootGeom


tdealtry:
It'll be useful for an identifier of which geometry was used in WCSimRootGeom, so that in output files this information is easily accessible. I see a couple of obvious ways to do this:
- Add a string data member (e.g. the same strings used for /WCSim/WCgeom)
- Add an enumeration of detectors

Let me know what method you prefer and I can add it in. Personally, I prefer the first method, as it's easier to specify different versions of the same geometry (if this is something that happens?), etc.
This is required for hyperk/hk-BONSAI#2

eosullivan:
Adding a string sounds reasonable. Do you have any preference @cvilelasbu? We had talked about adding something like this in to properly interface with fiTQun.

tdealtry:
I've just had a thought - having all the run options (geometry, noise, DAQ, particle gun / input vector file, ...) stored in the output file would be useful. Perhaps we could add a new tree with all this information in?

eosullivan:
Sorry, just rediscovering this now. Does this increase the file size much? Or the time to run the program?

eosullivan:
Also, can I assign you this issue?

tdealtry:
Also just rediscovering this. It'll be a tree with 1 entry, with 20-30 leaves. So it'll be negligible in terms of file size. Running should be quick too - it's just a matter of calling 20-30 `Get*()` methods in `EndOfRunAction()`. Go ahead & assign me, if you think this is a sensible solution.

PR #195, opened 2016-10-21
Add detector name to the ROOT output

tdealtry:
Partial fix of #132 (was doing it for something else, thought it'd be useful to pull request for e.g. hyperk/hk-BONSAI#2). I've added it to WCSimRootGeom, because that seems appropriate.

Standard validation: ks tests are 1

electrontest_detectorname_in_output


eosullivan:
@tdealtry I think this is now solved with PR #198, right?

tdealtry:
Yeah, it's resolved via #198

PR #198, opened 2016-12-21
Run options in output

tdealtry:
Solves #132 
I ran the standard verification; no changes are observed.
Everything from the WCSim messenger classes are saved (I think!), but other things (e.g. particle gun options) controlled by standard G4 classes are not. Requests for extra information to save are welcomed!
I put a new tree (with one entry) with one branch (the new class) - this branch could be put in `wcsimGeoT` (another tree with one entry & one branch) instead, but should probably be renamed, which is why I've gone down the new tree route to start with.
#195 can either be cancelled or kept (the detector name is in `WCSimRootOptions` here, and it can also be in `WCSimRootGeom`, as in #195, without conflict)
I'll put a pull request in at [WCSim-Documentation](https://github.com/WCSim/WCSim-Documentation) with information about how to add new options to the output when you've approved this format.

### DIFF
--- a/include/WCSimEnumerations.hh
+++ b/include/WCSimEnumerations.hh
@@ -15,12 +15,19 @@ typedef enum EDigitizerType {
   kDigitizerSKI
 } DigitizerType_t;
 
+typedef enum ERandomGeneratorType {
+  RANDOM_E_RANLUX=1,
+  RANDOM_E_RANECU=2,
+  RANDOM_E_HEPJAMES=3
+} WCSimRandomGenerator_t;
+
 class WCSimEnumerations
 {
 public:
 
   static std::string EnumAsString(DigitizerType_t d);
   static std::string EnumAsString(TriggerType_t t);
+  static std::string EnumAsString(WCSimRandomGenerator_t r);
   static TriggerType_t TriggerTypeFromString(std::string s);
 
 };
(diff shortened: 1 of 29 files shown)


eosullivan:
Let's leave the geometry information and electronics information in their own trees for now, like you have it in this PR. I can imagine adding new branches to both these trees with the OD info, once that is implemented.

Also, can you update this now that I've merged in a couple of other PRs? Some conflict is preventing me from compiling this branch.

include/WCSimEnumerations.hh
```diff
@@ -15,12 +15,19 @@ typedef enum EDigitizerType {
   kDigitizerSKI
 } DigitizerType_t;
 
+typedef enum ERandomGeneratorType {
+  RANDOM_E_RANLUX=1,
+  RANDOM_E_RANECU=2,
+  RANDOM_E_HEPJAMES=3
+} WCSimRandomGenerator_t;
+
```
tdealtry: This was done to keep `libWCSimRoot.so` compilable without a Geant4 installation; `libWCSimRoot.so` depends only on the files `WCSimRoot*.*`, `WCSimPmtInfo `WCSimEnumerations`

tdealtry:
I've fixed the merge issue.

In terms of file sizes, here are the numbers, using the standard verification mac file. Very small change compared to everything else!
The TTree "wcsimGeoT" takes 56855 bytes on disk
  Its branch "wcsimrootgeom" takes 56716 bytes on disk
The TTree "wcsimGeoT" takes 56855 bytes on disk
  Its branch "wcsimrootgeom" takes 56716 bytes on disk
Trees sizes are identical!
The TTree "wcsimT" takes 95112663 bytes on disk
  Its branch "wcsimrootevent" takes 95112366 bytes on disk
The TTree "wcsimT" takes 95112408 bytes on disk
  Its branch "wcsimrootevent" takes 95112218 bytes on disk
Your WCSim version tree is 0.000268104% larger than the master GitHub version
The TTree "wcsimRootOptionsT" takes 711 bytes on disk
  Its branch "wcsimrootoptions" takes 569 bytes on disk

I'll put in a PR with the code for this size check, and also run time checking

tdealtry:
Timings from BeginOfRunAction() to EndOfRunAction():
This code (with #199 added for time checking): 11430.5 seconds (CPU)   11461.4 seconds (real)
WCSim GitHub master (with #199 added for time checking): 11451.2 seconds (CPU)   11481.8 seconds (real)
So the PR version is actually a (very slightly) faster... likely a fluctuation; verification is still saying outputs are identical.

PR #199, opened 2017-01-12
Run time & file size checking

tdealtry:
For #198, I was asked to check changes to output file sizes and running time. Here's a pull request, so they can be reused in future:

- run time is the time from BeginOfRunAction() to EndOfRunAction(), and can be used/not with a .mac command. Default is off
- file size is a macro to be run on output files, and takes the same arguments as `verification_HitsChargeTime.C`

### DIFF
--- a/WCSim.mac
+++ b/WCSim.mac
@@ -149,5 +149,8 @@
 
 /WCSimIO/RootFile wcsim.root
 
+## set a timer running on WCSimRunAction
+#/WCSimIO/Timer false
+
 /run/beamOn 10
 #exit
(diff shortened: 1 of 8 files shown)


eosullivan:
This is certainly nice to have! @brichards64, we had talked some point about including a check for run time and output file size as part of the larger verification. Is this something that can be done directly in Travis CI, or would we need to incorporate something extra into the code like in this pull request?

brichards64:
I afraid i ahve been massively waylayed with DAQ work load but it is something we could easilly do. Travis itself is somewhat limited its for building only. But in the original verison i wrote had all sorts of varifications and checks and a file size would be an easy addition once i get everyhting else running

eosullivan:
There is a README file in the verification-test-scripts directory. Can you add a description of your new scripts?

PR #201, opened 2017-02-01
WCSim migration to G4.10 ### TO BE DELETED ### See pull request #202

P3tru:
This is the implementation of G4.10 inside WCSim.

Tests have been run to compare if there is a difference between version of Geant4 9.4.4 and Geant4 10.1.3 and results shows no differences for electrons and muons between [100-1000] MeV by setting the BERTINI physics list by default.
However, changes can be seen in muons at low energy (100 MeV) using the BINARY physics list. As a recommendation by Geant4 developers, the BERTINI physics list is set to the default one in file joboption2.mac.
Some results can be found at http://pprc.qmul.ac.uk/~zsoldos/G4results.pdf
It's worth noticing that the GEISHA models have been depreciated and are now removed from WCSim.

The biggest changes can be found in the WCSimPhysicsList class. Other changes are mainly made to incorporate the new headers for physical constants introduced by Geant4 version 10.

This work is based on Sam Short and Francesca Di Lodovico implementation. See https://github.com/WCSim/WCSim/pull/178

PR #202, opened 2017-02-08
Add G4.10.x compatibility

P3tru:
This is the implementation of G4.10 inside WCSim.

Tests have been run to compare if there is a difference between version of Geant4 9.4.4 and Geant4 10.1.3 and results shows no differences for electrons and muons between [100-1000] MeV.
As a recommendation by Geant4 developers, the BERTINI physics list is set to the default one in file joboption2.mac.

It's worth noticing that the GEISHA models have been depreciated and are now removed from WCSim.

The biggest changes can be found in the WCSimPhysicsList class. Other changes are mainly made to incorporate the new headers for physical constants introduced by Geant4 version 10.

This work is based on Sam Short and Francesca Di Lodovico implementation. See #178 

EDIT : Validation plots can be find here from electrons, muons and pions :
[G4FinalResults_Update.pdf](https://github.com/WCSim/WCSim/files/770389/G4FinalResults_Update.pdf)
[G4Validation_yoshida_v2.pdf](https://github.com/WCSim/WCSim/files/770390/G4Validation_yoshida_v2.pdf)

### DIFF
--- a/src/WCSimPhysicsList.cc
+++ b/src/WCSimPhysicsList.cc
@@ -66,7 +66,7 @@ void WCSimPhysicsList::ConstructProcess()
 {
   //AddTransportation();
   ConstructEM();
-  ConstructlArStepLimiter();
+  // ConstructlArStepLimiter();
   ConstructGeneral();
   ConstructOp();
   ConstructHad();
@@ -84,7 +84,8 @@ void WCSimPhysicsList::ConstructProcess()
 #include "G4MuBremsstrahlung.hh"
 #include "G4MuPairProduction.hh"
 #include "G4hIonisation.hh"
-#include "G4MuonMinusCaptureAtRest.hh"
+// FDL replaced in 4.9 by G4MuonMinusCapture
+#include "G4MuonMinusCapture.hh"
 
 //K.Z.: New MultipleScattering classes 
 #include "G4eMultipleScattering.hh"
@@ -99,9 +100,9 @@ void WCSimPhysicsList::ConstructEM()
   //replaced by new G4MultipleScattering classes for e+-, mu+-, hadron and ions.
   //K. Zbiri, 12/30/2009
 
-  theParticleIterator->reset();
-  while( (*theParticleIterator)() ){
-    G4ParticleDefinition* particle = theParticleIterator->value();
+  aParticleIterator->reset();
+  while( (*aParticleIterator)() ){
+    G4ParticleDefinition* particle = aParticleIterator->value();
     G4ProcessManager* pmanager = particle->GetProcessManager();
     G4String particleName = particle->GetParticleName();
      
@@ -163,8 +164,8 @@ void WCSimPhysicsList::ConstructEM()
                particleName == "mu-"    ) {
     //muon
       //G4VProcess* aMultipleScattering = new G4MultipleScattering();
-       G4VProcess* aMultipleScattering = new G4MuMultipleScattering();
-       G4VProcess* aBremsstrahlung     = new G4MuBremsstrahlung();
+			G4VProcess* aMultipleScattering = new G4MuMultipleScattering();
+			G4VProcess* aBremsstrahlung     = new G4MuBremsstrahlung();
       G4VProcess* aPairProduction     = new G4MuPairProduction();
       G4VProcess* anIonisation        = new G4MuIonisation();
       //
@@ -187,10 +188,10 @@ void WCSimPhysicsList::ConstructEM()
       // MF , stolen from CWW, april 2005
       if (particleName == "mu-")
         {
-          G4VProcess* aG4MuonMinusCaptureAtRest =
-            new G4MuonMinusCaptureAtRest();
-          pmanager->AddProcess(aG4MuonMinusCaptureAtRest);
-          pmanager->SetProcessOrdering(aG4MuonMinusCaptureAtRest,idxAtRest);
+          G4VProcess* aG4MuonMinusCapture =
+            new G4MuonMinusCapture();
+          pmanager->AddProcess(aG4MuonMinusCapture);
+          pmanager->SetProcessOrdering(aG4MuonMinusCapture,idxAtRest);
         }
 
 
@@ -223,10 +224,10 @@ void WCSimPhysicsList::ConstructEM()
 void WCSimPhysicsList::ConstructlArStepLimiter(){
 
 #ifdef GEANT4_7_0
-  theParticleIterator->reset();
-  while( (*theParticleIterator)() ){
+  aParticleIterator->reset();
+  while( (*aParticleIterator)() ){
 
-    G4ParticleDefinition* particle = theParticleIterator->value();
+    G4ParticleDefinition* particle = aParticleIterator->value();
     G4ProcessManager* pmanager = particle->GetProcessManager();
 
      if ((!particle->IsShortLived()) &&
@@ -280,10 +281,10 @@ void WCSimPhysicsList::ConstructOp(){
   // theCherenkovProcess->SetMaxBetaChangePerStep(10.0);
   // theCherenkovProcess->SetTrackSecondariesFirst(true);
 
-  theParticleIterator->reset();
-  while( (*theParticleIterator)() )
+  aParticleIterator->reset();
+  while( (*aParticleIterator)() )
   {
-    G4ParticleDefinition* particle     = theParticleIterator->value();
+    G4ParticleDefinition* particle     = aParticleIterator->value();
     G4ProcessManager*     pmanager     = particle->GetProcessManager();
     G4String              particleName = particle->GetParticleName();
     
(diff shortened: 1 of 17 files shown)


src/WCSimPhysicsList.cc
```diff
@@ -163,8 +164,8 @@ void WCSimPhysicsList::ConstructEM()
                particleName == "mu-"    ) {
     //muon
       //G4VProcess* aMultipleScattering = new G4MultipleScattering();
-       G4VProcess* aMultipleScattering = new G4MuMultipleScattering();
-       G4VProcess* aBremsstrahlung     = new G4MuBremsstrahlung();
+			G4VProcess* aMultipleScattering = new G4MuMultipleScattering();
+			G4VProcess* aBremsstrahlung     = new G4MuBremsstrahlung();
```
eosullivan: Revert the spacing of these two lines back to how they were.

P3tru:
I have corrected the indentation problem of WCSimPhysicsList class. I work with the CLion IDE and I don't have any indentation problem on my local version, don't know why he got mixed up when I commited my work. Hope the code will be more readable now.
I removed the geisha line from joboptions2.mac.
Let me know when you are happy with the reviewing on the code and I will run the same tests with electrons and muons at 100, 500 and 1000 MeV and check that we still see the same things.

src/WCSimPhysicsList.cc
```diff
@@ -66,7 +66,7 @@ void WCSimPhysicsList::ConstructProcess()
 {
   //AddTransportation();
   ConstructEM();
-  ConstructlArStepLimiter();
+  // ConstructlArStepLimiter();
```
eosullivan: Why is this line commented out? Is it necessary for it to be removed for Geant4.10?

eosullivan:
You also need to change things in WCSimPhysicsMessenger.cc. The default is set to BINARY and the GEISHA option is still in there. It works fine since we explicitly set the model in jobOptions2.mac, but we want BERTINI to be the default if no other model is set.

src/WCSimPhysicsList.cc
```diff
@@ -522,66 +551,55 @@ void WCSimPhysicsList::ConstructHad()
 	  // Choice defined in jobOptions.mac, which is
 	  // read in before initialization of the run manager.
 	  // In the absence of this file, BINARY will be used.
```
eosullivan: Once you change WCSimPhysicsMessenger.cc, BERTINI will be the default, so this comment should be updated.

src/WCSimPhysicsList.cc
```diff
@@ -522,66 +551,55 @@ void WCSimPhysicsList::ConstructHad()
...
-	  if (gheishahad) {
-	    G4LEProtonInelastic* theLEInelasticModel = new G4LEProtonInelastic;
-	    theInelasticProcess->RegisterMe(theLEInelasticModel);
-	  }
-	  else if (bertinihad) {
-	    G4CascadeInterface* theBertiniModel = new G4CascadeInterface;
-	    theInelasticProcess->RegisterMe(theBertiniModel);
+	  if (bertinihad) {
+	   G4CascadeInterface* theBertiniModel = new G4CascadeInterface;
+	   theBertiniModel->SetMaxEnergy(5.0*GeV);
```
eosullivan: Do you have a reference for where these min and max energies are coming from? If so, please add the reference to the comments somewhere.

src/WCSimPhysicsList.cc
P3tru: That was an artefact when I was debugging the new particleIterator object. This method works fine and should be uncommented.

src/WCSimTrajectory.cc
```diff
@@ -93,7 +96,7 @@ void WCSimTrajectory::ShowTrajectory(std::ostream& os) const
 void WCSimTrajectory::DrawTrajectory(G4int i_mode) const
 {
   // Invoke the default implementation in G4VTrajectory...
-  G4VTrajectory::DrawTrajectory(i_mode);
+  G4VTrajectory::DrawTrajectory();
```
eosullivan: Is this change necessary for the Geant4.10 upgrade?

src/WCSimPhysicsList.cc
P3tru: To be honest, I am not sure. I use to make my tests by setting a non-overlapping range and at the end we also wanted to test to setup an overlapping range between the physics lists to see if there was no crashes. This values came from Sam and I am looking for references at the moment.

eosullivan: Ok, please check so we can be sure things are set to the right value. There are different values for different particles, so it does look to be deliberate.

eosullivan:
Besides the issue with the min/max energy values, everything looks ok to me, so once you have verified these values, go ahead and do your final checks. We should aim to merge this after the software premeeting on Monday so that no one is caught off-guard when the develop branch of the code stops working.

src/WCSimTrajectory.cc
```diff
@@ -93,7 +96,7 @@ void WCSimTrajectory::ShowTrajectory(std::ostream& os) const
 void WCSimTrajectory::DrawTrajectory(G4int i_mode) const
```
tdealtry: Can you `\* i_mode *\` here now, to suppress unused variable compiler warnings? Or maybe better yet, just drop the `G4int` argument to `WCSimTrajectory::DrawTrajectory` altogether?

src/WCSimPhysicsList.cc
P3tru: This values seems to be the one by default for the FTFP_BERT Physics List.
See http://geant4.cern.ch/support/proc_mod_catalog/physics_lists/hadronic/FTFP_BERT.html

P3tru:
Here you can find the validation plots comparing the orgin version of WCSim with the forked one. 
I attached the pptx as well if needed to recover the plots.
[G4FinalResults.pdf](https://github.com/WCSim/WCSim/files/769035/G4FinalResults.pdf)
[G4FinalResults.pptx](https://github.com/WCSim/WCSim/files/769036/G4FinalResults.pptx)

eosullivan:
So, there are many plots here, but the interesting ones are missing. For the general validation, please show 4.9.4 (binary) and 4.10 (Bertini) since this was be the change of this pull request. Have you compared your plots from #201 with these plots? Is it possible to show them overlaid so we can see there are no changes?

P3tru:
Apologize, that was a unfortunate oversight...
Here are the plots 
[G4FinalResults_Update.pdf](https://github.com/WCSim/WCSim/files/769881/G4FinalResults_Update.pdf)
[G4FinalResults_Update.pptx](https://github.com/WCSim/WCSim/files/769882/G4FinalResults_Update.pptx)

tdealtry:
I see that the plots are for Geant4.10.1.3. Is there any reason to not go to the [latest](http://geant4.web.cern.ch/geant4/support/download.shtml) Geant4 release - 4.10.3?
@federiconova has found that there is a [radioactivity gamma cascade bug](http://indico.ipmu.jp/indico/event/100/session/2/contribution/66/material/slides/0.pdf) in 4.10.1. I'm unsure whether there has been a patch in 4.10.1 to fix it, or whether it is fixed in 4.10.2, but it is fixed in 4.10.3

P3tru:
I heard that the tests was expected with G4.10.1.3 (as Yoshida-san did for pions) so I kept on with this version, but I don't recall seeing any difference by using G4.10.3.0. I can re-run some simulation if wanted, it was not clear to me to which version WCSim wants to migrate --- just that we want to migrate to G4.10

eosullivan:
I guess the main reason is that Geant4.10.3 was just released in December and we started this process before then. Being at the cutting-edge of having the very best Geant physics is a moving target. It's unfortunate that we've already found an issue that will affect our simulations. 

Moving to Geant4.10.3 might be more work than just installing a new version and running this PR with it. It looks like there are significant to the sensitive detector code, among other things.

tdealtry:
Focussing on getting a G4.10.1.3-compatible version out sounds like the priority, so let's do that. How about @federiconova checks for the bug when this is merged, and makes an issue if it is an issue?

tdealtry:
Do we want some pion plots posting too?

eosullivan:
Yes, let's stick with 4.10.1.3. I'm reluctant to have more delays on this upgrade, since it is holding up the fiTQun tuning. I don't love the idea of changing the recommended Geant version on a regular basis either since that also might cause issues for the fiTQun tuning (as well as just being a general pain for users and generating a lot of comparison work for developers), but it sounds like moving to Geant4.10 will solve the major problems for fiTQun and a small perturbation after that might not have a large effect. @federiconova, please try out the new release when we upgrade and make an issue outlining the problem so we can consider it for the future. 

And, yes, good idea on including some pion plots. @P3tru, please include some of Yoshida-san's plots after we finalize things at the premeeting today.

P3tru:
Do you mean to just upload her presentation ? Or run some pions simulation at 100 and 500 MeV as for the electrons and muons ?

eosullivan:
I think just uploading her presentation (or some select plots from it) is sufficient.

P3tru:
There are the final validation plot for pions, which agrees with Yoshida's plots.
[pi+300MeV_90404-100103.pdf](https://github.com/WCSim/WCSim/files/777340/pi.300MeV_90404-100103.pdf)

eosullivan:
resolved with PR #211

Issue #109, opened 2015-09-21
Difference of R3600's timing resolution between WCSim and SKDetSim


YujiOkajima:
I compared the hit timing resolution of R3600 between WCSim and SKDetSim.
As a result, the timing resolution of low energy seems to be different.
![timing_resolution_r3600_compare](https://cloud.githubusercontent.com/assets/8589041/9989444/4480d918-6093-11e5-8c86-44a7ed4b4cf9.png)
Blue line shows the timing resolution in WCSim, and black markers shows the timing resolution in SKDetSim.

I tried to fit the graph based on the function used for Box & Line PMT.
As a result, the timing resolution was well fitted like the red line in the figure below.
![timing_resolution_r3600_compare3](https://cloud.githubusercontent.com/assets/8589041/9989932/b9f4cb74-6097-11e5-9f54-867f7b4be717.png)

eosullivan:
How does this work in skdetsim? Is there a separate timing resolution function written in the code that is just for low energy? If so, is it easy to track down what the differences are between the low E function in skdetsim and the function in WCSim?

YujiOkajima:
@eosullivan 
The process about hit timing smearing seems to be different between ATMPD mode and LOWE mode.
However, the both timing resolution table seems to be the same as shown in my previous comment.
In SKDetSim, timing resolution is introduced by table.
The below figure shows the timing resolution of R3600 in WCSim and SKDetSim with wide range
![timing_resolution_r3600_compare4](https://cloud.githubusercontent.com/assets/8589041/10015848/bc817d4e-615d-11e5-8156-302c9257c08d.png)

eosullivan:
I don't understand what you mean by "both timing resolution table seems to be the same as shown in my previous comment". Which timing resolution tables seem to be the same?

PR #193, opened 2016-10-18
Add options for timing and charge resolutions for PMT hits

tdealtry:
Implemented by truncating the floating point values to the user-specified precision
See #188 for discussion

Standard verification. KS tests are 1

electrontest_digit_time_rounding

PR #203, opened 2017-02-19
modify timing resolution based on SKDetSim

YujiOkajima:
Based on the #109, 20 inch PMT timing resolution function is modified.
(timing resolution table of SKDetSim is in /usr/local/sklib_g77/skofl_16c/const/tdcres.dat)

I checked the relationship between charge and the resolution between nominal PMT with 0.9-1.1 p.e. charge and the other PMTs.
the result is shown in the figure below.
tres_q_withthr
The red solid line is the expected value, dashed red line is expected value without correction at Q<0.5,
and black dots shows the MC result.
(Error bar is the fitting error of hit time )
Q<0.5 correction seems not to be introduced in SKDetSim, but this request does not correct it.

From this result, the MC result seems to follow the expected value.

### DIFF
--- a/src/WCSimPMTObject.cc
+++ b/src/WCSimPMTObject.cc
@@ -77,11 +77,21 @@ G4double PMT20inch::GetExposeHeight() {return .18*m;}
 G4double PMT20inch::GetRadius() {return .254*m;}
 G4double PMT20inch::GetPMTGlassThickness() {return 0.4*cm;}
 float PMT20inch::HitTimeSmearing(float Q) {
-  float timingConstant = 10.0; 
-  float timingResolution = 0.33 + sqrt(timingConstant/Q); 
-  // looking at SK's jitter function for 20" tubes
-  if (timingResolution < 0.58) timingResolution=0.58;
-  float Smearing_factor = G4RandGauss::shoot(0.0,timingResolution);
+  G4float sig_param[4]={2.395,0.649,0.6002,2.307};
+  G4float lambda_param[2]={0.7782,0.05526};
+  
+  G4float sigma_lowcharge = sig_param[0]*(exp(-sig_param[1]*Q)+sig_param[2]);
+  
+  G4float highcharge_param[2];
+  highcharge_param[0]=2*sig_param[0]*sig_param[1]*sig_param[3]*sqrt(sig_param[3])*exp(-sig_param[1]*sig_param[3]);
+  highcharge_param[1]=sig_param[0]*((1-2*sig_param[1]*sig_param[3])*exp(-sig_param[1]*sig_param[3])+sig_param[2]);
+  G4float sigma_highcharge = highcharge_param[0]/sqrt(Q)+highcharge_param[1];
+  
+  G4float sigma = sigma_lowcharge*(Q<sig_param[3])+sigma_highcharge*(Q>sig_param[3]);
+  G4float lambda = lambda_param[0]+lambda_param[1]*Q;
+  G4float totalsigma = sqrt(sigma*sigma+1/lambda/lambda);
+
+  float Smearing_factor = G4RandGauss::shoot(0.0,totalsigma);
   return Smearing_factor;
 }
 


eosullivan:
Ok, thanks for putting this in. Please update us once the validation is completed. 

Also, can you remind me, have you checked if this is an issue for the other PMTs as well, or does this only affect the R3600 PMTs?

YujiOkajima:
I have validated the timing resolution by using the deviasion of hit time interval between nominal PMT with 1.9-2.1 p.e. charge and the other PMTs

compare_tres_large_mod

red curve is expected value, and black dots are MC result.

When the charge is small, the resolution seems to follow expected value.
However, when the charge is large, the resolution seems not to follow it.
I think this is because the hit time smearing method is changed.

When the Box & Line PMT performance is introduced, the timing resolution is set by using total charge of each PMT.
Currently, the hit time of each photo electron seems to be evaluated, and the smearing is done by using the charge from each photo electron.
So even if the total charge of PMT is large, each photo electron hit time is smeared with 1 p.e. level deviation (hits in a integration window is regarded as only one hit, and the hit time is set to the earliest hit time in the window, and the charge is integrated).

I will try to introduce the parameter to old WCSim and check the timing resolution again.

@eosullivan 
this issue does not affect to current Box & Line PMT parameters.
On the other hands, the timing resolution function of other size R3600 like PMTs, like 8 inch PMT, follows the same style function as 20 inch R3600.
(20 inch PMT before the modification: σ[ns]= 0.33+sqrt(10.0/charge[p.e.]))
(8 inch PMT: σ[ns]= 0.33+sqrt(1.890/charge[p.e.]))
So it is posible that small size PMT has worse resolution than real. However I can not modify it because I don't have the resolution table.

YujiOkajima:
I have validated the timing resolution by using the RMS of hit time interval between nominal PMT with 0.9-1.1 p.e. charge and the other PMTs with old version WCSim (2015 Sep. 1 version)

timing_resolution_compare_1pe

From this figure, the timing resolution seems to be introduced correctly.

tdealtry:
@YujiOkajima, could you redo the validation with my branch for #194? This would just be a cross check to make sure that the hit time smearing is again doing what it should be.

YujiOkajima:
@tdealtry 
All right, I will try to do it.
Should I simulate basically following to unmodified daq.mac setting?
When I tried to use the up-to-date version WCSim, the integration window range was set to 1 ns because the hit time smearing seems to be done for all hits in the window.

tdealtry:
Yes, try the integration window as is - the smearing is moved back from the hits to the digits in that pull request so you shouldn't have to fiddle anything. There may be some subtle differences because the order of adding dark noise/digitising/triggering has been corrected since September 2015 (release was v1.5.0, 1st Feb 2016), but the defaults in daq.mac are the same as in the old code.

eosullivan:
@YujiOkajima, could you rerun your validation with the new develop branch and with Geant4.10.1 (in addition to @tdealtry's comments above).

YujiOkajima:
@tdealtry 
I'm sorry for delay.
I have evaluated the timing resolution based on your branch.
In order to fix the trigger time, I have modified your code a little.
The result is shown in the figure below.
The red solid line is the expected resolution, and black dots are MC result.
tres_1pe_fixtrig
From this figure, the timing resolution seems to be introduced correctly.

@eosullivan 
I will try to run with new branch.
However, geant4.10.1 is not introduced in my working space (sukap).
I tried to use geant4.10.03.p01, but error was output when it is compiled.
The error message is shown below.

src/WCSimVisManager.cc:134:36: error: no matching function for call to 'WCSimVisManager::PrintAvailableGraphicsSystems()'
     PrintAvailableGraphicsSystems ();

Maybe I can run it in the same environment as Yoshida-san, but currently I can not 
connect to the server.
I will do it when the network is recovered.

tdealtry:
Thanks @YujiOkajima this is a good validation of #194.
> In order to fix the trigger time, I have modified your code a little.

Can you please explain what you did exactly?

YujiOkajima:
@tdealtry 
I changed the line 242-243 of src/WCSimWCTrigger.cc.
from
      triggertime = digit_times[this_ndigitsThreshold];
      triggertime -= (int)triggertime % 5;
to
      triggertime = 0.;

eosullivan:
This pull request and #194 seem like they will impact each other. @YujiOkajima and @tdealtry, is there a preferred order that these should be merged in?

tdealtry:
The first of @YujiOkajima's 3 plots suggests to me that currently the timing resolution isn't correct for any PMT. The third of the 3 plots suggests #194 will fix it. So you probably want to put #194 in first.

eosullivan:
Ok, #194 is now merged, so please revalidate this pull request so we can verify that it still works with Geant4.10 and that you don't need any further modification to the code now that #194 is on the develop branch.

eosullivan:
@YujiOkajima, have you had a chance to look this over? We will be making a new release of the code soon and I think all that this needs is to be revalidated with Geant4.10 using the most recent version of the develop branch.

YujiOkajima:
@eosullivan 
I'm sorry for delay.

When Geant4.10.3 is used for this simulation, the same error as I mentioned in the comment posted on May 1 is output, so currently I can not analyze it.
Currently I have been generating MC by using Geant4.10.0.
Maybe the result can be shown in today.

YujiOkajima:
The analysis is finished.
I have evaluated with the same software modification as May. 1.
And I checked the timing resolution from all PMT hits (different setting from previous one).
 
timing_resolution_validation_20170902

Red line is the expected value and black dots shows the analyzed value from MC outputs.
As shown in this figure, the resolution seems to follow the expected value well.

eosullivan:
I thought the fix from May 1 was because #194 wasn't merged yet, but now it is. Are you validating with the most recent development branch?

YujiOkajima:
@eosullivan 
No it isn't.
May. 1 version is Tom's branch. So I think #194 modification is already introduced.
I want to remove the trigger time shift to evaluate timing resolution without comparison with 
reference PMT hit time.
Currently I'm generating the MC again without any modification.
I will show you the hit time resolution by the same style as May. 1 and more past figures.

By the way, the error with Geant4.10.03.p01 seems to be caused by
G4VisManager.hh modification.

void PrintAvailableGraphicsSystems   () const;
is changed to
void PrintAvailableGraphicsSystems   (Verbosity) const;
So this error may be happen.

YujiOkajima:
Currently I have checked without any change of the code and evaluate the timing resolution with reference PMT as shown in May 1.
But the resolution shows the large difference between MC output and expected.
The tendency is not shown in the May. 1 figure.
Therefore the improvement after #194 is possible to affect this analysis.

timing_resolution_20170904

tdealtry:
@YujiOkajima thanks for doing this. Can you please clarify a few things:
- [The most recent plot](https://github.com/WCSim/WCSim/pull/203#issuecomment-326873103) uses the current WCSim develop branch, with no modifications?
- [2 plots ago](https://github.com/WCSim/WCSim/pull/203#issuecomment-326746092) uses the current WCSim develop branch, with the modification: `triggertime = 0.;`?

- From your explanation of the plots, I don't see how the variable trigger time should have an effect; for a given trigger, all the digits have the same time offset. Could you please explain your method in more detail? Also, please explain the input simulation to your analysis.
> deviasion of hit time interval between nominal PMT with 1.9-2.1 p.e. charge and the other PMTs

YujiOkajima:
@tdealtry 

・The most recent plot uses the current WCSim develop branch, with no modifications?

Yes. I downloaded the current WCSim develop branch without any modification.

・2 plots ago uses the current WCSim develop branch, with the modification: triggertime = 0.;?
Yes. I changed only the triggertime to 0.

・From your explanation of the plots, I don't see how the variable trigger time should have an effect; for a given trigger, all the digits have the same time offset. Could you please explain your method in more detail? Also, please explain the input simulation to your analysis.

All right, I will describe my analysis method.
This method is used to remove the trigger time uncertainty.
To check the timing resolution of PMTs, the difference between digitized and true hit time have to be checked.
But the standard time is different between them.
In order to correct this, the difference of standard time is estimated by using 
the difference of digitized and true hit time of reference PMT.
The reference PMT is selected from that with certain charge (0.9-1.1 p.e. for )
After that the distribution of 
(Digitized hit time of each PMT - True hit time of each PMT) - (Digitized hit time of reference PMT - True hit time of reference PMT)
is made and check the deviation value.

In this study, the estimated trigger time has uncertainty from the timing resolution of reference PMT.
This effect is corrected considering the reference PMT has 1 p.e. charge.

Because when trigger time set to constant value (=0) the shift is disappeared,
trigger time is possible to be changed in one event.
But I don't understand why such error is shown.
Current analysis uses only one trigger. 
Therefore even if the trigger is generated 2 or more, I thought it will not affect.

You can see the simulation setting macro file on

/disk01/usr5/okajima/work/WCSim/WCSim_test_Tom/WCSim_test_lowthr.mac
/disk01/usr5/okajima/work/WCSim/WCSim_test_Tom/daq_lowthr.mac

I changed only dark rate and energy of electron in the WCSim.mac
and global threshold in daq.mac.

Maybe I found the origin of error.
The digitized hit time seems to be integer type and discreet.
Because true hit time is continuous value, the difference is shown as error.
The below figure shows the (digitized hit time)-(charge) distribution (bin width of hit time: 0.1 ns)

qvst_readpmt

I think it is necessary to be fixed.

YujiOkajima:
I think it is happen because the "digit_time" in the line 359 of WCSimWCTrigger.cc is defined as integer.
This value seems to be digitized hit time and it should be continuous value.
I also checked the (digitized hit time)-(charge) distribution (bin width of hit time: 0.125 ns) 
changing this variable to float type.

qvst_distri_corrected

tdealtry:
Thanks for the explanations @YujiOkajima. Still a couple of questions:
- Does the time resolution analysis use just one event? (I don't have access to the .mac files; it'd also be good for posterity to post them here)
- Does changing the digit time from int to float fix the previous plot? (i.e. do the black points line up on the red line?) I shouldn't expect so, as there is the same issue with discreet digit times whether the trigger time is set as normal, or fixed to 0

The int->float change is implemented as part of the pending #193. Note however that this still gives the option for non-continuous digit times, which is the case for realistic electronics. The default value is set to 0.4 ns (see Sec 5.1 of the [SK NIM paper](http://www-sk.icrr.u-tokyo.ac.jp/~masato_s/class/sk-detector.pdf))

YujiOkajima:
The timing resolution did not changed even the type of digit_time from int to float.
And The difference between true and digitized hit time of reference PMT seems to be changed.
In previous version, the algorithm worked correctly, so something change affect to the reference PMT.
It have to be checked.

eosullivan:
Ok, please try and explore why your change seems to not be working. I don't think we would wait on this for the production run, since this only affects the 20" PMTs and not the B&L PDs which would be used in the HK runs we'll be doing, but it would be nice to have for someone who might want to do comparisons to the 20" PMT.

Issue #140, opened 2016-01-19
Solve warning deprecated headers


tfeusels:
Many annoying warnings appear during compilation about the deprecated header: #include < ext/hash_map > in WCSimDetectorConstruction.hh.
Compiling with gcc 4.4.7. 
This header should be replaced for a cleaner compilation.
According to backward_warning.h, the replacement header should be: < unordered_map > but the namespace and all hash-related statements need to be updated accordingly as well. The latter can only be used with new C++11 though: -std=c++0x

tdealtry:
Closed via #218

PR #204, opened 2017-02-22
Fix/root opts cmake

tfeusels:
Could not compile through cmake due to new RootOptions in ROOT output. This pull request fixes that by adding them to CMakeLists.txt. It also fixes compiler warning about hidden virtuals from TOption.
@tdealtry, can you quickly check the Print() fix? I think I fixed it correctly but a cross check to reduce that warning is helpful. 
Less compiler warning makes it easier to spot real compiler errors.

### DIFF
--- a/src/WCSimRootOptions.cc
+++ b/src/WCSimRootOptions.cc
@@ -25,7 +25,7 @@ WCSimRootOptions::~WCSimRootOptions()
 }
 
 //______________________________________________________________________________
-void WCSimRootOptions::Print(Option_t *option) const
+void WCSimRootOptions::Print(Option_t *) const
 {
   cout
     << "Detector construction:" << endl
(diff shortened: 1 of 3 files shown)


tdealtry:
Thanks for doing this and sorry for forgetting about CMake.
Good spot on the compiler warning. #140 really needs fixing...

src/WCSimRootOptions.cc
```diff
@@ -25,7 +25,7 @@ WCSimRootOptions::~WCSimRootOptions()
 }
 
 //______________________________________________________________________________
-void WCSimRootOptions::Print()
+void WCSimRootOptions::Print(Option_t *option) const
```
tdealtry: Can you change this to `(Option_t *)` or `(Option_t * /*option*/)`, to get rid of a compiler warning? :wink:

tfeusels:
#140 is not fixed with this pull request. That fix is coming in the next pull request.
Thanks for the fine tuning. Will fix this immediately.

PR #205, opened 2017-02-23
Add option in macro to switch off QE completely.

tfeusels:
For comparing different PMT types and different sizes and also for debugging purposes it has been proven very useful to switch off the QE completely. Instead of hacking the QE curves, this option makes it easy possible through the macro.


eosullivan:
I assume that it takes a while to run with this mode turned on, but it seems like a useful feature to have in the code.

tfeusels:
That all depends on what you're simulating (single PMT mode vs full detector) and why you want to turn this mode on (eg. understanding single photosensor response). This will probably take as long as the mode where the QE is only applied when photons hit the cathode.

eosullivan:
Can you check to make sure that this still behaves as you would expect after the merge of #211 and with Geant4.10.1? I would be surprised if there was any difference, but would like to make sure.

eosullivan:
I see that you updated your branch and pushed a new commit. Has the verification been completed, and if so, is it ready to go?

tfeusels:
I compared the PMTQEMethod "Stacking_Only" (filename "opt1") to the new one "DoNotApplyQE" (filename "opt2"), which in both the StackingAction as in WCSD does not apply the QE, with the standard electrontest.mac (so verification with SK detector). 
As expected the QE-off one runs much slower and gives you ~4 times more hits (the average QE over the Cherenkov spectrum). So also with G4.10.1 everything works as expected and this is good to go.

This option is mainly useful for single photon particle guns for understanding PMT behaviour and acceptance, to make sure your simulations are not 1/4 efficient. 
compare_qeoptions

PR #206, opened 2017-02-23
Adding a folder for macros 

tfeusels:
Addressing issue #81: adding a folder "macros" (which I already use in my personal local branch).
I copied the vis and daq macros over there to start. All useful macros should go there for reference.
This can later on get further subdirectories if it gets too crowded with macros.
If requested I can remove the vis and daq.mac macros from the main folder, but need to update WCSim.mac and CMakeLists.txt then accordingly.


tdealtry:
Useful change! But don't you want to move (rather than copy) these macros?
If so, you need to change a line in `WCSim.mac`:
`/control/execute daq.mac`
to
`/control/execute macros/daq.mac`

tfeusels:
Yes, I just wanted to get feedback first. I'll go ahead with it then and then the pull request should be ready to go.

JostMigenda:
Useful change indeed! Unfortunately, though, it seems a few files (`jobOptions.mac`, `jobOptions2.mac` and `tuning_parameters.mac`) are currently getting deleted instead of moved. @tfeusels, can you please check that again?

tfeusels:
@JostMigenda Thanks for checking! Sorry about that. Forget to git add them. Now it should be good to go.

eosullivan:
This is a good first step that cleans up the WCSim home directory. It would be nice to have an "examples" subfolder in macros/ where we could put WCSim.mac, to start, and eventually perhaps put more complicated examples.

tfeusels:
I would rather have subfolders in a separate pull request, unless you want to just copy WCSim.mac over there, which is trivial. If you propose to move it, this will change all relative paths in WCSim.cc and WCSim.mac and will make this a bit messier.
It's convenient (and assumed in WCSim.cc, hence the binary) that the main macro is in the same place as the executable.

eosullivan:
Ok, let's leave it like this and make improvements in a future pull request.

eosullivan:
After the recent batch of merges, this pull request now has unresolved conflicts. Can you go through and resolve them?

tfeusels:
Conflicts resolved.

Issue #149, opened 2016-02-03
Muons invisible in some visualizations


tfeusels:
WCSimVisManager.cc defines the colour for mu+ as white and mu- as black. As the OGLSX and OGLQt uses black as background, mu- are invisible. It seems that the RayTracer background is white (?) so mu+'s will be invisible. Any other colour for muons would be better.

PR #207, opened 2017-02-23
Fix issue149 

tfeusels:
This pull request addresses issue #149 by making mu-'s white, mu+'s silver and adding background colour options for visualizations. Now RayTracer is also black as OGLS's default.


eosullivan:
Neutrons are already grey, so they will look like mu+s? You could change the neutron colour to cyan or something to fix this. You should also update the README with the new colours.

Issue #208, opened 2017-02-28
Bug in photocathode implementation


tfeusels:
The issue is that photons that are passing through the PMT glass and then hit the Blacksheet, without passing through the Interior, so not going through the photocathode, are actually counted as hits.

This can be seen here (theBoundaryProcess->SetVerboseLevel(1)):
Photon at Boundary! 
 thePrePV:  TestSinglemPMT-glassFaceWCPMT
 thePostPV: WCBarrelCellBlackSheet                                         => **transition from glass to BlackSheet**
 Old Momentum Direction: (-0.993144,0.0110169,-0.116378)
 Old Polarization:       (0.000624568,-0.995035,-0.099525)
 New Momentum Direction: (-0.993144,0.0110169,-0.116378)
 New Polarization:       (0.000624568,-0.995035,-0.099525)
 *** Absorption *** 
WCSimWCDigitizerSKI::DigitizeHits START WCHCPMT->entries() = 1  =>**It's a hit?**

While this has been gone unnoticed due to the small glass thickness, relative to the size of 20" PMTs, this fakes too good small PMT performance (if the relative glass thickness is overestimated) and is anyway wrong.
Note: pull request with fix in progress...

Issue #100, opened 2015-07-27
Use the proper R3600 PMT collection efficiency


eosullivan:
The R3600 PMT collection efficiency measured by Hamamatsu is 73%. We will implement this in the code and will remove the extra factors that have been mimicking this efficiency factor. This will require further investigation, but will likely be related to the ratio factor in the quantum efficiency routines, and the absorption length tuning. When this is done, the correctionFactor introduced to the B&L phototubes in PR #98 will be removed. This will be implemented after issue #9 since the absorption factor tuning will be sensitive to the physics used.

PR #209, opened 2017-03-01
Fix photocathode bug where photons were not going through the cathode.

tfeusels:
Problem: photons can go through the sensitive detector (glass)
and be killed in the next volume by Absorption, eg. in the BlackSheet, but will then
still be counted as hits here (issue #208 ). 
One solution would be to recode the cathode implementation as in G4/examples/extended/optical/LXe cathode implementation.
or easier: make sure they cross the cathode to be counted as a hit, as the logical Boundary "cathode" can't be a sensitive detector.
The latter fix is what this pull request does.

### DIFF
--- a/src/WCSimWCSD.cc
+++ b/src/WCSimWCSD.cc
@@ -110,6 +110,18 @@ G4bool WCSimWCSD::ProcessHits(G4Step* aStep, G4TouchableHistory*)
       &&(particleDefinition == G4OpticalPhoton::OpticalPhotonDefinition()))
     return false;
 
+  G4StepPoint        *postStepPoint = aStep->GetPostStepPoint();
+  G4VPhysicalVolume  *postVol = postStepPoint->GetPhysicalVolume();
+  //if (thePhysical)  G4cout << " thePrePV:  " << thePhysical->GetName()  << G4endl;
+  //if (postVol) G4cout << " thePostPV: " << postVol->GetName() << G4endl;
+
+  //Optical Photon must pass through glass into PMT interior!
+  // What about the other way around? 
+  // TF: current interior won't keep photons alive like in reality
+  // Not an issue yet, because then interior needs to be a sensitive detector, when postStepPoint is the glass.
+  if(postVol->GetName() != "InteriorWCPMT")
+    return false;
+
   //  if ( particleDefinition ==  G4OpticalPhoton::OpticalPhotonDefinition() ) 
   // G4cout << volumeName << " hit by optical Photon! " << G4endl;
     


eosullivan:
Do you have some plots to show how this will affect basic outputs? If this significantly decreases the number of photons we detect, it may already be accounted for in the tuning of the simulation.

tfeusels:
This will (and should) reduce the number of hits detected per PMT. The change will depend on glass thickness and PMT size, which I doubt is accounted for any tuning parameter. They are all constant scalings as far as I can see. Also, current behaviour is wrong and has angular dependence as well, ie. it depends how the photons scatter in the PMT glass and hit the inside (and should count as hit) or hit the back side (blacksheet) and should not count as hit (which they currently do).

The geometric (projected) photocathode area is currently overestimated because of this:
- for 20" PMTs (4cm glass thickness): 3.5% overestimation
- for 8" PMTs (5.5cm glass): 10.7% overestimation
- for 3" PMTs ("if one assumes 4cm glass, which is too thick"): 24% overestimation
of detection efficiency due to too large photocathode area coming from the glass-blacksheet interface.
 The actual reduction in detected number of hits depends of course on the angular distribution of photons hitting the PMTs so is more difficult to predict. I'll illustrate this asap with the standard verification plots.

tfeusels:
The standard verification comparison plots are shown in attachment (standard electrontest.mac uses SK tank and then 1000 500 MeV e- events, iirc). As expected will the number of hits after the fix be lower on the order of a few percent depending on the angle of incidence of the photons on the PMT glass.
compare_electrontest_devel_vs_fix209

eosullivan:
So, I see how having the glass as the sensitive detector can lead to an overestimation if the photon goes through the glass but does not pass into the PMT interior for whatever reason. However, my understanding is that the SK detector in WCSim is tuned to have the right light output, so imagine that there is some tuning factor that is already decreasing the light output by 4%. There may be a small angular dependance that is not correct, and this 4% factor may not be right for all PMTs, but by implementing this PR we will now be underestimating everything. I think this is an issue, and from your numbers it looks like this will become even more of a problem for the smaller PMTs that could be used for the OD, but we should think of a fix that won't affect the 20" PMTs. Then, when we redo our tuning, we can fully put this is and tune with the cathode properly coded.

tfeusels:
Then a retuning is very urgent and should be attached to this pull request.  I noticed at least two "arbitrary fudge/tuning" factors in WCSim for which I couldn't find a reference or motivation. If you want to keep the total light output to be unchanged for WCSim SK, for which I assume these factors were chosen, then I would prefer to just update those factors and have this pull request fix the proper angular dependence and PMT glass thickness and size dependence. 
The two factors are:
- rgcff=0.32 in tuning_parameters.mac: which can be reduced to increase the light output.
- ratio in WCSimStackingAction.cc, coming from skdetsim but not really clear to me what the reason is for this factor.

If we want to keep the light output from WCSim SK the same (assuming it matches skdetsim and matches data real PMTs), then I would suggest to update rgcff in tuning_parameters for now together with this photocathode bugfix. So, the rgcff tuning parameter to me looks like the one absorbing the 4% difference now as it's just a scaling factor.

eosullivan:
We have been due for a retune for some time now (see this [milestone](https://github.com/WCSim/WCSim/milestone/5)). The last retune was done before my time, but my understanding is that it was fairly time-intensive. I will have to dig up the information on how it was done before, and this will realistically not be started until the end of the summer, at the earliest. 

I hate adding in these hack factors, but if the light output is really being affected for these smaller PMTs, we can think about how to add this in. Check out issue #100 and the related PRs to see how we did a similar implementation for the R3600 collection efficiency. This is even more complicated since the factor is related to the PMT-specific geometry, but we do have an angular-dependant collection efficiency array already implemented in WCSimPMTObject.cc that could be done on a PMT-by-PMT basis, so that could be a way to add it in that is pretty intuitive to users, and that could be done for just the smaller PMT sizes.

cvilelasbu:
Hi @eosullivan,

Apologies in advance if this is a naive comment...

Looking at @tfeusels's validation with the 500 MeV electrons in Super-K (pointing upwards from the centre of the tank) it looks like this bug fix introduces a shift of ~1.4% in the energy scale. The Super-K energy scale uncertainty in the 2003 NIM paper is 1% at low energy (< 16 MeV, from LINAC and DT) and 2.6% at high energies (from a variety of control samples).

So the shift introduced by this bug fix is at most comparable to the energy scale uncertainty at Super-K, with the caveat that these are 500 MeV electrons at the centre of the tank, pointing straight up.

Is there any reason to tune WCSim to SKDETSIM to a precision better than the SKDETSIM systematic uncertainties measured at Super-K? If not, and pending validation at different energy scales, it seems to me that this PR fixes a clear bug in how hits are produced in WCSim without compromising the agreement between WCSim and reality (via SKDETSIM and Super-K data).

eosullivan:
@cvilelasbu, good point that the effect shown in the plot above is not so big for 20" PMTs (and less than I thought). @tfeusels, can you clarify? In the comments above you calculated that it was more like a ~4% effect. Is that for a more extreme case than the test particle gun you ran?

PR #210, opened 2017-03-28
Feature Outer-Detector (Hyper-K style)

P3tru:
OD geometry added to the ConstructCylinder() method, w.r.t the philosophy of the Hyper-K detector. The other configuration had not been touched and tests shows that the results as the same as before.
This pull request is the first part for the OD, the digitization will follow shortly after the review of this feature.

Several new parameters configuring the OD dimensions and photocoverage are in control of the user as for the ID. A new option, WCODPMTShift allows to defined a "shift" between the rows of the PMTs, to play with the grid without changing the photocoverage.

Some modification appears only in preparation for the digitization of the OD.

A bug has been found in the optical properties of the tyvek. Commenting the finish of the surface seems to correct the bug, has to be discussed.
At the moment, all the new geometry is contained inside ConstructCylinder(). With some modifications, one can create a new method ConstructOD() if it respects more the WCSim repo philosophy, which has to be discussed during the review I guess.


.gitignore
```diff
@@ -23,3 +23,11 @@ geofile.txt
 
 ##ignore compiled python scripts
 *.pyc
+
```
tdealtry: Think you want to remove all these changes to .gitignore

src/WCSimDetectorConfigs.cc
```diff
@@ -219,8 +219,8 @@ void WCSimDetectorConstruction::SetHyperKGeometry()
   WCPMTName           = PMT->GetPMTName();
   WCPMTExposeHeight   = PMT->GetExposeHeight();
   WCPMTRadius         = PMT->GetRadius();
-  WCIDDiameter          = 70.8*m; // = 74m - 2*(60cm ID wall + 1m OD)
-  WCIDHeight            = 54.8*m; // = 60m - 2*(60cm ID wall + 2m OD)
+  WCIDDiameter          = 70.8/6.*m; // = 74m - 2*(60cm ID wall + 1m OD)
```
tdealtry: I think the best thing to do would be to leave the HyperK geometry alone as is (for backwards compatibility), but add a new geometry construction method with the OD, callable from the .mac file with something like "HyperKWithOD"

tfeusels:
Is it true that OD PMTs are only constructed based on photocathode coverage? Is there an option to make sure they're back to back with ID PMTs or does that require same coverage?

include/WCSimDetectorConstruction.hh
```diff
@@ -80,7 +82,8 @@ public:
   G4int    GetMyConfiguration()   {return myConfiguration;}
   G4double GetGeo_Dm(G4int);
   G4int    GetTotalNumPmts() {return totalNumPMTs;}
-  
+  G4int    GetTotalNumPmtsOD() {return totalNumPMTsOD;}
```
JostMigenda: I’d suggest renaming this to `GetTotalNumODPmts` and `totalNumODPMTs`, to be consistent with `Get_ODPmts` (and `ComputeWCODPMT`, `GetODTubeID` _etc._)

JostMigenda: … and similarly for `WCPMTODName` ➔ `WCODPMTName` and several others throughout this file. This is not about the name/radius/… of the “PMT outer detector” but about the “outer detector PMTs”, so I think consistently calling them `ODPMT` in all variable/function names will make the code slightly easier to read and modify later on.

src/WCSimDetectorConfigs.cc
JostMigenda: Also, I don’t understand this division by 6. Was this simply some testing artefact that accidentally made its way into the PR? Or is there some reason for it?

eosullivan:
@P3tru, I remember that you said that you used the visualizer to check the building of the OD PMTs. Can you include some of those figures here in the pull request description? It would be helpful to picture what is happening as I read through the code (and also is a good way to catch any problems).

src/WCSimConstructCylinder.cc
```diff
@@ -1307,8 +1666,159 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructCaps(G4int zflip)
...
+
+  } // END isODConstructed
+
+  return logicCapAssembly;
 
 }
 
 
+void ComputeWCODPMT(G4int NPMT, G4double *NPMTHorizontal, G4double *NPMTVertical){
+  switch (NPMT) {
```
eosullivan: Why is this function needed?

src/WCSimConstructMaterials.cc
```diff
@@ -803,7 +803,7 @@ void WCSimDetectorConstruction::ConstructMaterials()
 
    OpWaterTySurface->SetType(dielectric_dielectric);
    OpWaterTySurface->SetModel(unified); 
-   OpWaterTySurface->SetFinish(groundbackpainted); //a guess, but seems to work
+   // OpWaterTySurface->SetFinish(groundbackpainted); //a guess, but seems to work
```
eosullivan: Can you comment a bit more on this bug in the tyvek properties that you found?

src/WCSimDetectorConfigs.cc
```diff
@@ -232,6 +232,63 @@ void WCSimDetectorConstruction::SetHyperKGeometry()
...
+  ////////////////////////////////////
+  // OD Parameters --- Beta version //
+  ////////////////////////////////////
+  isODConstructed = true;
+
+  // OD Dimensions //
+  WCODLateralWaterDepth    = 1.*m;
+  WCODHeightWaterDepth     = 2.*m;
+  WCODDeadSpace            = 600.*mm;
+  WCODTyvekSheetThickness  = 2.0*cm; // Au pif
```
eosullivan: 2 cm seems pretty thick for tyvek. Maybe we should change this to 1 mm to be consistent with the ID code (though that number also looks like a guess). Also, do you think this should be hard-coded into the geometry? I don't have any strong feelings, but it seems like similar parameters are hard-coded in.

eosullivan:
I looked over the code and it seems like it is well implemented. I'll check out a copy of this pull request and play around with it a bit. In the meantime, can you include some visualizations to this pull request so we can see how the OD geometry looks?

src/WCSimDetectorMessenger.cc
```diff
@@ -144,6 +146,8 @@ void WCSimDetectorMessenger::SetNewValue(G4UIcommand* command,G4String newValue)
 		  WCSimDetector->Cylinder_12inchHPD_15perCent();
 		} else if ( newValue == "HyperK" ){
 		  WCSimDetector->SetHyperKGeometry();
+		} else if ( newValue == "HyperKWithOD" ){
```
eosullivan: You should also include this new option in WCSim.mac

eosullivan:
Ok, I played around with the code and looked at the OD PMTs with RayTracer and everything looks reasonable. Please have a look at my inline comments and add the visualization figures and we'll take it from there.

eosullivan:
One last comment (for now), you should also think about how to include the new OD geometry in the WCSim documentation. This currently exists as a separate repository (https://github.com/WCSim/WCSim-Documentation) so this would be a separate pull request that should be arranged to go in at the same time as this pull request.

src/WCSimConstructCylinder.cc
P3tru: This function computes the nb of PMTs on the barrel side if the user decides to not define WCPMTODperCellHorizontal and WCPMTODperCellVertical by setting them to 0. Therefore one has the possibility to only define a coverage and WCSim will compute the nb of PMTs to create itself. I added a description inside WCSimDetectorConfigs, SetHyperKWithODGeometry() method.

src/WCSimConstructMaterials.cc
P3tru: Unfortunately I haven't found that much information on the bug. Perhaps someone already familiar with material optical properties in Geant4 could enlighten us? Regarding for the OD, it doesn't really matter because I will submit in the next PR the WLS plates behind the PMTs. I checked by varying the different options for optical materials surface and couldnt see a difference whatsoever for light leaking between ID/OD.

src/WCSimDetectorConfigs.cc
P3tru: Gotcha, I am also done with an other PR with OD parameters inside the messenger class, so the user could vary all OD parameters straight from the macro files. Apologies it took so long...

P3tru:
Please find a few figures of the Hyper-K OD implementation.
There is a full Hyper-K OD, and a zoom for an Hyper-K at 1/36th of its real size, to have better visualisation. You can also find pictures with twice much PMTs and an other one with a shift between PMTs rows (by changing the OD parameters).

hyper-k_36th
hyper-k_36th_4pmtsod
hyper-k_36th_withshift
hyper-k_full

src/WCSimConstructMaterials.cc
eosullivan: But, how do you know there is a bug? What kind of error were you getting or what kind of behaviour were you seeing?

eosullivan:
What's up with the top and bottom OD PMTs floating in space? Are some of the materials set to invisible to make visualization of the PMTs easier?

src/WCSimConstructMaterials.cc
P3tru: I get a seg fault during execution, and it seems to execute now on my WLS plates branches.

P3tru:
Yes exactly, If I show the tyvek on caps it's very hard to see the caps PMTs. I can try to produce one if you rather like.

eosullivan:
Ok, I think this looks basically ready to go. Now that there are no further coding changes, can you run some validation? I would like to see that the code output doesn't change for the usual Hyper-K geometry (shouldn't, but will make sure that there were no changes to the main part of the code) and also that there are no changes when you compare the usual Hyper-K geometry with the new HK-OD geometry (since we aren't reading out the OD yet, these should still be the same).

P3tru:
Ok sure ! Do you want me to proceed as we did with the G4 PR ? Comparing charge/timing hits on both version with Hyper-K and Hyper-K OD ?

eosullivan:
Yes, exactly. Thanks.

P3tru:
Some slides attached comparing upstream/develop with origin/feature-OD, with comment and questions.
[HyperKOD.pdf](https://github.com/WCSim/WCSim/files/1497175/HyperKOD.pdf)

P3tru:
Please find additional validation plots after our discussion during the software meeting.
You were absolutely right I had introduced a bug affecting the volume of the innerAnnulusRadius and outerAnnulusRadius of the cylinder structure for all geometries of WCSim. 

Increasing this space seems to increase the nb of mean photons collected on the PMTs by a tiny percentage. I made a study without the OD geometry for Hyper-K and varying artificially this "deadspace" and find compatible results.
Also the presence of tyvek sheets affects the nb of photons collected and increases a bit too.

I pushed the commit with the bug fix and now I have exactly the same results between origin/develop and this branch for standard Super-K and Hyper-K geometries.
[HyperKOD.pdf](https://github.com/WCSim/WCSim/files/1499939/HyperKOD.pdf)

eosullivan:
Ok, I guess this is maybe what we would expect. Having tyvek behind the PMTs could increase the probability for a photon to be reflected back into the ID, and having extra water behind the PMTs could mean that photons that are created in the new dead space and being reconstructed inside the ID. This pull request will not change any of the existing geometries and the validation seems to be giving sensible results for the new OD geometry. I am satisfied. Is this ready to go in?

PR #211, opened 2017-04-13
Set FTFP_BERT as default physics list

Tomoyo-Yoshida:
This pull request is motivated by WCSim migration to Geant 4.10.1 from 4.9.4.
This branch is forked from Stephane's pull request 202.

FTFP_BERT physics list is selected instead of WCSim physics list in jobOptions.mac to simulate pion hadronic interactions properly. Some validation plots will come.


Tomoyo-Yoshida:
Here are some validation plots of Geant4.10.1 FTFP_BERT physics lists. All results are as expected, except for a long tail observed in total charge of 10GeV e-. For more detail, please find the attached slides.
[yoshida_G4.10_FTFP_BERT_validation.pdf](https://github.com/WCSim/WCSim/files/925133/yoshida_G4.10_FTFP_BERT_validation.pdf)

tfeusels:
Shouldn't this pull request then also remove the WCSim physics list from pull request 202 as this is shown to not work? From G4.10.1 onwards the user should use this default physics list and not have the freedom to make serious mistakes by choosing the previous/WCSim defined one.

Tomoyo-Yoshida:
A comment about the  long tail observed in total charge of 10GeV e- in the slides I attached.
I found that FTFP_BERT physics model includes electro-nuclear process while WCSim physics list does not. Sometimes, GeV order electrons are scattered or generate pions via interactions with protons or oxygen nucleus, which consist of low charge component.

eosullivan:
I agree with @tfeusels that the WCSim physics list should be removed in this pull request since we have found this unexplained behaviour with Geant4.10. Also, could you add a reference for your findings on the extra electro-nuclear processes that the FTFP_BERT includes? I would like to understand that a bit better.

tfeusels:
While I'm not suggesting to use this list as default (yet?), the neutron people should have a look at the variation: FTFP_BERT_HP (with HP as High Precision Neutron model for neutrons < 20 MeV). I know that using HP makes the simulation slower, so it might be interesting as an optional list if this model is a "decent" default model for simulations where neutrons are important while others should then probably use the faster FTPF_BERT.
Interesting, slightly older reference: http://geant4.slac.stanford.edu/MSFC2012/ChoosePhys.pdf

@eosullivan: To see which models are included by any physics_list, the standard Geant4 shows that.
@Tomoyo-Yoshida sent her WCSim logfile over e-mail (attached here) which outlines:
===================================
                          Hadronic Processes for e-
  Process: electronNuclear
        Model:   G4ElectroVDNuclearModel: 0 eV  ---> 1 PeV
====================================
As reference, the WCSim physics list (also attached logfile) does not include this.

More technical details:
the FTFP_BERT list is defined here: source/physics_lists/lists/include/FTFP_BERT.icc
which adds EmExtraPhysics. That includes lepton photonuclear reactions (constructors/gamma_lepto_nuclear/src/G4EmExtraPhysics.cc: G4BertiniElectroNuclearBuilder
which apparently can be switched off). And the G4BertiniElectroNuclearBuilder adds the above model (constructors/gamma_lepto_nuclear/src/G4BertiniElectroNuclearBuilder.cc). 

To me this looks very reasonable to include and use the FTFP_BERT out of the box. But we might want to list the used models (printed in the logfile) on the Wiki here.

[log_FTFP_BERT.txt](https://github.com/WCSim/WCSim/files/953382/log_FTFP_BERT.txt)
[log_WCSim.txt](https://github.com/WCSim/WCSim/files/953383/log_WCSim.txt)

eosullivan:
Ok, thanks for including this info here, both for me and for posterity. I propose that we merge this in and then @P3tru can make a pull request where the WCSim physics list is removed. So, are we in agreement that this validation seems complete (everyone, but especially @Tomoyo-Yoshida and @P3tru)? I am satisfied. Is this pull request ready to go?

Tomoyo-Yoshida:
Thank you for the information, @tfeusels.
Could you wait merging until tomorrow, @eosullivan? Now I changed the codes in my local branch to remove WCSim physics list and am redoing validation. I can include the deletion and also mention FTFP_BERT_HP physics list in jobOptions.mac, if you could wait till tomorrow.

eosullivan:
I think it's too premature to put anything about the FTFP_BERT_HP list, so leave that out. If you are in the process of removing the WCSim physics list and revalidating, then we can wait until tomorrow. If it looks like it will take longer, let's make removing the old physics list a separate pull request as I suggested.

Tomoyo-Yoshida:
Ok, then I put FTFP_BERT_HP aside.
Now I am generating small particle gun samples using modified code, and it seems to work properly so far. I think I can finish revalidation by tomorrow.

tfeusels:
I agree that the FTFP_BERT_HP is part of a separate pull request. Just wanted to mention it already here. This should probably be looked into when Gd is being studied or any neutron-related studies.

Thanks for the quick testing and removing the WCSim default list. Happy we're converging on this pull request.

Tomoyo-Yoshida:
Now WCSim physics list is removed and validation is finished. Total charge distributions are completely same as those before the removal.

P3tru:
Apologize for delay, I just got time to checkout Yoshida's code, compile, and test it with muons and pions at 500 MeV. I do confirm that I get the same behaviour as before the removal.

I don't see any reasons myself not to carry on with the merge and to abandon the custom WCSim physics list. 

In last Yoshida's commit, she also removed WCSimPhysicsList class completely, which is not the same as removing the option to use the list. Are we all OK with this ? I mean, are we sure in the future we won't need to build an other custom physics list ourselves ?

In my opinion, I will say no. FTFP_BERT as all the physics we need and we can add in the near future FTFP_BERT_HP for precise neutrons capture study, and we won't need to build a physics list "manually".

Thanks for having pushed these last commits without WCSimPhysicsList.

tfeusels:
For this pull request, I would remove the whole WCSimPhysicsList completely. 
In the future (6months+?), this should be replaced by the G4NeutCascadeResponse list, which ND280 will be using soon. The BERTINI model for pion inelastic inelastic interactions is way better than BINARY but still has some issues describing external scattering data (and there are some issues with the pion elastic interactions). The pion inelastic interactions are solved by using the NEUT cascade model for secondary interactions (as it is used for final state interactions.) The G4NeutCascade interface, which essentially replaces BERTINI by NEUT, has been under developed for awhile and will be default for ND280 soon, reducing their dominant detector systematic of pion secondary interactions. Currently we are finalizing the validation with G4.10.1 and in some future this should also become the default for HyperK.
 From the Geant4 side they are interesting in updating their BERTINI pion-inelastic tuning, so that default model might become better but I foresee this in the longer future. 
To implement the above, I can easily go through the history of this file, as adapt the validated G4NeutCascadeResponse code to work with WCSim. This should be a known issue to track which I might to take on at some point when it becomes high priority. 
Let's go ahead with the current version of this pull request. Just wanted to let everyone know what's coming at some point.

WCSim.cc
```diff
@@ -59,27 +57,14 @@ int main(int argc,char** argv)
   // Set up the messenger hooks here, initialize the actual list after loading jobOptions.mac
   WCSimPhysicsListFactory *physFactory = new WCSimPhysicsListFactory();
 
-  // Currently, default model is set to BINARY
+  // Currently, default physics list is set to FTFP_BERT
```
eosullivan: Add in a comment that we removed the custom physics list option in versions later than 1.6.0.

eosullivan:
If it becomes necessary to create a custom physics list in the future, that code will still be available in the repository. We could add in a comment about removing the custom physics list option in WCSim versions later than 1.6.0 so that it is clear where to find this. Then, I think this looks ready to go.

eosullivan:
Ok, this looks good. It should be mentioned that earlier discussion about this is in PR #202.

PR #212, opened 2017-04-21
fix indices saved to WCSimRootCherenkovDigiHit::GetPhotonIds

marc1uk:
The indices stored in WCSimRootCherenkovDigiHit::GetPhotonIds are incorrect. This makes it impossible to determine such things as signal-to-noise ratio in digits, or which/how many digits came from a given primary particle.

In theory these indices should represent the index of the corresponding photon hits in the WCSimRootCherenkovHitTime TClonesArray, as with the return of the (poorly named) WCSimRootCherenkovHit::GetTotalPe(0) method. But in actuality, the PhontonIds represent only the index of the photons **within the appropriate digit in the WCRawPMTSignalCollection**. This collection gets flattened to generate the CherenkovHitTime array, and in the process the relative offset of each entry within the collection is lost. Since not all hits get digitized, these offsets can't be recovered. 

Fortunately the fix is very simple - introduce a running counter that turns the relative index into an absolute one, as done in this PR. 

To validate, apply the attached patch, which introduces debug printouts into WCSimWCDigitizer.cc to print the tube, time, index and parent of each digit. Then run the attached ROOT macro to print the corresponding information from the wcsim.root output file. Note that not all entries printed in WCSimWCDigitizer.cc get digitized or are within the trigger windows, so not all will be in the ROOT macro output. By comparing the printed information, though, it's possible to match up the entries and verify the saved information is correct. 
Apply the patch with `git apply {patchfile}'
[validation_patch.txt](https://github.com/WCSim/WCSim/files/948036/validation_patch.txt)
Verify the output returned by the following ROOT macro
[parent_validation.txt](https://github.com/WCSim/WCSim/files/948035/parent_validation.txt)


tdealtry:
Thanks @marc1uk. I'm not sure this is required (please feel free to poke holes in my method to get the info in the example below) but it will make it much more convenient & fast (i.e. it removes the need for a loop over the CherenkovHits to find the hits with the correct PMT ID)

This example below that calculates the number of noise/Cherenkov/other hits for each digit, and can be inserted into a WCSim macro using the standard naming conventions

```
const int ncherenkovdigihits = wcsimrootevent->GetNcherenkovdigihits();
const int ncherenkovhits      = wcsimrootevent->GetNcherenkovhits();

// Loop through elements in the TClonesArray of WCSimRootCherenkovDigHits                                                                                                          
for(int idigipmt = 0; idigipmt < ncherenkovdigihits; idigipmt++) {
  //get the digit                                                                                                                                                                  
  TObject *element = (wcsimrootevent->GetCherenkovDigiHits())->At(idigipmt);
  WCSimRootCherenkovDigiHit *wcsimrootcherenkovdigihit =
    dynamic_cast<WCSimRootCherenkovDigiHit*>(element);

  //find out whether this is due to noise, a real photon, or both                                                                                                                  
  int tube_id = wcsimrootcherenkovdigihit->GetTubeId();
  int timeArrayIndex = -1;
  int peForTube      = -1;
  //loop through the WCSimRootCherenkovHit array to find the entry for this tube ID                                                                                                
  for(int ipmt = 0; ipmt < ncherenkovhits; ipmt++) {
    TObject *Hit = (wcsimroottrigger0->GetCherenkovHits())->At(ipmt);
    WCSimRootCherenkovHit *wcsimrootcherenkovhit =
      dynamic_cast<WCSimRootCherenkovHit*>(Hit);
    int tubeNumber     = wcsimrootcherenkovhit->GetTubeID();
    if(tube_id == tubeNumber) {
      timeArrayIndex = wcsimrootcherenkovhit->GetTotalPe(0);
      peForTube      = wcsimrootcherenkovhit->GetTotalPe(1);
      break;
    }
  }//ipmt                                                                                                                                                                          
  int n_noise_hits = 0, n_photon_hits = 0, n_unknown_hits = 0;
  if(timeArrayIndex == -1) {
    if(verbose)
      cout << "No PMT hits found for digit " << idigipmt << " with tube ID " << tube_id << endl;
  }
  else {
    if(verbose)
      cout << peForTube << " PMT hits found for digit " << idigipmt << " with tube ID " << tube_id << endl;
    //loop over the rawhits ids of hits that made up the digit                                                                                                                     
    vector<int> rawhit_ids = wcsimrootcherenkovdigihit->GetPhotonIds();
    if(verbose)
      cout << rawhit_ids.size() << " rawhits made up this digit" << endl;
    for(unsigned int irawhit = 0; irawhit < rawhit_ids.size(); irawhit++) {
      int this_rawhit = rawhit_ids[irawhit];
      if(verbose)
        cout << "Attempting to look for rawhit " << this_rawhit+1 << " in WCSimRootCherenkovHitTime array...";
      if(this_rawhit >= peForTube) {
        cerr << " There are only " << peForTube << " rawhits in this PMT" << endl;
        continue;
      }
      //now look in the WCSimRootCherenkovHitTime array to count the number of photon / dark noise hits                                                                            
      TObject *Hit = (wcsimroottrigger0->GetCherenkovHitTimes())->At(timeArrayIndex + this_rawhit);
      WCSimRootCherenkovHitTime *wcsimrootcherenkovhittime =
        dynamic_cast<WCSimRootCherenkovHitTime*>(Hit);
      const double hittime  = wcsimrootcherenkovhittime->GetTruetime();
      const int    parentid = wcsimrootcherenkovhittime->GetParentID();
      if(verbose)
        cout << " hit time " << hittime << " " << parentid << endl;
      if(parentid == -1) {
        n_noise_hits++;
      }
      else if(parentid < 0) {
        n_unknown_hits++;
      }
      else {
        n_photon_hits++;
      }
    }//irawhit                                                                                                                                                                     
  }//hits in this PMT found                                                                                                                                                        
}//idigipmt // End of loop over Cherenkov digihits                                                                                                                                 
```

marc1uk:
Hi Tom,
Thanks for the comment. I have to admit your method seems like it would work, I hadn't thought of using the TubeID to match up the Hits array with the Digits array. Thanks to that, I won't have to regenerate my simulation results. :) But nonetheless, the PR still seems like a good idea; the method is neither efficient nor by any means intuitive.

tdealtry:
Agreed, I like it.
Perhaps you could do (what I should of done a while ago...) and add (with this PR) an example of how to do the digit->hit truth matching in sample_readfile.C, with maybe signal-to-noise ratio of digits.

marc1uk:
Sure, I'd be happy to. I'm a little busy right now, but I'll add it soon.

marc1uk:
Accidentally closed the issue somehow. 
Anyway, I've added the long-awaited modifications to the example script. For the first 3 digits it pulls the photons, pulls the corresponding WCSimRootCherenkovHitTime and prints the index from the TClonesArray, the photon hit time, and it's parent track id. Hopefully that's enough to demonstrate things.

While I was modifying the sample script it I also made some other printout changes to clarify wording on things that I don't think are helpful. Things like printing `Jmu 1` or "Mode 8" is meaningless for someone who hasn't already dug through the code to see where they come from; the variables are poorly named to begin with, and this is a good place to explain things.  In fact, many of the printed parameters are only properly filled for nuance primaries, but the default primary source is an electron gun, for which many variables just print 0...

tdealtry:
Thanks for this - clarifying everything in the sample root script is a great addition to the PR.
Doing a close/open to force a Travis rebuild (it was initially failing due to this PR changing what's stored in the output, as expected)

Issue #162, opened 2016-04-24
Add Continuous Integration System


cwwalter:
Work was already done on this by Benjamin Richards etc. using Travis CI.  I didn't see any issue about it so am entering this so it is not forgotten.

tdealtry:
System is up & running again, so closing this issue

PR #215, opened 2017-05-02
ROOT wrapper: rootwc

tdealtry:
A root wrapper (`rootwc`) that automatically loads the relevant WCSim libraries and includes, so as to make running root macros in compiled mode much easier (it is of course good practice to run macros in compiled mode, for both speed and debugging purposes).
This wrapper also loads the T2K plotting style, to give WCSim output some consistency. Of course, this can be modified/removed as required.
This pull request also modifies `sample_readfile.C` for running in compiled mode (adds includes, removes bugs)
This is part of #88, but I feel is a useful addition that people might want to use before the whole bulk of that gets put in (or not, depending on what the solution of #162)

### DIFF
--- a/GNUmakefile_full
+++ b/GNUmakefile_full
@@ -48,10 +48,9 @@ DOXYGEN_EXISTS = 0
 endif
 
 .PHONY: all
-all: rootcint lib bin shared libWCSim.a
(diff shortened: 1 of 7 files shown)


tdealtry:
This was merged in via #285

PR #217, opened 2017-05-03
Fix floating point exception error

tfeusels:
I ran some particle guns with the new develop and encountered a few floating point exceptions (now handled by new Geant4's FPEdetection). These were caused by non-optical photons with a total energy of 0 in ProcessHits for which the wavelength calculation didn't make sense. Moved the wavelength calculation down to be only for optical photons.

The error encountered was: "ERROR: 8 - Floating point divide by zero."  This stops the full simulation and therefore this bugfix is urgent.


eosullivan:
Ah, weird that this didn't come up with the particle gun runs that were done for the verification steps. Anyways, it makes more sense for the wavelength definition to be in the optical photon part of the code, since that is the only place it is used. Good catch!

PR #218, opened 2017-05-05
Resolve compiler warnings of deprecated header.

tfeusels:
This pull request addresses issue #140. It does use C++11 features but most standard compilers (gcc 4.4 onwards) all include C++11 features or at least the ones being used. This is more a coding improvement and makes compilation much cleaner.


include/WCSimPMTObject.hh
```diff
@@ -16,9 +16,9 @@ public:
   virtual G4double GetExposeHeight()=0;
   virtual G4double GetRadius()=0;
   virtual G4float* Getqpe()=0;
-  virtual G4float* GetQE()=0;
-  virtual G4float* GetQEWavelength()=0;
-  virtual G4float  GetmaxQE()=0;
+  virtual G4double* GetQE()=0;
```
eosullivan: Why are you changing all the floats to doubles?

tfeusels: Otherwise you'll get a "narrowing conversion" error in C++11. You'll need double math inside the array.

eosullivan:
I just merged #199. Can you check that the switch to doubles doesn't significantly increase the output file size?

tfeusels:
I can check but it will take some time. The QE's are just used in calculations while for the output file size, it depends on what we save and those file types didn't change. Do we directly save the QE's in the output file now?

tfeusels:
Nope, it doesn't.

brichards64:
your most recent change has broken the compilation on travis

eosullivan:
I see this travis compilation error warning, and then a subsequent commit that seems to have fixed it. Is this ready to go or are you still working on it?

tdealtry:
Thanks for this @tfeusels I'll check the travis builds ok, and then merge

PR #221, opened 2017-10-19
Add some info on the trigger window

tdealtry:
Supplement to #117


eosullivan:
Can we check with an electronics expert what the window time in HK will likely be? It seems like the electronics group should already have some idea of what this number should be.

tdealtry:
@eosullivan I've modified the daq.mac comment to make it clear that it's up to analysers to decide if/how to change this - the point of the comment is to flag that this parameter is potentially important

And yes, 400 ns is what the DAQ group are currently using in the SN production

I'll leave it to you to close (either via merge, or not)

PR #222, opened 2017-11-08
Small PMTs options (3" and 5")

P3tru:
This PR add two new PMTs of 3" and 5" inside WCSim. This PR contains also the OD feature, to show compatibility.


src/WCSimPMTObject.cc
```diff
@@ -1634,3 +1634,331 @@ G4float BoxandLine12inchHQE::GetDarkRateConversionFactor(){
...
+  return wavelength;}
+
+G4float  PMT5inch::GetmaxQE(){
+  const G4float maxQE = 0.211;
+  return maxQE;
+}
+
+// Should be actual PMT Dark Rate, not effective dark rate in detector including other LE noise
+G4float PMT5inch::GetDarkRate(){
+  const G4float rate = 400*CLHEP::hertz;   //SKI value set in SKDETSim.
```
eosullivan: Is this a reasonable dark rate to use for these tubes?

eosullivan:
Ok, great. It's a bit hard to read with the OD commits in there, but we can get this cleaned up after PR #210 is included. Can you remind me, are both the 3" and 5" options currently being considered for the OD?

src/WCSimPMTObject.cc
P3tru: I used the measurements we made at QM from the ETEnterprise PMT's prototype we plan to use on the OD. Perhaps I can rename them to include the fact they came from ET ?

P3tru:
Yes I could revert the commit until the OD geometry, it was to show that both commit will merge without conflict. 
So at the moment option with 3" is favoured, but we keep working with 5" for cases with hybrid configuration. I won't say that 5" are unlikely for the final geometry but they are worth testing in any case.

PR #224, opened 2017-11-09
Docker Integration

brichards64:
Added a Docker file and linked repository to DockerHub so that we can have automated Docker Images produced on every repository commit.

These Docker images can be used to execute the code instead of compiling it for each specific system varient (and any problems that entails). They give a unified distribution OS with dependancies to give consistent running on any system irrespective of the machine type used.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,5 @@
+### Created by Dr. Benjamin Richards (b.richards@qmul.ac.uk) based on docker file from Stephane Zsoldos (s.zsoldos@qmul.ac.uk)
+
 ### Download base image from cern repo on docker hub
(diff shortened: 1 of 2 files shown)


eosullivan:
The commit history for this pull request is strange, with very old commits being included. Since this docker file is separate, could you just check out the current develop branch, add in the docker file, and redo this pull request?

eosullivan:
Also, if I wanted to try out this docker implementation, what would I do? This is just a file that allows docker to build this on an external website, right? Could you give some further instructions so people can try this out?

brichards64:
If made a new pull with a cleaned up commit history and also added
instructions to the readme

PR #227, opened 2017-12-01
Added Docker file for docker integration and updated README

brichards64:
Added instructions for usage in README and cleaned up commits


brichards64:
cleaned up commits and added usage instructions to the README.
once this push is made the automated system will make a docker image for the what ever the latest commit is so then you can test it.

We should also make images for the tagged releases too. So when you tag a release of WCSim you also tag the Docker Image.

PR #231, opened 2018-01-18
Small PMTs option (3" and 5" from ETEL)

P3tru:
Added small PMTs option of 3" and 5", inherited from 20" classes but with adapted QE and dark rates from the measurement made at QM


eosullivan:
Are there some plans to get more accurate measurements for these PMTs? I notice in this pull request that many variables (including the QE) are just a copy of the SK 20" PMT?

P3tru:
Yes, for the moment only size and dark noise are accurate. I have the measurement for the 3" option, don't know how I mixed up the commits. I'll push them right away

P3tru:
Just double check a second time and the good QE is already there for the 3" PMT option.
This is the plot of the datasheet QE :

P3tru:
We don't have much info on the 5" PMT yet. As soon as we can do measurements or we can get some more from ETEnterprises I'll implement them

src/WCSimPMTObject.cc
```diff
@@ -1634,3 +1634,332 @@ G4float BoxandLine12inchHQE::GetDarkRateConversionFactor(){
...
+      0.999994, 0.999994, 0.999994, 0.999994, 0.999995,
+      0.999995, 0.999995, 0.999996, 0.999996, 0.999996,
+      0.999996, 0.999998, 0.999999, 1.000000, 1.000000,
+      // Dummy element for noticing if the loop reached the end of the array
+      0.0
+  };
+  return qpe0;
+}
+
+//Currenly the PMT QE info is the same as 20 inch.
```
eosullivan: Ok, I see now that this QE isn't the same as the 20" PMT. So, you should remove this comment.

eosullivan:
When you add a new PMT, you need to scale the QE values by 1/0.73. The full gory details are here: https://github.com/WCSim/WCSim/pull/98 , but the short version is that there is some tuning of WCSim that is suppressing the photon output to match what we see with the 20" PMTs. This became an issue when we were adding the B&L PMTs with a QE measured outside the detector, and so we added a collection efficiency factor to the QE. Look at how this is done for the B&L PMTs on lines 1421- 1438 of src/WCSimPMTObject.cc [here](https://github.com/WCSim/WCSim/blob/develop/src/WCSimPMTObject.cc#L1421) .

P3tru:
Just pushed the correction factors for 5inch and 3inch QE, let me know if you also want to see some validation plot for the PR, or if it's fine to merge it as is.

eosullivan:
Ok, the change looks good. Some validation would be good, and also some characterization (what does the properties of these PMTs look like in WCSim? Is it the behaviour what we would expect from it's specs?)

P3tru:
I am not sure I understand what you asked. The only thing I changed for adding this small PMTs was to adapt the dimensions and QE, unfortunately the setup is not ready at QM to measure mapping or timing info right now. What do you have in mind to check the properties of these PMTs look like in WCSim ? I can do the same validation as before to see that SuperK and HyperK geometry are not affected by this new PMTs, since they are not constructed anyway.

eosullivan:
Yes, certainly the validation that shows our default geometries are unchanged is important. For characterizing the small PMTs, you could run a detector geometry that includes these PMTs and show the very basic characteristics match (you see less photons that the 20" PMT, the new PMTs don't crash the code when they are used, etc).

P3tru:
Ah OK I see ! I'll do that asap !

eosullivan:
Should this be in the new release, or can it wait until the next one? It kind of goes with this new OD functionality, but I would like to tag out this release soon. Do you have the validation plots where you use these tubes made?

P3tru:
Here comes a few plot for validation : 
- First, difference between dev and feature/smallPMTs
[dev_smallPMTs.pdf](https://github.com/WCSim/WCSim/files/2514282/dev_smallPMTs.pdf)
- Plots showing the charge collected for a Super-K detector made of 3" and 5" PMTs (in place of the 20") 
[5inch.pdf](https://github.com/WCSim/WCSim/files/2514283/5inch.pdf)
[3inch.pdf](https://github.com/WCSim/WCSim/files/2514284/3inch.pdf)
- And finally the difference between those, to see that they are different as expected 
[3_5.pdf](https://github.com/WCSim/WCSim/files/2514291/3_5.pdf)

I hope this convinces you that there are no crash due to this new feature and that everything behaves accordingly!

eosullivan:
I'm having a hard time following some of these plots. For dev_smallPMTs.pdf, what configuration are you running? And then in 5inch.pdf and 3inch.pdf, what is the red line?

P3tru:
- dev_smallPMTs.pdf : I compare the Hyper-K default geometry from the develop branch with the Hyper-K default geometry from my SmallPMTs branch. This is to ensure that my code doesn't cock up anything using the default geometry of Hyper-K
-  Then, 5inch.pdf and 3inch.pdf are made from the simulation of the Super-K detector where I replaced the 20" PMTs by 5" and 3" PMTs. This was made because the auto validation software compare everything with a default geometry, and I wanted to show that WCSim will not crash if you change the size of the PMTs, i.e my code is actually doing something :)
- And 3_5.pdf is the difference between the 3" and 5" configuration, to prove that the light collected is therefore directly linked as the size of the PMTs, as expected. The idea with these plots was to show that something was indeed happening.

Apologies for this very slow answer I hadn't realize what was the problem back in the time. I hope this will convinced you that the new PMTs are indeed available and they're not messing up with the dev version of WCSim.

eosullivan:
Ok, looks good. Can you resolve the conflicts and then I'll merge it in.

PR #232, opened 2018-01-23
Add improved application of QE in StackingAction

P3tru:
Add a new way to apply the QE inside the StackingAction class to reduce the nb of photons processed by WCSim when several PMT collection are defined when PMT QE do not overlap. This is necessary to simulate WLS plates accurately without working in the QEPMTMethod mode 3 or 4, where all the photons are processed and the simulation is considerably slowed.
I had to add a new PMTObject method called GetNbOfQEDefined(), returning the nb of points (wavelength) for which the QE is defined.
I haven't remove the previous method applying the QE in StackingAction, which leaves the user to switch between the 2 algorithm accordingly. 
In my opinion, using a TGraph instead of the "manual extrapolation" of the QE could also increase the speed of the simulation. Validation plot are on-going.

### DIFF
--- a/include/WCSimDetectorConstruction.hh
+++ b/include/WCSimDetectorConstruction.hh
@@ -93,8 +93,14 @@ public:
 
   G4float GetPMTQE(G4String,G4float, G4int, G4float, G4float, G4float);
   G4float GetPMTCollectionEfficiency(G4float theta_angle, G4String CollectionName) { return GetPMTPointer(CollectionName)->GetCollectionEfficiency(theta_angle); };
+  G4float GetStackingPMTQE(std::vector<G4String>);
 
   WCSimPMTObject *CreatePMTObject(G4String, G4String);
+
+  WCSimBasicPMTObject *CreateCombinedPMTQE(std::vector<G4String>);
+  WCSimBasicPMTObject *CustomPMT;
+  void SetCustomPMTObject(WCSimBasicPMTObject *PMT){CustomPMT=PMT;}
+  WCSimBasicPMTObject* GetCustomPMTObject(){ return CustomPMT;}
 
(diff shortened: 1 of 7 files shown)


brichards64:
Looks like you have compilation issues https://www.travis-ci.org/WCSim/WCSim/builds/332080271

src/WCSimDetectorConstruction.cc: In member function 'void WCSimDetectorConstruction::CreateCombinedPMTQE(std::vector<G4String, std::allocator<G4String> >)':
src/WCSimDetectorConstruction.cc:284: warning: '>>' operator will be treated as two right angle brackets in C++0x
src/WCSimDetectorConstruction.cc:284: warning: suggest parentheses around '>>' expression
src/WCSimDetectorConstruction.cc:284: error: 'QEMap' was not declared in this scope
src/WCSimDetectorConstruction.cc:284: error: '>>' should be '> >' within a nested template argument list
/root/HyperK/Geant4-10.1.3-Linux/include/Geant4/CLHEP/Units/PhysicalConstants.h: At global scope:
/root/HyperK/Geant4-10.1.3-Linux/include/Geant4/CLHEP/Units/PhysicalConstants.h:71: warning: 'CLHEP::electron_charge' defined but not used
make: *** [/root/HyperK/WCSim/exe/tmp/Linux-g++/WCSim/WCSimDetectorConstruction.o] Error 1

P3tru:
Thanks Ben for pointing that out, seems that my gcc7.2 has no problem to compile but need to rewrite the maps to be compatible with gcc4.4...

tdealtry:
Thanks for this @P3tru Given the description, I've had a look at #297 and it seems like it's included there. So @mahditaani can you let me know if it makes most sense to review/merge this before #297 (e.g. if you've made no further changes to the StackingAction), or to close this without merging this and just review/merge #297

PR #235, opened 2018-02-16
Feature/oglsqt

tfeusels:
OGLQt visualization support.


eosullivan:
Can you summarize what the advantage is for this vis package (vs the ones we already have)? I'm not so familiar with this one, and users who are looking to choose a visualization might be interested to know.

P3tru:
Maybe I can comment as I previously stolen Tom's code for visualizing the OD geometry :).

Qt driver add a nice interactive mode with interface to easily visualize the detector geometry and use for example the mouse to rotate/zoom on the detector. It's extremely useful for debugging purposes as you can quickly interact with the geometry, compare to the openGL driver for example.

With this feature, running the simulation without input (So just as './WCSim') will prompt the interface where you can load the visOGLQT.mac file and visualize the detector geometry, and I can confirm it doesn't affect the batch mode and run as usual when you use your usual WCSim.mac macro file.

Thanks again @tfeusels for this feature as in my study case for the OD was extremely useful !

Issue #234, opened 2018-02-07
comments in NUANCE input file?


JostMigenda:
For data management reasons, I think it would be good to have some way to keep track of how (and with what parameters) a NUANCE input file was generated; similar to what Tom implemented for WCSim’s output in #198.

Since the input file is simply a flat text file, its probably easiest to add a few lines that are "commented out" at the start of the text file. As far as I can tell from [documentation](http://neutrino.phy.duke.edu/nuance-format/), though, the NUANCE file format doesn’t have any way of adding comments. Would it be reasonable to add a custom syntax for this?

JostMigenda:
Looking at [the original nuance paper](https://www.sciencedirect.com/science/article/pii/S0920563202017565) and the [Internet Archive version](https://web.archive.org/web/20100612184104/http://nuint.ps.uci.edu/nuance/documentation.htm) of the original documentation web page, it seems pretty clear that there is indeed no comments syntax.

My suggestion would be to interpret lines starting with "#" as comments. (Simply because I’m hoping to use this comment syntax in the supernova event generator [sntools](https://github.com/JostMigenda/sntools), which is written in Python. If there are good arguments for using a different character, that’d be fine with me.)

This would be a simple change to WCSim (ignore lines starting with "#" in [WCSimPrimaryGeneratorAction.cc](https://github.com/WCSim/WCSim/blob/develop/src/WCSimPrimaryGeneratorAction.cc#L95)) and would still allow WCSim to use input files that don’t contain comments.

The only compatibility issue I see is that other software using the NUANCE input format might not be able to read files with such comments. This seems a negligible issue for two reasons, though:
1) I’m not aware of any other software that we’re currently using which relies on this input format.
2) As long as the comments occur only at the beginning of the text file, it would be trivial to delete them manually, where strictly necessary.

tdealtry:
Closed via #277

PR #239, opened 2018-04-06
Added new multi-vertex event generator using particle gun and nuance-…

tz-m:
In order to simulate certain types of events, it is necessary to generate multiple particles from _different_ vertices. In particular, pileup events, SN bursts, calibration light sources, etc. I think this is not possible using the current generator action. So, I've added the functionality which creates a new vertex for every track in the nuance-formatted input file using the particle gun, allowing multiple vertices in each event.

First tests show this works as expected, but I'm not exactly sure how the change propagates errors to other parts of the simulation/analysis, if at all.

This should also work for single-vertex, multiple-track events, as muline appears to be designed for.

### DIFF
--- a/include/WCSimPrimaryGeneratorAction.hh
+++ b/include/WCSimPrimaryGeneratorAction.hh
@@ -74,6 +74,7 @@ private:
   G4bool   useGunEvt;
   G4bool   useLaserEvt;  //T. Akiri: Laser flag
   G4bool   useGPSEvt;
+  G4bool   useMultiVertEvt; // Thiesse: use multi-vertex per event
   std::fstream inputFile;
   G4String vectorFileName;
   G4bool   GenerateVertexInRock;
@@ -110,6 +111,9 @@ public:
 
   inline void SetGPSEvtGenerator(G4bool choice) { useGPSEvt = choice; }
   inline G4bool IsUsingGPSEvtGenerator()  { return useGPSEvt; }
+
+  inline void SetMultiVertEvtGenerator(G4bool choice) { useMultiVertEvt = choice; }
+  inline G4bool IsUsingMultiVertEvtGenerator() { return useMultiVertEvt; }
 
   inline void OpenVectorFile(G4String fileName) 
   {
--- a/src/WCSimPrimaryGeneratorAction.cc
+++ b/src/WCSimPrimaryGeneratorAction.cc
@@ -8,6 +8,7 @@
 #include "G4ParticleTable.hh"
 #include "G4IonTable.hh"
 #include "G4ParticleDefinition.hh"
+#include "G4OpticalPhoton.hh"
 #include "G4ThreeVector.hh"
 #include "globals.hh"
 #include "Randomize.hh"
@@ -76,6 +77,7 @@ WCSimPrimaryGeneratorAction::WCSimPrimaryGeneratorAction(
   useGunEvt    = false;
   useLaserEvt  = false;
   useGPSEvt    = false;
+  useMultiVertEvt = false;
 }
 
 WCSimPrimaryGeneratorAction::~WCSimPrimaryGeneratorAction()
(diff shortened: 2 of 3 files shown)


tdealtry:
Thanks for this Matt this will be very useful - I'll be making use of it soon!
A few comments:
- You say it works as expected. Can you show some plots? It's important to verify both that the true track/vertex information is correct, and that there are hits recorded for all true vertices (easy way would be to put a second vertex at a very late time)
- You say that this should also work for traditional muline files. Can you compare events put through "muline" and "multivert" to see if they are indeed identical? (This will then give the option to just extend muline rather than have both)
- You've got a link to http://neutrino.phy.duke.edu/nuance-format/ explaining the traditional nuance format, but you've extended the format. This means that we need a record of the extended nuance somewhere (especially as there are other extension suggestions, e.g. #234). Perhaps a page on the WCSim GitHub wiki @eosullivan?
- You've not actually described the syntax for the nuance extension anywhere. For starters, can you comment here?

eosullivan:
Can you elaborate on how this is different from the recently merged PR #228 ? Although the coding strategies look different, the objectives look the same.

tz-m:
As far as I could tell (I'm relatively new to nuance), the documentation linked does not restrict the numbers of vertices or tracks in any event. So, while the example provided has one vertex, several incoming particles, and several outgoing particles, I don't see any reason why one couldn't write an event along the lines of:

begin
nuance
vertex
track
vertex
track
...
end
stop

It's all how you define the word "Event". For some cases, it means a single neutrino-nucleus interaction, but for others (the examples I mentioned above) it could mean something totally different.

Anyway, I could be wrong, but I don't see how this event definition goes against "proper" nuance, though it may be unconventional or contrary to the original intention. Doesn't make it _wrong_...

More comments on the way...

tz-m:
To test muline vs. multivert, I used the example event from the nuance documentation website. Both give the same output number of hits, but different absolute hit times. I've dumped Ncherenkovdigihits, Nvtxs, Vtxvol, Ntrack, SumQ, track energies, truetimes, and parent track ids. I have no reason to believe any other variables will  All fields are exactly the same between the two methods, apart from the truetime. This is because my code actually implements the nuance-specified vertex time (in the example it's ~5e4 ns), and the original code doesn't.

**multivert**
multivert

**muline**
muline

Muline does not work on a multivertex event as I've specified in my previous comments here.

tz-m:
Lastly, I created a simple event with two 10 GeV electrons in the same event. The first at time 0, and the second at 1e5 ns. Here is a histogram of the hit times (dark noise is turned on and the dark window is very very large). As expected, both particles are seen in the same event. Running the same input file into muline gives no hits.

```
begin
nuance 93
vertex 0 0 0 0 
track 11 1e4 1 0 0 0
vertex 0 0 0 1e5
track 11 1e4 1 0 0 0
end
stop
```

htimes

tz-m:
Oh, and here's the output of the verification script.
verification_plots

eosullivan:
There is already [the laser](https://github.com/WCSim/WCSim/blob/a4571d6dee2a15b2ced1240210f00e4e0db65eab/WCSim.mac#L140) option in WCSIm. Is the laser option sufficient for what you are doing, or is this somehow different?

tz-m:
The laser option uses the GeneralParticleSource. The code proposed here uses the particle gun. Before you brought up the GPS and I spent some time looking through the documentation, I didn't know about the "/gps/source/multiplevertex true" option. I've since tested it and have confirmed that it is possible to somewhat reproduce what I had originally intended. There are a few differences, though. 

First, the GPS input file size is more than double the nuance file size per event since there are more fluff characters. 

Second, the outputs are different and I can't figure out why. I took the same ~9000 input particles in one of my pre-simulation files, formatted the same data into both multivertex nuance, and GPS, and passed both into WCSim. The verification plots show similar, but not identical output. Yes, it's possible there's still a bug in my multivertex code above that means something isn't exactly correct, but if so then I'd expect muline to suffer from the same thing since I borrowed a lot from that bit of code.
gps_multivert_compare

Finally, the input file formats (nuance vs. GPS) are different and provide different benefits. For example, you get mother/daughter/target particle information with nuance, but not with GPS. Obviously you can have parameterised sources of particles with GPS, but not with nuance.

Personally, I'm happier with using nuance because of the smaller input file size. Possibly other groups might prefer it as well.

tz-m:
I've updated the code a little. It now has better documentation of the modified nuance format, and allows comments, see #234. Also, some other minor changes to the loop structure.

tdealtry:
@AlexFinch would you mind taking a look at this, and see if there's anything that #277 hasn't brought? From what I can tell
- This allows the nuance file to not include target/info lines. I'm a bit on the fence about this. In favour, is that there are probably use cases where it'll reduce .kin filesize of large inputs (e.g. light source calibration where 1000s opticalphotons are created in the .kin per event). On the other hand, we don't want to run a large-scale NEUT/sntools/other-event-generator MC production, only to realise that the inputs weren't formatted correctly so we've lost the truth - enforcing the target/info lines to exist will mitigate this to some extent
- It allows optical photons to be simulated (with unused PDG code 30). According to [Geant4.10.1](https://github.com/Geant4/geant4/blob/geant4-10.1-release/source/particles/bosons/src/G4OpticalPhoton.cc#L72) opticalphotons have an PDG code 0 (but using this in the kin file creates an exception...). So this would be a nice feature to add
  - It seems like this has been fixed in [4.10.7](https://github.com/Geant4/geant4/blob/geant4-10.7-release/source/particles/bosons/src/G4OpticalPhoton.cc#L72) in that the PDG code becomes -22 (which is an "anti-gamma" seems like an odd choice they made...)

PR #242, opened 2018-10-16
Build the WCSimWCPMT and WCSimWCAddDarkNoise instances for the OD no …

tdealtry:
…matter what.

Where they are currently built in the constructor of WCSimEventAction, the WCSim.mac file hasn't been read, therefore even if you're planning to use a geometry with an OD, GetIsODConstructed() will always be false since the default geometry is SuperK (without an OD). 

Currently, WCSim.mac fails because `/DarkRate/SetDetectorElement OD` is not a valid. Fails also when using
```
/WCSim/WCgeom HyperKWithOD
/WCSim/Construct
```
A better fix is much more involved - I'll create an issue

### DIFF
--- a/src/WCSimEventAction.cc
+++ b/src/WCSimEventAction.cc
@@ -82,16 +82,12 @@ WCSimEventAction::WCSimEventAction(WCSimRunAction* myRun,
   ///// -- OD -- /////
   ////////////////////
   WCSimWCPMT* WCDMPMT_OD;
-  if(detectorConstructor->GetIsODConstructed()) {
-    WCDMPMT_OD = new WCSimWCPMT( "WCReadoutPMT_OD", myDetector, "OD");
-    DMman->AddNewModule(WCDMPMT_OD);
-  }
+  WCDMPMT_OD = new WCSimWCPMT( "WCReadoutPMT_OD", myDetector, "OD");
+  DMman->AddNewModule(WCDMPMT_OD);
 
   WCSimWCAddDarkNoise* WCDNM_OD;
-  if(detectorConstructor->GetIsODConstructed()) {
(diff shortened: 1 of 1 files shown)


tdealtry:
I'm not sure what's going on with Travis here. According to the [Travis link](https://travis-ci.org/WCSim/WCSim/builds/442222455?utm_source=github_status&utm_medium=notification) everything is green. According to the [WCSim validation plots page](https://wcsim.github.io/Validation/) everything is green. So don't know why it's reporting pending here on github...

eosullivan:
So, what will happen if no OD is constructed? Will creating these WCDMPMT_OD and WCDNM_OD objects cause problems? What does this mean for the code now? That the dark noise is not properly being set for the OD?

tdealtry:
In the WCSimEventAction constructor, the geometry isn't known.
* By default (i.e. before a geometry has been made) `GetIsODConstructed()` returns `false`.
* Therefore if you're wanting to use an OD geometry, it won't work, because the `WCSimWCPMT` and `WCSimWCAddDarkNoise` instances are never created

This pull request fixes this issue
* This is a simple fix to allow both WCSim.mac to work, and to be able to use geometries with ODs.
* The more in depth fix is a rather fiddly modification of the dark rate messenger (similar to the DAQ messenger)

Everywhere else in WCSimEventAction, `GetIsODConstructed()` is called (and returns the right answer because the geometry has been set by this point)
* When you're using a non-OD geometry, the `WCSimWCPMT` and `WCSimWCAddDarkNoise` instances are never touched again
* When you're using an OD geometry, the `WCSimWCPMT` and `WCSimWCAddDarkNoise` instances are there for when they're required

P3tru:
Thank you for taking care of this Tom! I don't know what Travis is doing right now, but I agree this has to be merged in priority.

Issue #245, opened 2018-10-19
NDigits2 Trigger does not handle digit removal properly


marc1uk:
The Ndigits2 trigger removes digits from the digits collection in order to avoid duplication when then applying the second level NDigits trigger scan.
However, digit times in the digits collection are stored in a map of digit number vs time. 
When digits are scanned over in WCSimWCTriggerBase::AlgNDigits and WCSimWCTrigger::FillDigitsCollection, digit times are retrieved with a key that is the iteration number of a  simple for loop. This assumes keys follow a monotonic increase from 0 to ndigits. However, when WCSimWCDigi::RemoveDigitizedGate is called it simply erases the digit from the map, breaking this assumption. 

At present the bug goes unnoticed due to the unsafe c-style map addressing in WCSimWCDigi::GetTime (and many others). Replacing this with a safe `map::at` style addressing and placing `try{ } catch (... )` statements around the digit time retrieval demonstrates the problem. (apply the attached patches to `include/WCSimWCDigi.hh` and `src/WCSimWCTrigger.cc`

[trigpatch.txt](https://github.com/WCSim/WCSim/files/2494634/trigpatch.txt)
[digipatch.txt](https://github.com/WCSim/WCSim/files/2494635/digipatch.txt)

tdealtry:
Good spot! Thanks for this.
Could you put a pull request in for a fix for this? I guess it means changing the loops over  in WCSimWCTriggerBase::AlgNDigits() and WCSimWCTriggerBase::FillDigitsCollection():
`for ( G4int ip = 0 ; ip < (*WCDCPMT)[i]->GetTotalPe() ; ip++)`
to loops over the e.g. times map
Swapping from `[]` to `map::at()` in the times & other maps/vectors in WCSimWCDigi would also be a good plan

marc1uk:
Sure, i'm working on a fix and should have something to submit soon.

marc1uk:
Pull requested created: https://github.com/WCSim/WCSim/pull/247
For some reason this link gives me a 404 error, but i can't re-submit the PR because github tells me one is already open. Hopefully it will become visible soon.

tdealtry:
Closed via #247

PR #247, opened 2018-10-22
Fix for issue #245, support removal of digits in NDigits2 trigger.

marc1uk:
Loops in WCSimWCTrigger::AlgNDigits and WCSimWCTrigger::FillDigitsCollection now use an iterator to loop over the times map and retrieve the keys present.
In WCSimWCDigi.hh several uses of the operator[] were replaced with map::at.
fDigiComp was removed from the sort in WCSimWCDigi::SortArrayByHitTime as this has no entries at the time when it is called (so switching use to fDigiComp.at() produces an exception). 
#245


include/WCSimWCDigi.hh
```diff
@@ -139,29 +160,32 @@ public:
   void SortArrayByHitTime() {
```
tdealtry: I agree that where `SortArrayByHitTime()` is called currently, we don't need to sort `fDigiComp`. But maybe we want to future proof ourselves for potentially using this sorting method elsewhere in the future by keeping it in. And maybe making the method name more descriptive such as:
```suggestion
  void SortAllMapsByHitTime() {
```
Alternatively, we could do as you suggest and get rid of `fDigiComp` sorting here, and rename to
```suggestion
  void SortMapsByHitTimeForDigitizer() {
```

tdealtry: I agree that where `SortArrayByHitTime()` is called currently, we don't need to sort `fDigiComp`. But maybe we want to future proof ourselves for potentially using this sorting method elsewhere in the future by keeping it in. And maybe making the method name more descriptive such as:
```suggestion
  void SortAllMapsByHitTime() {
```
Alternatively, we could do as you suggest and get rid of `fDigiComp` sorting here, and rename to
```suggestion
  void SortMapsByHitTimeForDigitizer() {
```

include/WCSimWCDigi.hh
```diff
@@ -91,21 +91,42 @@ public:
...
-  inline G4float GetPreSmearTime(int gate)   {return time_presmear[gate];};
+  inline G4float GetPe(int gate)     {return pe.at(gate);};
+  inline G4float GetTime(int gate)   {
+    try {
+      return time.at(gate);
+    }
+    catch (...) {
+      G4cerr<<"Exception occurred while attempting to use WCSimWCDigi::GetTime to retrieve time for pe "
+            << gate << " from map of times. The time map has "<<time.size()<<" entries:" << G4endl;
+      for (auto& x: time){
```
tdealtry: @eosullivan can comment, but I'm not sure WCSim is using C++11 yet

tdealtry: @eosullivan can comment, but I'm not sure WCSim is using C++11 yet

tdealtry:
The Travis build hasn't started for this. Any ideas @brichards64 or @eosullivan?

marc1uk:
Thanks for the feedback. Pushed a new commit to remove c++11 and re-enable sorting of the fDigiComp map, provided it's size is the same as the other maps.

tdealtry:
Great! Thanks for this @marc1uk!

PR #254, opened 2018-11-09
Add an equality method for WCSimRootGeom & WCSimRootPMT

tdealtry:
In a similar vein to #166 but for geometries


eosullivan:
Can you resolve this conflict caused by a recent merge?

tdealtry:
Just a quick explanation here. ComparisonPassed() was defined in the .cc files of both WCSimRootEvent & WCSimRootGeom. This obviously wouldn't compile (same function declared twice). So I moved it into new WCSimRootTools .cc & .hh files

In WCSimRootTools there are template versions of the ComparisonPassed() methods. But these are commented out due to them not correctly linking. If anyone can work out how to correctly add them to the LinkDef.h file, then please do! I'll create an issue for it.

PR #262, opened 2018-11-23
Some changes to give better agreement with TriggerApplication

tdealtry:
* Use `double` instead of `float` to calculate sumQ to fill the trigger header
* Change `WCSimWCTriggerBase::LongTime` from 1 to 10 ms and make the NoTrigger trigger readout from `-WCSimWCTriggerBase::LongTime` instead of 0, in order to save absolutely every digit - which is the purpose of this trigger. (NoTrigger dropping negative digits was one of the things I've found was going wrong during my validation since Wednesday night)
* In `AlgNDigits()`, have a loop over all digits to calculate the first & last trigger decision window times to use (rather than starting at 0, which may not be the first digit, or may be way too early)
* Don't apply the 950 ns offset (`WCSimWCTriggerBase::offset`) when using the NoTrigger trigger. Doing this (then reverting in TriggerApplication) was causing all digits to be different (at a low level ~1E-4 ns) due to floating point precision issues.
* Add a protection that means `multiDigitsPerTrigger` set to `false` won't cause digits to be dropped for the NoTrigger trigger

### DIFF
--- a/src/WCSimWCTrigger.cc
+++ b/src/WCSimWCTrigger.cc
@@ -409,10 +409,9 @@ void WCSimWCTriggerBase::FillDigitsCollection(WCSimWCDigitsCollection* WCDCPMT, 
 
 	  //first apply time offsets
 	  float peSmeared = (*WCDCPMT)[i]->GetPe(ip);
-	  G4double digihittime = -triggertime
-	    + digit_time;
+	  G4double digihittime = digit_time;
 	  if(triggertype != kTriggerNoTrig)
-	    digihittime += WCSimWCTriggerBase::offset;
+	    digihittime += WCSimWCTriggerBase::offset - triggertime;
 
 	  //get the composition information for the triggered digit
 	  std::vector<int> triggered_composition = (*WCDCPMT)[i]->GetDigiCompositionInfo(ip);


src/WCSimWCTrigger.cc
```diff
@@ -32,7 +32,7 @@
 #endif
 
 const double WCSimWCTriggerBase::offset = 950.0; // ns. apply offset to the digit time
-const double WCSimWCTriggerBase::LongTime = 1E6; // ns = 1ms. event time
```
eosullivan: Is this reasonable to change the LongTime variable to 10ms? This is the value that will be in the actual trigger?

tdealtry: `LongTime` is used in two places
* `GetPostTriggerWindow()` (and, as of this PR, `GetPreTriggerWindow()` only for the NoTrigger trigger.
* Setting the end (and, as of this PR, start) times of the loop in `AlgNDigits()`, before the start/end time have been found by looping over all digits.

Therefore the effect of `LongTime` doesn't change the trigger, other than making it more robust in instances of
* long time windows
* not setting the time window simulated to start at 0 (particle gun & muline defaults)

I don't remember exactly why I choose 10 times more (but recall the neutron capture lifetime on H is about 0.2 ms). An even greater range would be better.

PR #264, opened 2018-12-20
Add how to run WCSim to the readme

tdealtry:
I've just realised that the only place that how to run WCSim is at https://github.com/WCSim/WCSim/wiki/Tutorial:-How-to-set-up-GitHub-(in-the-context-of-WCSim) This is quite an unexpected place for new people to find the information! So I've added a little section to the README.md

Additionally I've cleaned up the formatting a bit

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,9 +113,11 @@ Extra docker commands:
 
 ## Running WCSim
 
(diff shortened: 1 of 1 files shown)


eosullivan:
Good catch, @tdealtry!

PR #266, opened 2019-01-18
Allow uniform time/energy distributions in MakeKin.py

tdealtry:
This pull request
* Sets the time of events to that of the `vertex` line in the nuance format
* Expands #265 to save all particle energy/pdg/direction/vertex positions, in case where there are multiple
* Expands MakeKin.py 
  * Allow uniform range of energies
  * Add time option, can be fixed or uniform range
  * Add random seed option
  * Add HyperK as an option for the detector


tdealtry:
Renaming because the reading times has been superceeded by #277 

So this now just expands MakeKin.py
- Allow uniform range of energies
- Add time option, can be fixed or uniform range

PR #271, opened 2019-02-23
bug fix:  Make daq.mac PreTriggerWindows take effect

marc1uk:
Currently the `/DAQ/*Trigger/PreTriggerWindow` options have no effect, because in the `WCSimWCDAQMessenger.cc` file the setting is only transferred from the messenger internal variable to the Trigger class if the `PreTriggerWindow >= 0` is satisfied. But the PreTriggerWindow is defined as being negative, so this condition is never met.
This PR initializes the messenger internal variables with -1E6 and then changes the condition for transferring this to the Trigger class to be `PreTriggerWindow > -1E6`

### DIFF
--- a/include/WCSimWCDAQMessenger.hh
+++ b/include/WCSimWCDAQMessenger.hh
@@ -72,7 +72,7 @@ private:
   G4bool                StoreNDigitsAdjustForNoise;
   G4UIcmdWithAnInteger* NDigitsPreTriggerWindow;
   G4int                 StoreNDigitsPreWindow;
-  G4bool                NDigitsPreWindowSetByUser; //because this is typically negative, a foolproof workaround
+  G4bool                NDigitsPreWindowSetByUser; //because this is typically negative, a foolproof(?) workaround
   G4UIcmdWithAnInteger* NDigitsPostTriggerWindow;
   G4int                 StoreNDigitsPostWindow;
 
--- a/include/WCSimWCDAQMessenger.hh
+++ b/include/WCSimWCDAQMessenger.hh
@@ -72,6 +72,7 @@ private:
   G4bool                StoreNDigitsAdjustForNoise;
   G4UIcmdWithAnInteger* NDigitsPreTriggerWindow;
   G4int                 StoreNDigitsPreWindow;
+  G4bool                NDigitsPreWindowSetByUser; //because this is typically negative, a foolproof workaround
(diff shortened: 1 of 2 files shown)


tdealtry:
Good spot!
Personally I'd tackle the fix in a slightly different way, that has maybe less confusing logic. 
`WCSimWCTrigger::SetNDigitsPreTriggerWindow()` and `SetSaveFailuresPreTriggerWindow()` (and the post trigger equivalents) force the input to be the absolute value. So let's enforce this in the messenger too.
This would have the added benefit of the messenger and trigger class printouts not be inconsistent (and therefore less confusing for the user)
What do you think?

marc1uk:
The trouble with using signs to determine if the user has made a call to `/DAQ/*Trigger/SetPreTriggerWindow` is that the resulting behaviour depends on how the user is expecting signs to be interpreted (or not).
If the user knows the sign of PreTriggerWindows is negative, they may try to use:
`/DAQ/TriggerNDigits/PreTriggerWindow-200`, but then if the messenger checks if the user has set a value by testing `StoreNDigitsPreWindow >= 0`, the messenger will deduce the setter wasn't called. 
Similarly the vice versa can happen if we check for `StoreNDigitsPreWindow < 0` and the user isn't aware the value should be negative. Casting away the sign entirely then makes the check pointless, as it's always satisfied.

marc1uk:
It could also be confusing if the user calls `/DAQ/TriggerNDigits/PreTriggerWindow 200` and the software responds `Setting PreTriggerWindow to -200`, or even worse, reports different signs at different stages. If the intention is to avoid confusion, that seems much worse.

tdealtry:
Well this has been a while... sorry for that

I've realised that this is actually two different issues, so I've modified your PR to address them in different ways
- The `SaveFailuresPreWindow` doesn't have trigger-class specific defaults, so doesn't have to be checked to see if it's different from an "out-of-normal-range" default value
  - likewise I've changed it for `SaveFailuresPostWindow` too)
- The `NDigitsPreWindow` can been tagged with a boolean that it's been changed from `SetNewValue()` (rather than checking an aribtary `-99` or `-1E6`) which should be robust
  - Technically all the other variables that get defaults from the trigger/digitizer classes could be handled this way, but since they're all unambigiously positive, I've not done this since checking for `>= 0` works

I've tested that it works for a variety of positive & negative values in the mac file, and it seems robust. @marc1uk if you're still using WCSim, please do take a look. Otherwise I'll merge in a few days

PR #272, opened 2019-12-03
A more direct comparison of digits

tdealtry:
When using TriggerApplication, the order of the digits can be different to the order of the digits in WCSim. This causes a simple side-by-side comparison of digits to fail.
This PR does a digit-by-digit comparison of all digits on the same PMT, until the common digit is found.


eosullivan:
Does WCSim need to change? Is it not possible to change the trigger application to adapt this?

tdealtry:
It's been about 18 months since I wrote it, so can't offer a detailed response, but essentially why a difference is happening is
* When running WCSim with the `NoTrigger` trigger, all digits on the same PMT are saved together in the `TClonesArray`
* When running WCSim with the `NDigits `trigger, all digits on the same PMT in the same trigger are saved together in the `TClonesArray`, but other digits on that PMT in a later trigger are saved in a different part of the `TClonesArray`
* The way TriggerApplication works is looping over all digits in the order it gets them (i.e. from a WCSim file that has the `NoTrigger` trigger applied). It then
  * Removes digits from the "0th trigger `TClonesArray`" that aren't in the 0th trigger (this is why there are holes in the `TClonesArray` and the purpose of the `_slots` variations of variables/methods that were in previous PRs you merged). The order of digits here are identical between WCSim and TriggerApplication
  * Creates new "subsequent trigger `TClonesArray`s". It sees digits in the order it received them from WCSims `NoTrigger`, which is different to how WCSim deals with them directly when using its `NDigits`. Hence some divergence in the ordering

So, in short, if we ever want the option to compare WCSim trigger & TriggerApplication trigger output (we probably do, unless WCSim drops triggering) then we need this `deep_comparison` option (which is turned off by default)

PR #275, opened 2020-02-06
Radon v1.8.0 g4.10.03p03

gpronost:
Add radon BG model for Low energy analysis


WCSim.mac
```diff
@@ -52,7 +53,7 @@
 #/WCSim/LCConfig Mirror_OldLC # Mirror design in a branch of archive/LightCone at 2011.
 #/WCSim/LCConfig Mirror_2018Oct # A short mirror design.
 
-#/WCSim/Construct # This must be uncommented in order to construct a new detector configuration.
```
eosullivan: Please put back all the defaults here in WCSim.mac and just add in the new part you need.

make.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
eosullivan: And remove these new make files? Unless they are necessary for this pull request.

src/WCSimDetectorConfigs.cc
```diff
@@ -219,8 +219,11 @@ void WCSimDetectorConstruction::SetHyperKGeometry()
   WCPMTName           = PMT->GetPMTName();
   WCPMTExposeHeight   = PMT->GetExposeHeight();
   WCPMTRadius         = PMT->GetRadius();
-  WCIDDiameter          = 70.8*m; // = 74m - 2*(60cm ID wall + 1m OD)
-  WCIDHeight            = 54.8*m; // = 60m - 2*(60cm ID wall + 2m OD)
+  //WCIDDiameter          = 70.8*m; // = 74m - 2*(60cm ID wall + 1m OD)
+  //WCIDHeight            = 54.8*m; // = 60m - 2*(60cm ID wall + 2m OD)
```
eosullivan: Put back the default value. (yes, we need to still put in the new dimensions, but this will be separate)

src/WCSimEventAction.cc
```diff
@@ -227,10 +227,10 @@ void WCSimEventAction::EndOfEventAction(const G4Event* evt)
     G4String name =   WCIDCollectionName;
     G4int collectionID = SDman->GetCollectionID(name);
     if(collectionID>-1) WCHC = (WCSimWCHitsCollection*)HCE->GetHC(collectionID);
-    G4cout << G4endl;
-    G4cout << "WCSimEventAction::EndOfEventAction ☆ (WCSimWCHitsCollection*)" << WCIDCollectionName
-           << " has " << WCHC->entries() << " entries" << G4endl;
-    G4cout << G4endl;
+    //G4cout << G4endl;
```
eosullivan: Is this part of the code necessary for your radioactivity simulation? If not, restore default.

src/WCSimEventAction.cc
```diff
@@ -352,10 +352,10 @@ void WCSimEventAction::EndOfEventAction(const G4Event* evt)
...
       G4int collectionID = SDman->GetCollectionID(name);
       if(collectionID>-1) WCHC_OD = (WCSimWCHitsCollection*)HCE->GetHC(collectionID);
-      G4cout << G4endl;
-      G4cout<< "WCSimEventAction::EndOfEventAction() (WCSimWCHitsCollection*)" << WCODCollectionName
-            << " has " << WCHC_OD->entries() << " entries" << G4endl;
-      G4cout << G4endl;
+      //G4cout << G4endl;
+      //G4cout<< "WCSimEventAction::EndOfEventAction() (WCSimWCHitsCollection*)" << WCODCollectionName
+      //      << " has " << WCHC_OD->entries() << " entries" << G4endl;
+      //G4cout << G4endl;
```
eosullivan: Again, here.

src/WCSimEventAction.cc
```diff
@@ -379,6 +379,7 @@ void WCSimEventAction::EndOfEventAction(const G4Event* evt)
     WCDChitsID_OD = DMman->GetDigiCollectionID("WCRawPMTSignalCollection_OD");
     WCDC_hits_OD = (WCSimWCDigitsCollection*) DMman->GetDigiCollection(WCDChitsID_OD);
     // printouts
+    /*
```
eosullivan: Here

src/WCSimEventAction.cc
```diff
@@ -387,10 +388,11 @@ void WCSimEventAction::EndOfEventAction(const G4Event* evt)
       G4cout << "no";
     }
     G4cout << " entries" << G4endl;
-
+    */
```
eosullivan: And here.

src/WCSimEventAction.cc
```diff
@@ -626,7 +629,7 @@ void WCSimEventAction::EndOfEventAction(const G4Event* evt)
   }
 
   //fill correct variables for track from decay
-   G4cout << " Filling Root Event " << G4endl;
+   //G4cout << " Filling Root Event " << G4endl;
```
eosullivan: Ya, just go through and remove these changes you made that aren't part of the radioactivity simulation.

make.sh
gpronost: Actually, with Geant4 radioative decay doesn't properly compile with the "standard compilation" way. It's quite important to use CMake. For this purpose make.sh and make_sukap.sh are here to show how to do it. I can remove them if you prefer.

WCSim.mac
gpronost: Ok, I did it.

src/WCSimDetectorConfigs.cc
gpronost: Done

src/WCSimEventAction.cc
gpronost: Done, I kept only the modification needed for relative PMT hit time (which is optionnal via DAQ messenger) and about the overwrite of the tree.

src/WCSimEventAction.cc
gpronost: Done

gpronost:
I followed your comments, there are just two remaining "issue" I would like you to confirm if I need to modify or not.

eosullivan:
@tdealtry has pointed out that this pull request requires a different version of GEANT than our current standard, requiring us to update the recommended GEANT version. @gpronost is it necessary to have 4.10.03 for this pull request? Can you still implement this Rn generator with 4.10.1p03 instead or is there something specific you need in the later version?

gpronost:
The Rn generator can be used with 4.10.1p03, however since the radioactivity decay model was updated in 4.10.03 the results will not be exactly the same (I am not sure if it will have any impact on the WC simulation). 

There is just one line in the code which needs to be changed to be used with older version, I will comment in @tdealtry thread.

tdealtry:
@federiconova showed a couple of years ago that there is a bug in Geant4 particle decay, when there are multiple gammas released (the summed energy is wrong). I think this was present in 4.10.1, and fixed in a beta of 4.10.3. But my memory is hazy, so perhaps you can comment Federico?
So it would be nice to update the recommended version of Geant4 to at least 4.10.3. Of course there are potentially other changes in the models, so it's not trivial

gpronost:
> So it would be nice to update the recommended version of Geant4 to at least 4.10.3. Of course there are potentially other changes in the models, so it's not trivial

The nuclear state energies are considerably improved in 4.10.3 compared to 4.10.1, so I agree updating the 4.10.3 would be better from this point of view.

PR #276, opened 2020-02-18
Update README.md

tdealtry:
PR #275 made some changes that causes builds with G4.10.1 to fail. Updating the README to suggest a more up-to-date G4 version (not sure what the minimum version is, just using the version suggested by #275)

e.g. of a change: https://github.com/WCSim/WCSim/pull/275/files#diff-0cfe8dcf71fdd92dedfef801c5fd0751R145

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,9 +36,11 @@ More detailed information about the simulation is available in
 ## Current notes and how to build
 
 You should have a recent and working version of ROOT and GEANT4.
-(Known to work with GEANT 4.10.1p03 and ROOT v5.28.00)  You also need all of the G4
+(Known to work with GEANT 4.10.03p03 and ROOT v5.28.00)  You also need all of the G4
 data files including hadron xsecs etc.  Those are the only
-requirements.  The code should work with gcc 4.4.7. For v1.6.0 and earlier, use GEANT 4.9.4.p01.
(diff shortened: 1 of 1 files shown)


gpronost:
G4.10.1 can still be used, but indeed there is one issue with the compilation:

In G4.10.03, the function PrintAvailableGraphicsSystems (in WCSimVisManager.cc) needs an argument, whereas in G4.10.1 it can't have an argument. 

Unfortunately, I don't know if we can access to the G4 version from the compilateur, so I don't know if we can make WCSimVisManager.cc compatible with both versions. If PrintAvailableGraphicsSystems is not needed, maybe it would be simpler to comment it?

eosullivan:
I'm inclined to leave the version of GEANT as the recommended version if it is not needed to make the Rn pull request work. (It sounds like you just made the change to get WCSim to compile with the GEANT version you have, unless I am misunderstanding?).

gpronost:
> It sounds like you just made the change to get WCSim to compile with the GEANT version you have, unless I am misunderstanding?

Yes, that's correct.

eosullivan:
I will close this for now since we are once again compatible with 4.10.1. When we update the code to a new GEANT version, we can update the README as well.

PR #277, opened 2020-02-19
Feature/multiple vertices in nuance files

AlexFinch:
Modified PrimaryGeneratorAction so that it can read in multiple vertices in a file , for example like this..
$ begin
$ nuance 
$ vertex 1000.18466 2540.68005 -48.00089 74.22065
$ track 14 1600.00000 1.00000 0.00000 0.00000 -1
$ track 2212 935.98400 0.00000 0.00000 1.00000 -1
$ info 0 0 98
$ track 11 600.00000 -0.33171 0.65848 -0.67555 0
$ vertex -464.02342 -3358.18649 2334.91690 84.21316
$ track 14 1600.00000 1.00000 0.00000 0.00000 -1
$ track 2212 935.98400 0.00000 0.00000 1.00000 -1
$ info 0 0 98
$ track 11 600.00000 0.01788 -0.92563 -0.37801 0
$ end


src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -167,6 +167,7 @@ void WCSimPrimaryGeneratorAction::GeneratePrimaries(G4Event* anEvent)
 	    vtxs[iVertex] = G4ThreeVector(atof(token[1])*cm,
 				    atof(token[2])*cm,
 				    atof(token[3])*cm);
+	    G4double VertexTime=atof(token[4]);
```
ast0815: Doesn't this miss a time unit? I think the vector files I get from super nova simulations with sntools use milliseconds.

AlexFinch:
I can't find anything definitive as the units that time is defined in. I've set it to seconds for now.

tdealtry:
The default units in WCSim are ns. It's what WCSim saves things as, and it's what Geant assumes too (e.g. if you're using the /gun/time command https://geant4-userdoc.web.cern.ch/geant4-userdoc/UsersGuides/ForApplicationDeveloper/BackupVersions/V10.1/html/AllResources/Control/UIcommands/_gun_.html)
So I think ns should be the default.

However Lukas makes a good point that input software may use different conventions. So maybe the time unit should be a .mac file option?


A few of other suggestions
- Since you're playing with the muline input, perhaps you could implement comments to close #234 (i.e. just ignore lines starting with `#`)
- The extended nuance format should be documented somewhere. Maybe in README.md? http://neutrino.phy.duke.edu/nuance-format/ will help you get started (personally I think it's a good idea to bring a copy of that documentation off the Dukes website anyway, since, given time, links break...)
- It'd be nice for MakeKin.py to be extended to handle:
  - Multiple vertices per event (maybe have 2 modes. 1 with fixed number of vertices per event, 1 drawing random numbers of vertices from a Poisson)
  - A time option (again maybe a couple of options. All fixed, drawing from uniform range)

include/WCSimWCDigitizer.hh
```diff
@@ -38,7 +38,10 @@ public:
   void SetDigitizerPEPrecision      (double precision) { DigitizerPEPrecision     = precision; }; ///< Override the default digitizer charge resolution (p.e.)
 
   double Truncate(double value, double precision) {
-    if(precision < 1E-10) return value;
+      if(precision < 1E-10) return value;
+      float ratio=value / precision;
+      // Result of cast is undefined if value is outside range of int. Just do nothing in that case.
+      if(ratio>2147483647 || ratio<-2147483648) return value;
```
tdealtry, line 44: This looks good, but can you combine it with 20c056940603647a0bda8e9f37fbc62fef97ef4c from @ast0815? i.e. use `long int`, and the relevant max allowed values, instead?

src/WCSimWCDigitizer.cc
```diff
@@ -5,7 +5,7 @@
 // for memset
 
 #ifndef NPMTS_VERBOSE
-#define NPMTS_VERBOSE 10
+// #define NPMTS_VERBOSE 10
```
tdealtry, line 8: Not sure why you've turned that off? It's only used in printouts, and only when `WCSIMWCDIGITIZER_VERBOSE` is on

src/WCSimWCDigitizer.cc
```diff
@@ -119,8 +119,13 @@ bool WCSimWCDigitizerBase::AddNewDigit(int tube, int gate, float digihittime, fl
   //digitised hit information does not have infinite precision
   //so need to round the charge and time information
   double digihittime_d = Truncate(digihittime, DigitizerTimingPrecision);
+
   double peSmeared_d   = Truncate(peSmeared,   DigitizerPEPrecision);
 
+//  double Truncate(double value, double precision) {
+//    if(precision < 1E-10) return value;
+//    return precision * (int)(value / precision);
+//  }
```
tdealtry, line 128: Think you've accidentally committed something you didn't mean to!

AlexFinch:
I've added a .mac command /mygen/time_unit me_unit so that the time units can be set. By default it is ns.
I've also done a couple of unrelated things.
I've allowed lines in a vertex file to be commented out by starting with \# as requested in #234 and I've protected Truncate against integers overflowing and wrapping back to  negative number.

tdealtry:
Great thanks @AlexFinch! But it looks like 5107ded27575ef5ca00ee71613cf3e4b020ea74b is missing some features - probably forgot to `git add` or something

Nuance_MC_Format.txt
```diff
@@ -0,0 +1,74 @@
...
+    nuance: 
+
+      This line indicates the reaction mode for the event.
+
+    vertex: 
+
+       The first three numbers are the x, y and z coordinate, in cm. The
+    fourth number is the time, in simulation units normally ns.  To specify
+    different units use the command /mygen/time_units in the .mac file.
+
```
tdealtry, line 37: I think it's good to be specific about how to write files with multiple vertices
```suggestion
Multiple vertices can used in a single event by repeating the vertex, track, and info lines; the nuance line should not be repeated.
```

include/WCSimWCDigitizer.hh
AlexFinch: In fact I'll go further! I'll add this:

 ```
if(ratio>std::numeric_limits<long long>::max() || ratio<std::numeric_limits<long long>::min() ) return value;
    return precision * (long long)(value / precision)
```

File: doc/Nuance_MC_Format.txt [hunk 6]
```diff
@@ -33,7 +33,9 @@ Each line begins with a text tag, indicating the information it contains.
 
        The first three numbers are the x, y and z coordinate, in cm. The
     fourth number is the time, in simulation units normally ns.  To specify
-    different units use the command /mygen/time_units in the .mac file.
+    different units use the command /mygen/time_units in the .mac file.Multiple
```
tdealtry, line 36: pedantic comment but... can you add a space?

ast0815:
`long long`s are only available from C++11 onwards, I think. So that might break some compatibility somewhere. Unless WCSim is intended as C++11 and later anyway. I have not found any standard specifications in the Makefiles.

WCSim.mac
```diff
@@ -132,6 +132,8 @@
 #/mygen/vecfile inputvectorfile
 #/mygen/vecfile h2o.2km.001-009x3_G4.kin
 #/mygen/vecfile mu+.out
+# define units used for time in kin file
+#/mygen/time_unit "ns"
```
tdealtry, line 136: Probably don't need the `"` here. If they are required, then that is confusing since all the other parameters don't need them, and so should be fixed

ast0815:
Using `long int` means that it only works as intended on 64 bit Linux machines though. The best thing would be to use a fixed width type, I guess. E.g. `int64_t` from `<cstdint>`.

src/.gitignore
```diff
@@ -0,0 +1 @@
+WCSimRootDict.cxx
```
tdealtry, line 1: Not sure how you've got a `.cxx`!? The GNUmakefile uses `.cc` Maybe a Root6 thing?
But should be consistent - use the `$WCSIMDIR/.gitignore`, or move these two lines (without `src/`) from that file into this one
```
src/WCSimRootDict.cc
src/WCSimRootDict.h
```

src/WCSimWCTrigger.cc
```diff
@@ -36,10 +36,12 @@ const double WCSimWCTriggerBase::LongTime = 10E6; // ns = 10ms. event time
```
eosullivan, line 36: Are there any changes actually made to WCSimWCTrigger.cc? It looks like this is mostly inserted extra spaces. Can you fix this to just be the new changes required for this pull request?

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -29,681 +29,725 @@ vector<string> tokenize( string separators, string input );
...
+}
 }
 
 inline float atof( const string& s ) {return std::atof( s.c_str() );}
 inline int   atoi( const string& s ) {return std::atoi( s.c_str() );}
 
 WCSimPrimaryGeneratorAction::WCSimPrimaryGeneratorAction(
-					  WCSimDetectorConstruction* myDC)
-  :myDetector(myDC), vectorFileName("")
+    WCSimDetectorConstruction* myDC)
```
eosullivan, line 56: There are also many changes here in WCSimPrimaryGeneratorAction that are just spacing (although here I also see added code). Could this be cleaned up a bit?

AlexFinch:
I've made the changes to remove spurious whitespace. However I am having trouble getting github to show me the changes relative to the base. All I can get is changes relative to the previous commit. Does anyone know how to do this?

eosullivan:
Thanks, Alex! I believe the "files changed" tab https://github.com/WCSim/WCSim/pull/277/files Should include all changes.

AlexFinch:
Oh yes so it does, thanks.

JostMigenda:
After discussing the event mode at the UK DAQ meeting on Tuesday, I think it might be good to have one `$ nuance X` (where X is the reaction mode) line per vertex, instead of one per event?

(Maybe we should discuss this at next week’s UK DAQ meeting before merging this PR, because that would also affect how I write the NUANCE output file in sntools – and potentially other software in the DAQ toolchain that makes use of this reaction mode.)

eosullivan:
Ok, please let me know the outcome of your meeting! I also notice that there are still many whitespace changes in WCSimPrimaryGeneratorAction. Sorry to be a pain about it @AlexFinch but it is nice to keep the commit history clean.

eosullivan:
Actually, @JostMigenda, can you elaborate? Why would it be easier to have one nuance line per vertex, rather than per event?

JostMigenda:
Sure. The basic idea here is to treat a full supernova as one event[*], with every individual neutrino interaction corresponding to one vertex inside that event. Since the reaction mode corresponds to a single neutrino interaction and not to the whole SN, we need a separate `$nuance X` line for each vertex.
We discussed this at yesterday’s UK DAQ meeting and agreed that this will be necessary.

We also discussed how to extend the list of reaction modes to be more useful for SN neutrino interaction. The result is [already implemented in sntools](https://github.com/JostMigenda/sntools/commit/d847b4c03fdb9a861e2600f38fe0e2496966b2b8). It should require no modifications here, since WCSim just reads in the int from the NUANCE file and then writes it out, without interpreting it in the meantime.

[*] other folks from the UK DAQ group can probably explain in more detail than I can why this is the best way to handle a SN in the overall DAQ toolchain, but I’ll skip that digression for now

ast0815:
In short: Because it is the most realistic. Having each neutrino be a separate event works too (with the double precision timing), but it ignores things like pile-up, and potentially double counts or under-estimates noise hits (depending on how those are generated).

AlexFinch:
I believe all the remaining 'whitespace' changes in WCSimPrimaryGeneratorAction.cc are correcting incorrect faulty indentation so should be retained.

AlexFinch:
One way to implement the feature of having a different reaction mode per vertex is to use the currently unused information in the info line. It would be necessary to tell WCSim to treat that value
as a process which could be done either by defining a special value for the nuance line or by an additional command defined in the .mac file. e.g.
/mygen/fileFormat processPerVertex
Then one could use the first digit after info for the process type. Something like this

$ begin
$ nuance <special value or dummy value>
$ vertex 3503.27202 557.01173 363.52404 71.36341
$ track 14 1600.00000 1.00000 0.00000 0.00000 -1
$ track 2212 935.98400 0.00000 0.00000 1.00000 -1
$ info <process type vertex 1> 0 0
$ track -12 600.00000 -0.96561 0.20537 0.15942 0
$ vertex 3581.58013 -694.03763 2469.88853 5.72273
$ track 14 1600.00000 1.00000 0.00000 0.00000 -1
$ track 2212 935.98400 0.00000 0.00000 1.00000 -1
$ info <process type vertex 2> 0 0
$ track -12 600.00000 -0.88655 -0.17768 -0.42715 0
$ vertex 3107.18773 -1859.29403 -1172.59332 49.30521
$ track 14 1600.00000 1.00000 0.00000 0.00000 -1
$ track 2212 935.98400 0.00000 0.00000 1.00000 -1
$ info <process type vertex 3> 0 0
$ track -12 600.00000 0.96788 0.08393 0.23700 0
$ vertex 300.45898 -3419.30816 -2702.82836 59.79686
$ track 14 1600.00000 1.00000 0.00000 0.00000 -1
$ track 2212 935.98400 0.00000 0.00000 1.00000 -1
$ info <process type vertex 4> 0 0
$ track -12 600.00000 0.06470 -0.50836 -0.85871 0
$ vertex 787.48887 -705.88102 1153.35701 33.73082
$ track 14 1600.00000 1.00000 0.00000 0.00000 -1
$ track 2212 935.98400 0.00000 0.00000 1.00000 -1
$ info <process type vertex 5> 0 0
$ track -12 600.00000 0.81199 -0.24435 -0.53006 0
$ end

tdealtry:
Is it much harder to look for multiple `nuance` lines per event than doing this? I just think this method will be more prone to user error
Additionally, sntools has [implemented the code on the nuance line](https://github.com/JostMigenda/sntools/commit/d847b4c03fdb9a861e2600f38fe0e2496966b2b8#diff-03554bf924d627ac3509373e70aca25bR28)

AlexFinch:
Anything is possible to implement, but this seems to me to be the simplest as  it doesn't actually change the existing format, rather it chooses to interpret some currently dummy data in a new way.

JostMigenda:
In the existing format, the `nuance` line contains the reaction mode. Interpreting a different field as the reaction mode would break compatibility with the old format.
What happens if you encounter a file where the `nuance` line has one number and the `info` line has a different number? There is no way for WCSim to determine which of those is the reaction mode and which is just a dummy value.

In contrast, the format with one `nuance` line per vertex is very close to the original format. In fact, it would be possible to completely automatically switch back and forth between the traditional format (one neutrino interaction per event) and the new format (multiple neutrino interactions per event, each with its own vertex) by simply adding/removing the lines
```
$ end
$ begin
```
before each `nuance` line.

tdealtry:
Either way is a change relative to the current format WCSim recognises ;-)

If you do want to use the `info` line, I'd suggest turning it on with the `nuance` line would be best - it requires the software feeding into WCSim to conform to WCSim standards, rather than making sure the user remembers an odd option in the WCSim `.mac` file only when they require it

In terms of a number, I think given that [nuance numbers go up to +99](http://neutrino.phy.duke.edu/nuance-format/) and [NEUT numbers are similar (but can take negative values)](https://t2k.org/nd280/software/neutcodes), and sntools numbers are 6(?) digits, a value of -9999 maybe?

But as Jost says, this is splitting the definition of where the number is and will probably cause problems for someone at some point!

AlexFinch:
To reply to the specific question:
 What happens if you encounter a file where the nuance line has one number and the info line has a different number?

The process type would initially be set to the value in the nuance line, then be replaced by the value in the info line if a command in the mac file instructed it to behave that way.
Otherwise the nuance line value is the process type for everything  and the info line is ignored.
Alternatively if we go with a special value to trigger this behaviour then the value on the info line is the only meaningful value. This way would allow mixing of the two schemes on an event by event basis.

JostMigenda:
I agree with Tom that asking the user to remember to turn on/off a separate option in the .mac file is very error-prone and we should avoid that. Using a special value in the `nuance` line is not quite as error-prone, but I still worry a bit that someone, years from now, might decide to introduce new reaction codes for their own purposes (like we just did for the SN interaction channels!) and run into an issue then.

Even if that weren’t an issue, I still think that splitting the reaction code across two different fields (including one which was intended for some completely different information originally) is unnecessarily complex; it makes the code more complicated and deviates much further from the original NUANCE format than necessary.

The alternative suggestions seems much more elegant to me: Instead of allowing the lines
```
$ vertex …
$ track …
$ track …
$ info …
$ track …
```
to be repeated multiple times within one event, we’d allow the lines
```
$ nuance …
$ vertex …
$ track …
$ track …
$ info …
$ track …
```
to be repeated.
That is a very straightforward change compared to the original (one neutrino interaction per event) format (literally all we’re doing is dropping the `$ begin` and `$ end` lines between neutrino interactions). It doesn’t require us to repurpose any field and introduce a new way to switch between two possible meanings of that field. And after the modifications that @AlexFinch has already made, it should be a very simple change in the parsing code: Instead of the `while(token[0]=="vertex")` in [line 183 of WCSimPrimaryGeneratorAction.cc](https://github.com/WCSim/WCSim/pull/277/files#diff-ee36e2984dff98a9a80d9370f0b349a1R183), have a `while(token[0]=="nuance")` a few lines above.

AlexFinch:
I've implemented the format as requested by Jost. I've also updated MakeKin.py to be able to generated this format if required. Some other tidying up as well.

tdealtry:
Thanks Alex! Looks great!
One last thing though - it seems to me that the interaction modes associated with the nuance line aren't actually passed through to the output file - or more exactly only the last (or first?) is (i.e. there's a `GetMode()` but not a `GetModes()` like there is a `GetVtxs()` in https://github.com/AlexFinch/WCSim/blob/feature/MultipleVerticesInNuanceFiles/include/WCSimRootEvent.hh#L289)

AlexFinch:
Tom, you just opened a can of worms. The WCSimRootTrigger structure implicitly assumes there is just one of : mode, jmu,jp per event. Possibly other things as well. It does store multiple vertices, but forgets the connection between vertices and particles which is not helpful. In an event with multiple vertices per event the information on jmu, jp and mode and possiblly more should be stored per vertex. Additionally things are stored in fixed sized arrays whereas they should really be stored in vectors. It should be possible to extend the format without breaking existing code but it will take a bit of work.

tdealtry:
Tricky...
It was my understanding that all truth information (other than some true tracks) is stored in the 0th WCSimRootTrigger. So is it possible to expand mode etc in the same way the vertices are stored?
And yes I do agree that vectors would be better! Not only would it prevent problems with very large events (i.e. with more than `MAX_N_PRIMARIES` vertices), but would also cut the file size for small events.

Trying to do it any other way would be very complicated, since there is not a 1:1 mapping between "true vertices" and "triggers". A couple of examples to show this:
- 500 MeV muon vertex only. Will potentially get two triggers, one from the muon light, one from the Michel electron
- 100 4 MeV electrons (sufficiently time separated). The default trigger efficiency isn't 100%, so will get < 100 triggers. Alternatively could loosen the trigger cuts and allow more dark-noise only events, and get > 100 triggers!

AlexFinch:
Yes I wouldn't change that. Jost tells me there could be up to 1 billion neutrinos from a supernova, slightly more than the current limit of a paltry 900! I'll look into making the changes.

tdealtry:
Thanks Alex!

PR #279, opened 2020-03-13
Docker CentOS

tdealtry:
* Use CentOS7 rather than SL6
* Clean the files up a bit
* Now when you start the container, WCSim is already sourced
* Don't install everything in /root/, since this doesn't work with singularity. Instead install in /home/nu/ (i.e. a new user called nu). This _might_ now work with singularity (but `singularity build` doesn't work on macOS so I can't check)
* Remove Dockerfile.build, and move clone of WCSim from Dockerfile.base to Dockerfile
* Dockerfile.base no longer exectuable


tdealtry:
@brichards64 looks like something isn't working with the auto docker deployment. https://hub.docker.com/r/wcsim/wcsim/tags states the latest tag is still 19 days old

PR #280, opened 2020-03-13
WIP: Supernova simulation

ast0815:
Based on #277, adds some more changes to enable the simulation of supernova events as produced by sntools.

Currently sntools produces a nuisance file with a single event per interaction. The vertex time of the event is specified in ms. As of right now, WCSim just uses that time directly as the starting time for the particle in the Geant simulation. Since all times are stored as sinlge precision floats, this leads to an inacceptable loss of precision for late interactions/events.

The simplest fix would be to use double precision everywhere.

TODO:

- [x] Fix the timing precision degradation


ast0815:
Eradicating *all* single precision floats from the source should solve the timing precision issue...

No kill like overkill. 💥

tdealtry:
FYI you've got a merge conflict that needs resolving. https://github.com/WCSim/WCSim/pull/278/files is where

ast0815:
Easy enough to fix.

tdealtry:
Using the standard 10 event WCSim.mac file (SK geometry, 500 MeV electrons), here are the stats on file sizes using this version of the code

```
The TTree "wcsimGeoT" takes 67301 bytes on disk
  Its branch "wcsimrootgeom" takes 67165 bytes on disk
The TTree "wcsimGeoT" takes 56856 bytes on disk
  Its branch "wcsimrootgeom" takes 56716 bytes on disk
Your WCSim version tree is 18.371% larger than the master GitHub version

The TTree "wcsimT" takes 1863920 bytes on disk
  Its branch "wcsimrootevent" takes 1863516 bytes on disk
  Its branch "wcsimrootevent_OD" takes 621 bytes on disk
The TTree "wcsimT" takes 1152348 bytes on disk
  Its branch "wcsimrootevent" takes 1151942 bytes on disk
  Its branch "wcsimrootevent_OD" takes 621 bytes on disk
Your WCSim version tree is 61.7497% larger than the master GitHub version
```
So this is quite a change!

Then taking moving the standard `WCSimRootGeom.hh` and `WCSimRootEvent.hh` files into this version, and modifying them so that the `Set()` arguments are doubles rather than floats (to keep the vector/array arguments working)
```
The TTree "wcsimGeoT" takes 56856 bytes on disk
  Its branch "wcsimrootgeom" takes 56716 bytes on disk
The TTree "wcsimGeoT" takes 56856 bytes on disk
  Its branch "wcsimrootgeom" takes 56716 bytes on disk
Trees sizes are identical!

The TTree "wcsimT" takes 1152390 bytes on disk
  Its branch "wcsimrootevent" takes 1151979 bytes on disk
  Its branch "wcsimrootevent_OD" takes 621 bytes on disk
The TTree "wcsimT" takes 1152348 bytes on disk
  Its branch "wcsimrootevent" takes 1151942 bytes on disk
  Its branch "wcsimrootevent_OD" takes 621 bytes on disk
Your WCSim version tree is 0.00364473% larger than the master GitHub version
```
As expected, we're back to (almost) perfect agreement

Then if we change the variables required for long events in the file back to doubles (time of true tracks, true hits, digitised hits, trigger time (aka Date))

```
The TTree "wcsimGeoT" takes 56856 bytes on disk
  Its branch "wcsimrootgeom" takes 56716 bytes on disk
The TTree "wcsimGeoT" takes 56856 bytes on disk
  Its branch "wcsimrootgeom" takes 56716 bytes on disk
Trees sizes are identical!

The TTree "wcsimT" takes 1369925 bytes on disk
  Its branch "wcsimrootevent" takes 1369521 bytes on disk
  Its branch "wcsimrootevent_OD" takes 621 bytes on disk
The TTree "wcsimT" takes 1152348 bytes on disk
  Its branch "wcsimrootevent" takes 1151942 bytes on disk
  Its branch "wcsimrootevent_OD" takes 621 bytes on disk
Your WCSim version tree is 18.8812% larger than the master GitHub version
```
This is of course non-negligible, but much better than 60%!

The 2 commits are at https://github.com/tdealtry/WCSim/commits/SN_simulation for the interested

ast0815:
I removed all single point floats because I could not figure out where the times are actually calculated and how to ensure that the calculation are also using double precision (as opposed to just storing doubles in the root file while using single precision floats somewhere in the calculations). If someone can do this, please go ahead.

tdealtry:
My comment was just that this PR currently makes file sizes much bigger, for a precision that we don't need for every variable. But you can pull my commits to make the file size change much smaller (20% vs 60%), without sacrificing the double precision for everything in the running code (in case we need to make other variable be saved as doubles in the output file).

@eosullivan (and @sk1806) you input would be helpful!

sk1806:
@tdealtry  I'm not sure I have anything useful to add other than, what you said sounds sensible...  keeping file size down where possible always gets my support

eosullivan:
I guess the question is whether a 20% increase will be a problem when doing large production runs. @sk1806 You recently had run the numbers for our required storage. Will this be a problem?

tdealtry:
Unfortunately storing times as doubles is required for some productions (e.g. SN) because the number of sig figs stored in a floating point is only ~7 (~14 for a double). This means when counting to 20 seconds in nanoseconds (i.e. at the end of a SN burst), the precision on the stored time is on the order of 10 microseconds when using floats, which isn't good enough

eosullivan:
Sure, but it isn't required for **all** productions and this implementation doesn't allow for an easy option to change it. Just wondering if this has an adverse impact to our overall computing plans.

tdealtry:
Yeah I understand. I was thinking of saying that we could have different output file formats for different productions, but that will probably open up teriffying issues!

eosullivan:
Yes, I would prefer to avoid this as well. Just want us to consider the impact of this file size increase and we can think of solutions if it is an issue.

sk1806:
I'm not going to object from a computing point of view as it sounds necessary for the physics.  

I was going to suggest if it is reasonable to introduce a param to switch between but you two are already discussing that.  I don't really have a feel of how much havoc it will cause to have these different outputs (of course uniform output will always be more simple), I think this is more a question for you software people.

I am expecting the computing predictions to evolve quite a lot over the next few years while software is being developed, so I do plan to update the numbers as we go and won't hold people to the initial computing predictions, but optimization is obviously important.   As everyone knows, fitqun dominates time and wcsim dominates filesize, so this is also worth bearing in mind in general.

eosullivan:
Ok, as discussed on the software call this morning, I have merged in #277. Can you please look through this PR and resolve the conflicts?

ast0815:
I have resolved the conflicts.

PR #285, opened 2020-04-15
Root6 compatability

tdealtry:
HK is wanting to migrate from Root5 to Root6, so I've had a look at what needs to be done. The important point is that there are no changes to the core WCSim code required, and the output of WCSim is identical to the root5 output

The changes that I've made are:
* For all the root macros, add headers and fix bugs that root6 doesn't let you get away with
*  Add root wrapper so we don't have to have the verbose code to load libraries/plot styles in every single macro
* Put the new `.pcm` file ROOT6 makes in the right place
* Makes a header file not executable (not related to Root6, but something that needs doing)


eosullivan:
Amazing! This has been in the to-dos for some time. You have tested with ROOT6?

eosullivan:
Also, will WCSim still work with ROOT5 after these changes?

tdealtry:
Yep macros work on both 5 & 6 (and in compiled / not compiled modes too - although of course compiled is best practice)

PR #286, opened 2020-04-22
Minor changes

tdealtry:
There's a massive memory leak when reading a WCSim root file, just doing a simple`GetEntry()` when the `WCSimRootEvent` branch is active. This was noticed when running on large (>1000) event files.
This fixes that by using a `TClonesArray` instead of `TObjArray`, and by using `Delete()`/`Clear()` appropriately
When viewing the running with htop, the memory use now stays constant at 0.1% precision over 10000 events, instead of adding about 5% memory use with every ~1000 events.

The output is identical (tested by grepping the output of `sample_readfile.C` from a 20 MeV electron file with 10 events and NoTrigger trigger - unfortunately `complete_comparison.C` won't work due to the `WCSimRootEvent` redefinition)

I also noticed that WCSimRootOptions is filled with junk for the NoTrigger trigger - I've set these to sensible values now

### DIFF
--- a/GNUmakefile_full
+++ b/GNUmakefile_full
@@ -71,7 +71,11 @@ libWCSim.a : $(ROOTOBJS)
 rootcint: ./src/WCSimRootDict.cc
 
 movedict: rootcint
-	$(shell cp ./src/WCSimRootDict_rdict.pcm .)
+ifneq (,$(wildcard /opt/HyperK/WCSim/src/WCSimRootDict_rdict.pcm))
+	cp -f ./src/WCSimRootDict_rdict.pcm .
+	cp -f ./src/WCSimRootDict_rdict.pcm ${G4WORKDIR}/tmp/${G4SYSTEM}/WCSim/
+endif
+
 
 doxy:
 	@if [ ${DOXYGEN_EXISTS} = 1 ]; \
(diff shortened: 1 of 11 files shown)


tdealtry:
Ok, so this has turned into something different...
I've reverted the commits related to memory leaks when editing WCSimRootEvent objects - it turns out that they caused a memory leak in WCSim itself (oops!) so we've gone for a different solution (making our own WCSimRootEvent rather than editing). But do be warned it you are editing a WCSimRootEvent, you'll see a leak!

So what's left in this PR? A few minor edits
- Only attempt to copy the pcm file if it exists (it won't when runninng root5)
- WCSimRootOptions trigger options are filled with junk when running with the NoTrigger trigger - I've set these to sensible values now
- Added `WCSimRootCherenkovDigiHit::SetPhotonIds()`. This is required for HKDAQ/TriggerApplication#60

@eosullivan I know these are a bit random. Is this single PR ok? Or do you prefer 3 smaller ones?

tdealtry:
Sorry, adding a 4th thing to the PR
- Adding a new trigger type `kTriggerTestVertices`. This is required by HKDAQ/TriggerApplication#31 (since TriggerApplication writes WCSim-like ROOT files, we need to add new trigger types into WCSim itself)

PR #292, opened 2020-09-22
Updated Radon model

gpronost:
New version of the Radon model (will be presented at the [software/physics pre-meeting this evening](https://indico.ipmu.jp/indico/event/381/contributions/5329/attachments/3361/4351/20200922_HK_Rn_MC.pdf))


gpronost:
I did some small modifications in the code: 
- the Rn model parameter were save in a file with a timestamp in its name which is redundant with git management
- add the Rn concentration in the fiducial volume in the debug
- correct the value returned by GetMeanActivity() which was not the activity but the model integral
- some minor modification in the model to make it more consistent across the different layers

gpronost:
Hi, is there any news about merging this PR? 
(I changed the name since we are now at 1.9.2)

WCSim.mac
```diff
@@ -197,8 +197,9 @@
 #/DAQ/RelativeHitTime true
 # The following set which scenario is used to scale SK model to HK
 #/mygen/radon_scaling A # 	Relative scaling with respect to full ID volume (Pessimistic)
-#/mygen/radon_scaling B # 	Relative scaling with respect to fiducial volume
-#/mygen/radon_scaling C #	Absolute scaling with respect to ID border (Optimistic)
+#/mygen/radon_scaling B #	Absolute scaling with respect to ID border (Optimistic)
+#/mygen/radon_scaling 0 #	Uniform distribution
```
tdealtry: Can you flag which option is the default?

include/RnModel_Fit_Params.hh
```diff
@@ -0,0 +1,220 @@
...
+
+	// Fit parameter for R2 layer = 9
+	vParam_Z [9][0] = -12.0212;
+	vParam_Z [9][1] = -10.1307;
+	vParam_Z [9][2] = CONC_INT_R2_9;
+	vParam_Z [9][3] = CONC_BOTTOM;
+	vParam_Z [9][4] = CONC_BOTTOM + 0.824749;
+	vParam_Z [9][5] = 67.1916;
+	vParam_Z [9][6] = 0.0125308;
+
```
tdealtry, line 132: Where are the values for Z layer = 0 and 1 filled?

include/RnModel_Fit_Params.hh
```diff
@@ -0,0 +1,220 @@
+/*
+	Rn model parameters version 2020/09/25
+	Author: Guillaume Pronost
+*/
```
tdealtry, line 36: Can you please document somewhere what these `vParam_Z` and `vParam_R2` actually are?
It seems like it's saying something about the detector radon concentration at different detector positions, but there are unanswered questions e.g.
- Is R2 layer = 0 detector centre, or detector edge?
- Is Z layer = 0 detector bottom, or detector top?
- Do these layers only go up to the edge of the ID, or do they go into the dead space / OD region too?
- What are the `[0]` to `[6]` of the second indice?
- What do negative values mean?

include/WCSimGenerator_Radioactivity.hh
```diff
@@ -12,6 +12,13 @@
 
 using namespace std;
 
+#define	RnModel_Bin_Rmax	10
+#define	RnModel_Bin_Zmin	 2
+#define	RnModel_Bin_Zmax	16
+#define	RnModel_Bin_Rmax_Step	16
+#define	RnModel_Bin_Zmax_Step	18
+#define 	GetSign(a) 		a<0?-1.:1.
```
tdealtry: Can you please make these all caps e.g. `RNMODEL_BIN_RMAX` It helps code legibility to separate precompiler constants from everything else
Can you also please document what these constants are? Include units if relevant

include/WCSimGenerator_Radioactivity.hh
```diff
@@ -24,133 +31,82 @@ class WCSimGenerator_Radioactivity
...
+		static double DiffusionR(double x, double origin, double factor);
+		
+		// Concentration functions
+		static double RadonFormulaR(double* val, double* par);
+		static double RadonFormulaZ(double* val, double* par);
 		
 		// Function
 		TF2* thRnFunction;
 		G4double fIntegral;
+		G4double fIntegralReco;
```
tdealtry: What does `Reco` mean in this context?

include/WCSimGenerator_Radioactivity.hh
```diff
@@ -24,133 +31,82 @@ class WCSimGenerator_Radioactivity
...
+	
+		G4double CONC_MIDDLE;
+		G4double CONC_INT_R2_6;
+		G4double CONC_INT_R2_7;
+		G4double CONC_INT_R2_8;
+		G4double CONC_INT_R2_9;
 		
 		static G4double fRn_Border;
 		
+		static G4double fHK_Z_min;
```
tdealtry: I see from later in the code that you're doing some scaling for "current detector geometry" relative to SK. Therefore, since WCSim can run multiple geometries (e.g. HK, SK, IWCD) the HK part of the variable name has the potential to be confusing
So something like `fCurrDet_Z_min` would be preferable

include/WCSimGenerator_Radioactivity.hh
```diff
@@ -24,133 +31,82 @@ class WCSimGenerator_Radioactivity
...
+		// Parameter
+		G4double CONC_BOTTOM;
+		G4double CONC_CENTER;
+		G4double CONC_INTERMEDIATE;
+	
+		G4double CONC_MIDDLE;
+		G4double CONC_INT_R2_6;
+		G4double CONC_INT_R2_7;
+		G4double CONC_INT_R2_8;
+		G4double CONC_INT_R2_9;
```
tdealtry: Opposite comment to above. Can these be made not all caps? e.g. `fConcBottom`

src/WCSimGenerator_Radioactivity.cc
```diff
@@ -199,64 +90,71 @@ void WCSimGenerator_Radioactivity::Initialize() {
 	
 	// Radon radioactivity constant
 	fRnLambda =  log(2) / (3.824*24.*3600.);
+	fRnLambda_Global = 0;
+		
+	// Parameter
+	CONC_BOTTOM 		= 2.63; // mBq/m^{3} -> From Nakano-san et al.
+	CONC_CENTER 		= 0.1;  // mBq/m^{3} -> From Nakano-san et al.
+	CONC_INTERMEDIATE 	= 0.3;  // Arbitrary
```
tdealtry: Is there a good reason why these values are set here, rather than inside `RnModel_Fit_Params.hh`? Personally I like the new file (it means that all the calibration results (which will eventually come from a database, I presume) can be updated clearly independently from the actual code)

tdealtry (changes requested):
Thanks for this @gpronost 
I particularly like how you've removed a lot of constants, and replaced them with a handful of arrays

There's some specific comments inline, but a more general one is that (as a complete beginner to this) I've no idea what a lot of the variables in `WCSimGenerator_Radioactivity` are. So could you please document each (doxygen preferred, but not essential) in `WCSimGenerator_Radioactivity.hh`
Then I'll take a closer look at `WCSimGenerator_Radioactivity.cc`

Note that I'm going to open/close this PR to force a Travis build

gpronost:
Hi Tom, thank you for your comments. I will try to answer them as soon as I have some time (it's been a while since I wrote this code).

include/RnModel_Fit_Params.hh
gpronost: I don't understand your question, vParam_Z[0][..] and vParam_Z[1][..] are filled in include/RnModel_Fit_Params.hh

WCSim.mac
gpronost: Done, scenario A should be default.

include/RnModel_Fit_Params.hh
gpronost: Explanations added

include/RnModel_Fit_Params.hh
gpronost: Explanations added

include/WCSimGenerator_Radioactivity.hh
gpronost: Done. 
Except for GetSign, do you also want it in caps?

include/WCSimGenerator_Radioactivity.hh
gpronost: I modified the name of the "reco" variables, it was for the "reconstruction volume" I renamed them as "FV" (fiducial volume).

include/WCSimGenerator_Radioactivity.hh
gpronost: Done

include/WCSimGenerator_Radioactivity.hh
gpronost: Done, actually I moved some as pre-compiler constants

src/WCSimGenerator_Radioactivity.cc
gpronost: Not really indeed, I moved them as pre-compiler constants inside RnModel_Fit_Params.hh

gpronost:
I added more documentations (Not in doxygen however, I'm sorry for this). 
I also modified the style of the variable names to match your criteria.

src/WCSimPrimaryGeneratorMessenger.cc
```diff
@@ -55,12 +55,12 @@ WCSimPrimaryGeneratorMessenger::WCSimPrimaryGeneratorMessenger(WCSimPrimaryGener
...
-  radonScalingCmd->SetGuidance("Select scalling scenario");
+  radonScalingCmd->SetGuidance("Select scalling scenario, if scenario 0 is selected, Bi214 are generated uniformly");
   radonScalingCmd->SetGuidance("[usage] /mygen/radon SCENARIO ");
-  radonScalingCmd->SetGuidance("     SCENARIO : A, B, C");
-  radonScalingCmd->SetCandidates("A B C");
+  radonScalingCmd->SetGuidance("     SCENARIO : 0, A, B");
+  radonScalingCmd->SetCandidates("0 A B");
   param = new G4UIparameter("SCENARIO",'s',true);
-  param->SetDefaultValue("C");
+  param->SetDefaultValue("B");
```
```suggestion
  param->SetDefaultValue("A");
```
According to the mac file, A should be default for now

src/WCSimGenerator_Radioactivity.cc
```diff
@@ -199,64 +83,51 @@ void WCSimGenerator_Radioactivity::Initialize() {
...
+	fConc_Int_R2_8	= 0;
+	fConc_Int_R2_9	= 0;
 	
+	fRn_PerPMT   = 30.; // mBq/PMT (equilibrium), assuming 20 mBq/Band + 10 mBq/PMT 
+//	fRn_PerPMT   = 24.; // mBq/PMT (equilibrium), assuming 20 mBq/Band +  4 mBq/PMT 
+		
 	// Detector size:
 	// HK (Should be extract from detector construction?)
 	//fHK_Z_max   = 66.8/2.;
 	//fHK_R_max   = 64.8/2.;
```
tdealtry:
```suggestion
```

src/WCSimGenerator_Radioactivity.cc
```diff
@@ -271,1170 +142,525 @@ void WCSimGenerator_Radioactivity::Configuration(G4int iScenario, G4double dLife
 	if ( dLifeTime > 0 )
 		fRnLambda = log(2) / (dLifeTime);
 	this->SetScenario(iScenario);
+	G4cout << " ========================================================================== " << G4endl;
 	
 	G4cout << " Livetime: " << log(2) / fRnLambda << " sec " <<  G4endl;
```
tdealtry:
```suggestion
	G4cout << " Mean lifetime: " << log(2) / fRnLambda << " sec " <<  G4endl;
```
I think this is meant to be mean lifetime, rather than livetime (which puts me in mind of "what time period is being simulated?")

src/WCSimGenerator_Radioactivity.cc
```diff
@@ -271,1170 +142,525 @@ void WCSimGenerator_Radioactivity::Configuration(G4int iScenario, G4double dLife
...
 	G4cout << " Lambda:   " << fRnLambda << "     " <<  G4endl;
 	G4cout << " Activity on ID border:   " << fRn_Border << " mBq/m^3  " <<  G4endl;
-	G4cout << " PMT number: " << myDetector->Get_Pmts()->size() << G4endl;
-	G4cout << " Surface: " << (2. * TMath::Pi() * fHK_R_max * fHK_R_max + 2. * TMath::Pi() * fHK_R_max * fHK_Z_max * 2. ) << G4endl;
-	//G4cout << fHK_R_max << " " << fHK_Z_max * 2 << G4endl;
-	G4cout << " Mean activity in the full ID:   " << fIntegral << " mBq  ( Concentration: " << fIntegral / (2. * TMath::Pi() * fHK_R_max * fHK_R_max * fHK_Z_max) << " mBq / m^3 ) " <<  G4endl;
+	if ( fScenario == 0 ) {
+		G4cout << " Scenario 0: Uniform Rn concentration is assumed " << G4endl;
+		fConcentrationID = 0;
+		fConcentrationFV = 0;
```
tdealtry, line 149: I'm probably missing something obvious, but where are these set for scenario 0? From what I can see, "uniform" means "uniformly no radon"

gpronost:
Last commit (d0baffd) should not have any effect, as these modifications are already on develop. I don't understand why github find difference however. If it's an issue, I will revert it.

src/WCSimGenerator_Radioactivity.cc
```diff
@@ -271,1170 +142,525 @@ void WCSimGenerator_Radioactivity::Configuration(G4int iScenario, G4double dLife
...
-				//+ dConcPMT        / cosh( sqrt(Lambda/(fRnDiffusion_Coef*4.)) * (sqrt(fChangeRMax) - sqrt(x)) )   // PMT
-				//+ dConcPMT * 0.05 / cosh( sqrt(Lambda/(fRnDiffusion_Coef*150.)) * (sqrt(fChangeRMax) - sqrt(x)) );   // Diffusion
-	
-	double dNorm10 = max((dConc10-dConc0),0.);
+	//----------------------------------//
+	// Make value arrays:
+	double vZ [2]	= { lZ , lR2	};
+	double vR2[2]	= { lR2, lZ 	};
+	//----------------------------------//
+	fRnLambda_Global = par[0];
```
tdealtry, line 588: Again, I might be missing something, but where is `par[1]` used?
From the header file (and top of this file)
```par[1] is the M / bin```

tdealtry (changes requested):
Thanks for the documentation. It really helped.

I've put a few more comments inline. Additionally
- You're using `std::cout` and `std::endl` instead of `G4cout`, `G4err` and `G4endl`. Can you switch to the G4 versions please? (I seem to remember a few years ago something I was doing not using the G4 versions caused some mangled printout (e.g. lines getting mixed up))
- Can you revert the last commit (about cmake & stuff)? It is easier (if you want it in for any testing, pulling from develop **should** work ok, since there are no conflicts) I agree that github handles this situation badly (it maybe shows the diff between your branch and the last time you pulled from the develop branch?)

src/WCSimGenerator_Radioactivity.cc
gpronost: They are not set for scenario 0. However "uniform" doesn't means "uniformly no radon", it only means we generate radon decay randomly in the detector without following a model (with an "uniform" distribution)

src/WCSimGenerator_Radioactivity.cc
gpronost: Ah, indeed, it's an artifact of the old version, I removed it I think, I will change the comments.

gpronost:
I included your comments

src/WCSimGenerator_Radioactivity.cc
gpronost: Yeah, I am sorry, I very often misspell "lifetime"

tdealtry (approved):
Wonderful. Many thanks for this @gpronost
I've just pushed a trivial change (removing the comment about `par[1]` in the .cc file - I couldn't do it as a suggestion on github so thought easiest thing to do it there)
Will wait for the travis to go green, then I'll merge

PR #293, opened 2020-09-23
Fix missing events when using kin_converter.py & upgrade to support Python 3.x

JostMigenda:
I encountered an issue where kin_converter.py appears to leave out some events (using the current develop branch of WCSim). See steps to reproduce at the end of this post.

I suspect this is due to the seek position in the file being off due to conflicts between manual use of the `seek` and `tell` functions and automatically iterating over lines in the file in the GetVertex function. ([some background](https://stackoverflow.com/a/29641787))

This problem appears to be specific to Python 2.x; in this PR, I’ve made some changes to allow the code to run under Python 3.x where it produces the expected results.

<s>Note: Currently, the code runs under both Python 2.x and Python 3.x, but gives different results. To avoid issues and confusion, we should probably add a check and quit early if it is run under Python 2.x—what do you think?</s> [Edit: see Tom’s comment below]

@eosullivan @tdealtry @sk1806 I think this will affect the supernova test production; so it would be good if you could take a look at this as soon as possible!

Full details of the issue, including sample output:
```
> sntools path/to/intp2001.data -f nakazato --start 100 --end 103 -v --out outfile.kin

[output omitted]

> tail outfile.kin 

$ begin
$ nuance -1001001
$ vertex -2673.54751 1805.73575 -1853.35047 102.97220016
$ track -12 44.34931 0.00000 0.00000 1.00000 -1
$ track 2212 938.27231 0.00000 0.00000 1.00000 -1
$ info 0 0 166
$ track -11 41.41111 -0.07745 0.98407 0.16005 0
$ track 2112 941.21051 0.05766 -0.73261 0.67820 0
$ end
$ stop
```

➔ So there are 167 events in total. (Note the "info" line!)

```
$ python2.7 kin_converter.py --input-filename outfile.kin --input-time-unit ms --dark-noise-start 100000000 --dark-noise-end 103000000 --event-overlap 0 --v 1 --fixed-duration 1000000

Is time ordered
# Generated on 2020-09-23 14:55:39.443669 with the options:
# Namespace(channel='all', detector='HyperK', distance=10.0, endtime=103.0, format='nakazato', hierarchy='noosc', input_file='../../../Astro/sntools/fluxes/intp2001.data', output='outfile.kin', randomseed=None, starttime=100.0, verbose=1)
# Split by kin_converter 2020-09-23 14:57:53.468352
# --fixed-duration 1000000
# --event-overlap 0

Event 0 corresponds to range 100.000000 ms 101.000000 ms
Skipping to position in file 0
contains 53 vertices
Event 1 corresponds to range 101.000000 ms 102.000000 ms
Skipping to position in file 18432
contains 44 vertices
Event 2 corresponds to range 102.000000 ms 103.000000 ms
Skipping to position in file 36864
contains 43 vertices
```
➔ Note that the total number of events is now 53 + 44 + 43 = 140, so 27 events have gone missing? And indeed:

```
> tail outfile.kin.000000000

$ track 2112 940.99432 -0.01142 -0.82586 0.56375 0
$ nuance -1001001
$ vertex 1809.43467 2232.02952 3283.42538 100.97875478
$ track -12 12.18543 0.00000 0.00000 1.00000 -1
$ track 2212 938.27231 0.00000 0.00000 1.00000 -1
$ info 0 0 52
$ track -11 10.82326 0.06939 -0.85369 0.51614 0
$ track 2112 939.63448 -0.06596 0.81142 0.58073 0
$ end
$ stop

> head -n 14 outfile.kin.000000001
# Generated on 2020-09-23 14:55:39.443669 with the options:
# Namespace(channel='all', detector='HyperK', distance=10.0, endtime=103.0, format='nakazato', hierarchy='noosc', input_file='../../../Astro/sntools/fluxes/intp2001.data', output='outfile.kin', randomseed=None, starttime=100.0, verbose=1)
# Split by kin_converter 2020-09-23 14:57:53.468352
# --fixed-duration 1000000
# --event-overlap 0
# Event 1
# /DarkRate/SetDarkLow  101000000
# /DarkRate/SetDarkHigh 102000000
$ begin
$ nuance -1001001
$ vertex -1020.49160 -1717.83248 -145.20323 101.10556478
$ track -12 28.98788 0.00000 0.00000 1.00000 -1
$ track 2212 938.27231 0.00000 0.00000 1.00000 -1
$ info 0 0 62
```
➔ So the events with numbers 53 to 61 appear to have gone missing?

### DIFF
--- a/sample-root-scripts/MakeKin.py
+++ b/sample-root-scripts/MakeKin.py
@@ -1,7 +1,6 @@
 #!/usr/bin/python
-# import the math module  
-import math  
-  
+from __future__ import print_function
+
 from optparse import OptionParser
 import random
 from math import pi,sin,cos,sqrt
@@ -21,7 +20,7 @@ detectors = {"SuperK":[3368.15/2., 3620.],
              "Cylinder_60x74_20inchBandL_14perCent":[7400./2., 6000.],
              "Cylinder_60x74_20inchBandL_40perCent":[7400./2., 6000.]}
 
-for pname, no in pid.items():
+for pname, no in list(pid.items()):
     if pname.endswith('+'):
         pid[pname.replace('+','-')] = -1*pid[pname]
 
@@ -45,7 +44,7 @@ parser.add_option("-V","--nVerticesPerEvent",dest="verticesPerEvent",
                   help=" Average number of vertices to simulate per event. Default: %s" \
                   % (verticesPerEventDefault),
                   metavar="#", default=verticesPerEventDefault)
-optchoices = pid.keys()
+optchoices = list(pid.keys())
 optdefault = "mu-"
 parser.add_option("-t", "--type", dest="type",
                   help="Particle type to be generated. Choices: %s. Default: %s" \
@@ -69,7 +68,7 @@ parser.add_option("-d", "--direction", dest="dirname",
                   help="Type of direction. Choices: %s. Default: %s" \
                       % (optchoices, optdefault),
                   choices=optchoices, default=optdefault)
-optchoices = detectors.keys()
+optchoices = list(detectors.keys())
 optdefault = "SuperK"
 parser.add_option("-w", "--detector", dest="detector",
                   help="Detector water volume to use (for vertex positioning). Choices: %s. Default: %s" \
@@ -103,7 +102,7 @@ if options.vertname == "center":
 elif options.vertname == "random":
     randvert = True
 elif options.vertname == "wall":
-    print >>sys.stderr, "Wall not implemented yet"
+    print("Wall not implemented yet", file=sys.stderr)
     sys.exit(3)
 elif options.vertname == "minusx":
     if options.detector == "SuperK":
@@ -126,7 +125,7 @@ elif options.vertname == "plusz":
     else:
         particle["vertex"] = (0, 0, +1000. * detectors[options.detector][1] / detectors["SuperK"][1])
 else:
-    print >>sys.stderr, "Don't understand vertex",opttions.vertname
+    print("Don't understand vertex", options.vertname, file=sys.stderr)
     sys.exit(2)
 
 if options.dirname == "towall":
@@ -138,10 +137,10 @@ elif options.dirname == "tocap":
 elif options.dirname == "4pi":
     randdir = True
 elif options.dirname == "wall":
-    print >>sys.stderr, "Wall not implemented yet"
+    print("Wall not implemented yet", file=sys.stderr)
     sys.exit(3)
 else:
-    print >>sys.stderr, "Don't understand direction",options.dirname
+    print("Don't understand direction", options.dirname, file=sys.stderr)
     sys.exit(2)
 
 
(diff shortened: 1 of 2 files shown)


sample-root-scripts/kin_converter.py
```diff
@@ -1,5 +1,7 @@
 #!/bin/env python2
```
tdealtry, line 1:
```suggestion
#!/bin/env python3
```

sample-root-scripts/kin_converter.py
```diff
@@ -1,5 +1,7 @@
...
 
+from __future__ import print_function
+
```
tdealtry, line 4:
```suggestion
```

tdealtry (changes requested):
Thank @JostMigenda for both finding and fixing the bug!
I think given that it doesn't give sensible results with python2, we should just call it a python3 script - I think these changes will mean it's run in python3 by default (if the user makes it executable) and crash in python2

tdealtry:
And sorry @eosullivan this will require v1.9.1 tagging for WCSim, and @sk1806 a slight delay to the HK release

JostMigenda:
Since it would be a bit awkward if one script required Python 2 and the other one required Python 3, I added a commit with the very minor changes necessary to let MakeKin.py run under both versions of Python.

sample-root-scripts/MakeKin.py
```diff
@@ -21,7 +20,7 @@
              "Cylinder_60x74_20inchBandL_14perCent":[7400./2., 6000.],
              "Cylinder_60x74_20inchBandL_40perCent":[7400./2., 6000.]}
```
tdealtry, line 21:
```suggestion
             "Cylinder_60x74_20inchBandL_40perCent":[7400./2., 6000.],
             "HyperK":[7080./2., 5480.]}
```
These numbers are taken from https://github.com/WCSim/WCSim/blob/7ee6f09b8d8fa26d179ffb7a9b22571e172286cc/src/WCSimDetectorConfigs.cc#L222-L223

sample-root-scripts/MakeKin.py
```diff
@@ -45,7 +44,7 @@
                   help=" Average number of vertices to simulate per event. Default: %s" \
                   % (verticesPerEventDefault),
                   metavar="#", default=verticesPerEventDefault)
-optchoices = pid.keys()
+optchoices = list(pid.keys())
 optdefault = "mu-"
```
tdealtry, line 48:
```suggestion
optdefault = None
parser.add_option("-s", "--seed", dest="seed",
                  help="Random number seed to use. Default: None (use system time)",
                  metavar="SEED", default=optdefault)
optchoices = pid.keys()
optdefault = "mu-"
```

tdealtry: And then immediately after parsing the arguments, need (you didn't edit a line close enough so github wouldn't let me suggest for this one)
```
random.seed(options.seed)
```

tdealtry (changes requested):
Off topic, but since you're updating this it's probably a good idea to add a random seed option
Also, add the HK detector as currently defined in WCSim

JostMigenda:
I wasn’t sure whether to include that, since I didn’t want this PR to grow to large; but as it’s just a few lines, I’ll add it. Thanks, Tom!

@eosullivan As far as I’m concerned, this PR is now ready to be merged (& for a v1.9.1 tag); but please let me know if you want any changes! I’m reasonably free today & tomorrow and should be able to make requested changes within a few hours.

eosullivan:
@sk1806 I guess we don't distribute python in our HK software containers (the same way as we do with the GEANT, ROOT packages, for example)? I guess Python3 is become more and more the standard these days, so it's probably fine to do this (though I will admit that I still have Python2 running on my computer).

sk1806:
@eosullivan   For the containers we can install python3 and then set it to be the default so that should be fine. 
For people using their local clusters or whatever it could be more of an issue, but I think as you said we're at the stage where we can just assume people have or can get python3  (or we can suggest people use a container if not).

JostMigenda:
Development of Python 2 [has ended](https://blog.python.org/2020/04/python-2718-last-release-of-python-2.html) and most major scientific Python packages [have already dropped support or will do so by the end of the year](https://python3statement.org); so if there are any local clusters that don’t yet have Python 3 installed, WCSim will be the least of their problems. Furthermore, with my updates MakeKin.py still runs under Python 2; while kin_converter.py produced incorrect results anyway, so we’re not actually losing any functionality. For those reasons, I don’t think there are any compatibility issues with this PR.

PR #297, opened 2020-10-06
prod/ODWithWLS

mahditaani:
This is the latest version of the hyper-k od. This includes the most recent hyper-k geometry as well as fixes to the photocoverage code.


tdealtry:
Thanks for this @mahditaani (& co) We're planning to start reviewing this PR soon, so can you let us know if it's ready to go?
(Note that there are conflicts that need fixing anyway)

tdealtry:
@mahditaani I've done the merge, but can't push it to your branch. So can you pull the [merge commit](https://github.com/tdealtry/WCSim/commit/ea929653e038596a3c1a5d096316b305efd18d21) from my branch? Something like
```bash
#make sure you're on the right branch
git checkout prod/ODWithWLS
#grab my changes
git remote add tom git@github.com:tdealtry/WCSim.git
git pull tom prod/ODWithWLS
#push them to your repo
git push origin prod/ODWithWLS
```

mahditaani:
Hi Tom,

Thank you!

I will do it and run some checks.

All the best,
Mahdi

mahditaani:
@tdealtry I tried to run this and while the code compiles it does not build an OD geometry (aborts when you try to run OD_10k.mac). I am looking into this to see what the issue is.

HKComputing/Create1TeVUpMuJobs.py
```diff
@@ -0,0 +1,53 @@
+from HKODWCSimMacro import HKODWCSimMacro
```
tdealtry: Could you remove the files in this folder from this PR? And then submit a new PR with them?
This is a very much decoupled feature, and would be good to get more peoples' input on. This kind of mac file building & job submission is very useful, but since there isn't an official script in WCSim already, has been invented many times by many people. We should converge on one (at first glance, this looks nice)

HyperKODComputing/CreateJobs.py
```diff
@@ -0,0 +1,22 @@
+import sys
```
tdealtry: similarly please put this folders' files in the new PR

MuonFlux-HyperK-ThetaPhi.dat
```diff
@@ -0,0 +1,18000 @@
+0 0 0.005 0 0.01 1 0 2 1.45271e-12 452.074
```
tdealtry: This file should be documented - both it's format, and where the data has come from (a paper? some other software?)
Maybe comment lines e.g. starting with a `#`? Maybe in a README? Maybe in `WCSim.mac`?

I'd also be in favour of moving this file to a new folder `$WCSIMDIR/data` (or `inputs/` @eosullivan what do you think?), to keep $WCSIMDIR clean. We can then store in there other useful data files e.g. biasprofile.dat

WCSim.mac
```diff
@@ -172,6 +172,12 @@
 #/gps/ang/minphi 0 deg
 #/gps/ang/maxphi 360 deg
 
+# Or you can use a cosmics generator
+# Based on the SuperK flux extrapolated at HyperK site in Mozumi
+# Muons are generated according to this flux and their position is extrapolated outside the tank
+# Useful to study muons tracks in both OD and ID
+#/mygen/generator cosmics
+
```
tdealtry, line 182: It's good practice to show the existence of all options in `WCSim.mac`
```suggestion
#/mygen/cosmicsfile data/MuonFlux-HyperK-ThetaPhi.dat 

```
although it looks like the default in the Messenger is inputcosmicfile (which won't exist) so perhaps it should be uncommented

src/WCSimPrimaryGeneratorMessenger.cc
```diff
@@ -13,20 +13,26 @@ WCSimPrimaryGeneratorMessenger::WCSimPrimaryGeneratorMessenger(WCSimPrimaryGener
...
   fileNameCmd->SetGuidance("Select the file of vectors.");
   fileNameCmd->SetGuidance(" Enter the file name of the vector file");
   fileNameCmd->SetParameterName("fileName",true);
   fileNameCmd->SetDefaultValue("inputvectorfile");
+
+  fileNameCmdCosmics = new G4UIcmdWithAString("/mygen/cosmicsfile",this);
+  fileNameCmdCosmics->SetGuidance("Select the file of cosmics.");
+  fileNameCmdCosmics->SetGuidance(" Enter the file name of the cosmics file");
+  fileNameCmdCosmics->SetParameterName("fileName",true);
+  fileNameCmdCosmics->SetDefaultValue("inputcosmicfile");
```
```suggestion
  fileNameCmdCosmics->SetDefaultValue("data/MuonFlux-HyperK-ThetaPhi.dat");
```

src/WCSimWCSD.cc
```diff
@@ -160,24 +160,24 @@ G4bool WCSimWCSD::ProcessHits(G4Step* aStep, G4TouchableHistory*)
...
-    maxQE = fdet->GetPMTQE(WCCollectionName,wavelength,0,240,660,ratio);
-    photonQE = fdet->GetPMTQE(volumeName, wavelength,1,240,660,ratio);
+    // maxQE = fdet->GetPMTQE(WCIDCollectionName,wavelength,0,200,700,ratio);
+    maxQE = fdet->GetPMTQE(WCCollectionName,wavelength,0,200,700,ratio);
+    photonQE = fdet->GetPMTQE(WCCollectionName, wavelength,1,200,700,ratio);
     photonQE = photonQE/maxQE;
   }else if (fdet->GetPMT_QE_Method() == 3){
     ratio = 1./(1.-0.25);
-    photonQE = fdet->GetPMTQE(volumeName, wavelength,1,240,660,ratio);
+    photonQE = fdet->GetPMTQE(WCCollectionName, wavelength,1,200,700,ratio);
```
tdealtry: What's the reason for changing the wavelength range here? Is it valid for all PMTs?
Also, what's the reason for using `WCCollectionName` instead of `volumeName`?

This hopefully shouldn't be affecting the travis CI (which use `Stacking_Only` or QE method 1)

src/WCSimTuningParameters.cc
```diff
@@ -19,6 +19,9 @@ WCSimTuningParameters::WCSimTuningParameters()
  tvspacing = 100.0;
  topveto = false;
 
+ WCCladdingReflectivity   = 0.90; //
```
tdealtry: According to the tuning parameter messenger, this is the OD WLS cladding (rather than any other cladding). Can you use a more descriptive name, such as
```suggestion
 WCODWLSCladdingReflectivity   = 0.90; //
```

src/WCSimTrackingAction.cc
```diff
@@ -22,11 +22,6 @@ WCSimTrackingAction::WCSimTrackingAction()
   ParticleList.insert(-321); // kaon-
   ParticleList.insert(311); // kaon0
   ParticleList.insert(-311); // kaon0 bar
```
tdealtry, line 24: Think this needs reverting
```suggestion
  ParticleList.insert(-311); // kaon0 bar

  ParticleList.insert(11);
  ParticleList.insert(-11);
  ParticleList.insert(12);
  ParticleList.insert(-12);
```

src/WCSimTuningMessenger.cc
```diff
@@ -47,6 +47,16 @@ WCSimTuningMessenger::WCSimTuningMessenger(WCSimTuningParameters* WCTuningPars):
   TopVeto->SetParameterName("TopVeto",true);
   TopVeto->SetDefaultValue(0);
 
+  CommandWCCladdingReflectivity = new G4UIcmdWithADouble("/WCSim/tuning/WCCladdingReflectivity",this);
```
tdealtry: See my comment on WCSimTuningParameters.cc for renaming the OD WLS cladding variables to be more explicit

include/WCSimTuningParameters.hh
```diff
@@ -36,6 +36,12 @@ public:
 
   void SaveOptionsToOutput(WCSimRootOptions * wcopt);
 
+  G4double GetWCCladdingReflectivity() {return WCCladdingReflectivity;}
```
tdealtry: See my comment on WCSimTuningParameters.cc for renaming the OD WLS cladding variables/methods to be more explicit

include/WCSimTuningMessenger.hh
```diff
@@ -36,6 +36,8 @@ private:
   //For Top Veto - jl145
   G4UIcmdWithADouble* TVSpacing;
   G4UIcmdWithABool* TopVeto;
+  G4UIcmdWithADouble* CommandWCCladdingReflectivity;
```
tdealtry: See my comment on WCSimTuningParameters.cc for renaming the OD WLS cladding variables to be more explicit

WCSim.cc
```diff
@@ -32,6 +32,8 @@ void file_exists(const char * filename) {
 
 int main(int argc,char** argv)
 {
+
+  gRandom->SetSeed(0);
```
tdealtry:
```suggestion
```
gRandom seed is now set with `/WCSim/random/seed` in the same way as the CLHEP random generator is

include/jhfNtuple.h
```diff
@@ -62,7 +62,7 @@ static const char* ntDesc =
 "vecRecNumber:I,"
 "jmu:I,"
 "jp:I,"
-"npar[0,MAX_N_PRIMARIES]:I,"
+"npar[0,50]:I,"
```
tdealtry: Somethings gone wrong with the merge here... can you revert all changes in this file?

WCSim.mac
```diff
@@ -203,15 +209,14 @@
...
-## change the name of the output root file, default = wcsim.root
-
 # Needed to prevent other decay than Bi-214
 /grdm/nucleusLimits 214 214 83 83 
 # if you need to use time bias
 #/grdm/decayBiasProfile biasprofile.dat
 #/grdm/analogueMC 0
 #/grdm/BRbias 0
 
+## change the name of the output root file, default = wcsim.root
```
tdealtry, line 221: I can't put this comment in the right place, but...
line 23 - can you add something like (make sure the filename is correct)
```
# if you want to use the OD, please see $WCSIMDIR/macros/OD.mac for the latest design, as well as the complete set of OD geometry options
```

macros/OD.mac
```diff
@@ -0,0 +1,132 @@
+# Sample setup macro with no visualization
```
tdealtry, line 1: Please rename this file OD_13k.mac

macros/OD.mac
```diff
@@ -0,0 +1,132 @@
...
+#/DarkRate/SetDarkRate 4.2 kHz #This is the value for SKI set in SKDETSIM.
+#/DarkRate/SetDarkRate 8.4 kHz #For 20 inch HPDs and Box and Line PMTs, based on High QE 20" R3600 dark rate from EGADS nov 2014
+#/DarkRate/SetDarkRate 3.0 kHz #For 12 inch HPDs and Box and Line PMTs, based on High QE 20" R3600 dark rate from EGADS nov 2014
+
+# command to multiply the dark rate.
+# Convert dark noise frequency before digitization to after digitization by setting suitable factor
+# Again, this is now a PMT property and can be overridden here
+#/DarkRate/SetConvert 1.367  #For Normal PMT
+#/DarkRate/SetConvert 1.119 #For HPDs
+#/DarkRate/SetConvert 1.126 #For Box and Line PMTs
```
tdealtry: Can you move the relevant parts of this to below each 
```
/DarkRate/SetDetectorElement
```
line? i.e. have to copies of the DarkRate & DarkConvert parts, in a similar way to the DarkMode parts

There's a lot of potential of confusion without this, with people (me included) attempting to turn off ID dark noise but ending up doing the OD instead (even in geometries without the OD!)

This comment applies for the other OD mac files

macros/OD__10k_OD_PMTs__no_ID__no_dark_noise__1_GeV_horizontal_muon.mac
```diff
@@ -0,0 +1,132 @@
+# Sample setup macro with no visualization
```
tdealtry: Is this file required? If so, please rename it so it reflects the actual differences between this and OD_10k.mac (which seem to be a different WLS plate size, and a different spacial/energy distribution for the muons)
Having the same output filename as OD_10k.mac is potentially confusing too

macros/OD.mac
```diff
@@ -0,0 +1,132 @@
...
+/WCSim/HyperKOD/ODDeadSpace 600 mm
+/WCSim/HyperKOD/ODTyvekSheetThickness 1 mm
+/WCSim/HyperKOD/ODWLSPlatesThickness 0.6 cm
+/WCSim/HyperKOD/ODWLSPlatesLength 30 cm
+/WCSim/HyperKOD/PMTODperCellHorizontal 0
+/WCSim/HyperKOD/PMTODperCellVertical 0
+/WCSim/HyperKOD/PMTODPercentCoverage 0.28
+/WCSim/HyperKOD/ODPMTShift 0 m
+#/WCSim/HyperKOD/DeactivateWLS
+/WCSim/HyperKOD/BuildCladding
```
tdealtry: Can you please comment each parameter?
This applies for the other OD mac files

macros/OD.mac
```diff
@@ -0,0 +1,132 @@
...
+/gps/direction 1 0 0
+/gps/ene/type Gauss
+/gps/ene/mono 600 MeV
+/gps/ene/sigma 1 MeV
+
+
+
+## change the name of the output root file, default = wcsim.root
+
+/WCSimIO/RootFile wcsim_13k_OD_PMTs_no_ID_no_dark_noise_horizontal_muon.root
```
tdealtry: This says that there's no ID. How is that achieved? How can the ID be activated? It'd be good to have an example macro with ID+OD

src/WCSimTuningParameters.cc
```diff
@@ -19,6 +19,9 @@ WCSimTuningParameters::WCSimTuningParameters()
...
+ WCODTyvekReflectivity   = 0.97; //
```
tdealtry:
```suggestion
 WCODTyvekReflectivity   = 0.90; //
```
the default appears to be 0.97 in WCSimTuningMessenger.cc and tuning_parameters.mac

prod.py
```diff
@@ -0,0 +1,19 @@
+from CreateMacro import *
```
tdealtry: another thing for the macro/job submission PR
(FYI there appears to be something missing - grepping `CreateMacro` can't find anything - but the jobs ones exist)

tdealtry (changes requested):
Thanks for this @mahditaani 
Good spot on the part causing the OD to not build!

I've been through some of the changes (see mostly minor comments inline) but got a lot more to do tomorrow

For reference, https://github.com/WCSim/Validation/tree/master/Generate/macReference are the files run in the Travis CI, and the output is posted at https://wcsim.github.io/Validation/

include/WCSimPrimaryGeneratorAction.hh
```diff
@@ -5,19 +5,19 @@
...
 #include "globals.hh"
 #include "jhfNtuple.h"
-#include <vector>
-
 #include <fstream>
 
 #include "WCSimRootOptions.hh"
 #include "WCSimGenerator_Radioactivity.hh"
 
+#include "TH1D.h"
```
tdealtry:
```suggestion
```
this doesn't appear to be used

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -90,8 +96,66 @@ WCSimPrimaryGeneratorAction::WCSimPrimaryGeneratorAction(
   useGunEvt    		= false;
   useLaserEvt  		= false;
   useGPSEvt    		= false;
+  useCosmics            = false;
   useRadioactiveEvt  	= false;
   useRadonEvt        	= false;
+
+  // Create the relevant histograms to generate muons
```
tdealtry, line 103: From what I can work out, the `/mygen/cosmicsfile` option doesn't work correctly
When that is called, it reads a file into `inputFile` rather than `inputFileCosmics`, and the histograms created here in the constructor aren't updated. This needs fixing

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -90,8 +96,66 @@ WCSimPrimaryGeneratorAction::WCSimPrimaryGeneratorAction(
...
+      phiMin=(atof(token[6]));
+      phiMax=(atof(token[7]));
+      flux=(atof(token[8]));
+      Emean=(atof(token[9]));
+
+      hFluxCosmics->SetBinContent(binPhi,binCos,flux);
+      hEmeanCosmics->SetBinContent(binPhi,binCos,Emean);
+    }
+
+    TFile *file = new TFile("flux.root","RECREATE");
```
```suggestion
    TFile *file = new TFile("cosmicflux.root","RECREATE");
```

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -90,8 +96,66 @@ WCSimPrimaryGeneratorAction::WCSimPrimaryGeneratorAction(
...
+    hFluxCosmics->Write();
+    hEmeanCosmics->Write();
+    file->Close();
```
tdealtry: Is this required? Or was it just for debugging purposes?
Perhaps it could be an mac option if (and where) to save these histograms?

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -288,9 +352,6 @@ void WCSimPrimaryGeneratorAction::GeneratePrimaries(G4Event* anEvent)
   {      // manual gun operation
     particleGun->GeneratePrimaryVertex(anEvent);
 
-    //To prevent occasional seg fault from an un assigned targetpdg 
-    targetpdgs[0] = 2212; //ie. proton
-
```
tdealtry: Please revert this

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -332,20 +393,19 @@ void WCSimPrimaryGeneratorAction::GeneratePrimaries(G4Event* anEvent)
   }
   else if (useLaserEvt)
     {
-      targetpdgs[0] = 2212; //ie. proton
```
tdealtry, line 335: please revert this

src/WCSimWLSProperties.cc
```diff
@@ -0,0 +1,379 @@
...
+  return planck * (lightspeed/E);
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////
+// ELJEN EJ-286
+// http://www.eljentechnology.com/products/wavelength-shifting-plastics/ej-280-ej-282-ej-284-ej-286
+#define nEntries_WLS_rindex 60
+
+EljenEJ286::EljenEJ286() { SetgAbs(); SetgEm(); SethEm();}
+EljenEJ286::~EljenEJ286(){}
```
tdealtry, line 24: There's graphs created in the constructor (via the `SetgAbs()` etc methods). Please clean them up in the destructor

src/WCSimWLSProperties.cc
```diff
@@ -0,0 +1,379 @@
...
+
+  static const int emitArraySize = 15;
+  float emitWavelength[emitArraySize] = {390,400,410,420,430,440,450,460,470,480,490,500,510,520,530};
+  float emit[emitArraySize] =
+      {0.0, 0.1, 0.71, 0.98,
+       0.91, 0.63, 0.49, 0.35,
+       0.20, 0.13, 0.081, 0.052,
+       0.03, 0.021, 0};
+
+  emissionHist = new TH1F("hist","hist", 15, 390., 530.);
```
tdealtry, line 404: Please check if the graph exists, and delete it if it does

src/WCSimWLSProperties.cc
```diff
@@ -0,0 +1,379 @@
...
+    //    gAbs->SetPoint(i,round(hWL(GetPhotonEnergy_ABS()[i]))*nm,AbsAmp[i]);
+    gAbs->SetPoint(i,round(hWL(GetPhotonEnergy_ABS()[i]))*nm,1. - exp(-1.2*cm/GetAbs()[i]));
+    // std::cout << round(hWL(GetPhotonEnergy_ABS()[i]))*nm << "nm -> " << AbsAmp[i] << std::endl;
+  }
+
+}
+
+void EljenEJ286::SetgEm(){
+  gEm = new TGraph();
+  const int step = 10;
```
tdealtry:
```suggestion
```
Don't think this is used

src/WCSimWLSProperties.cc
```diff
@@ -0,0 +1,379 @@
...
+  for(int i=nEntries_abslength_eljen-1; i>0; i--){
+    //    gAbs->SetPoint(i,round(hWL(GetPhotonEnergy_ABS()[i]))*nm,AbsAmp[i]);
+    gAbs->SetPoint(i,round(hWL(GetPhotonEnergy_ABS()[i]))*nm,1. - exp(-1.2*cm/GetAbs()[i]));
+    // std::cout << round(hWL(GetPhotonEnergy_ABS()[i]))*nm << "nm -> " << AbsAmp[i] << std::endl;
+  }
+
+}
+
+void EljenEJ286::SetgEm(){
+  gEm = new TGraph();
```
tdealtry: It's more efficient to allocate the TGraph space required, rather than it having to do it multiple times as it's filled
```suggestion
  gEm = new TGraph(nEntries_eljen_component);
```
Also, please check if the graph exists, and delete it if it does

src/WCSimWLSProperties.cc
```diff
@@ -0,0 +1,379 @@
...
+  //       0, 0, 0, 0, 0, // 650 --- 600
+  //       0, 0, 0, 0, 0, // 600 --- 550
+  //       0, 0, 0, 0, 0, // 550 --- 500
+  //       0, 0, 0, 0, 0, // 500 --- 450
+  //       0, 0, 0.01, 0.05, 0.64, // 450 --- 400
+  //       0.93, 0.9999, 0.9999, 0.9999, 0.9999, // 400 --- 350
+  //       0.9999, 0.9999, 0.99, 0.97, 0.89, // 350 --- 300
+  //       0.71, 0.54, 0, 0, // 300 --- 250
+  //       0, 0, 0, 0, 0}; // 250 --- 200
+  gAbs = new TGraph();
```
tdealtry: It's more efficient to allocate the TGraph space required, rather than it having to do it multiple times as it's filled
```suggestion
  gEm = new TGraph(nEntries_abslength_eljen);
```
Also, please check if the graph exists, and delete it if it does

src/WCSimWLSProperties.cc
```diff
@@ -0,0 +1,379 @@
...
+  const int step = 10;
```
tdealtry:
```suggestion
```
Don't think this is used

include/WCSimWLSProperties.hh
```diff
@@ -0,0 +1,58 @@
...
+  G4double* GetAbs();
+
+  G4int     GetNumEntries_EM();
+  G4double* GetPhotonEnergy_EM();
+  G4double* GetEm();
+
+  void SetgAbs();
+  void SetgEm();
+
+  void SethEm();
```
tdealtry, line 65:
```suggestion
 private:
  void SetgAbs();
  void SetgEm();

  void SethEm();
```

src/WCSimRunAction.cc
```diff
@@ -103,6 +103,12 @@ void WCSimRunAction::BeginOfRunAction(const G4Run* /*aRun*/)
   //set detector & random options
   wcsimdetector->SaveOptionsToOutput(wcsimrootoptions);
   wcsimrandomparameters->SaveOptionsToOutput(wcsimrootoptions);
+
+  //set Tree for gathering photon info inside stepping action
+  photonTree = new TTree("photons","Photon Tree");
```
tdealtry: This looks like a nice feature, but we need to be careful about bloating file sizes. Can you provide some file size information?
I'd suggest adding an boolean option to save this in `WCSimRunActionMessenger`

src/WCSimConstructMaterials.cc
```diff
@@ -180,6 +181,28 @@ void WCSimDetectorConstruction::ConstructMaterials()
...
+  //---PVT - WLS plates eljen EJ-286
+  // linear formula [CH2CH(C6H4CH3)]n
+  density = 1.023*g/cm3;  // at 20deg
+  G4Material* WLS_PVT
+      = new G4Material("WLS_PVT",density,2);
+  WLS_PVT->AddElement(elC, 9); // PVT
+  WLS_PVT->AddElement(elH, 10);
+
+
+  //**cladding properties
```
tdealtry: Same comment about renaming this as being explicitly OD WLS cladding, rather than any other form of cladding

src/WCSimConstructMaterials.cc
```diff
@@ -838,11 +865,105 @@ void WCSimDetectorConstruction::ConstructMaterials()
...
+  const G4int nEntries_transmittance = 33;
+  G4double WLS_transmittance_vs_energy[nEntries_transmittance] =
+      { 1-0.020836, 1-0.0207796, 1-0.0207198, 1-0.0206584, 1-0.0205953,
+        1-0.0205288, 1-0.0204589, 1-0.0203874, 1-0.0203125, 1-0.0202344,
+        1-0.0201512, 1-0.0200647, 1-0.019975, 1-0.0198787, 1-0.0197774,
+        1-0.0196696, 1-0.0195553, 1-0.0194329, 1-0.0193041, 1-0.0191639,
+        1-0.019016, 1-0.0188553, 1-0.0186803, 1-0.0184931, 1-0.0182871,
+        1-0.0180644, 1-0.0178189, 1-0.0175494, 1-0.0172519, 1-0.0169209,
+        1-0.0165512, 1-0.0161367, 1-0.0156689};
+  WlsPlasticMPT->AddProperty("TRANSMITTANCE", WLSProps->GetPhotonEnergy(), WLS_transmittance_vs_energy, nEntries_transmittance);
```
tdealtry: Does it make sense for this to go into `WCSimWLSProperties`, like the other WLS properties?

include/WCSimDetectorConstruction.hh
```diff
@@ -309,6 +371,7 @@ private:
   // ############################# //
 
   bool isODConstructed;
+  bool isCombinedPMTCollectionDefined;
```
tdealtry, line 374: This isn't initialised in the constructor to `false`. It should be

include/WCSimDetectorConstruction.hh
```diff
@@ -337,6 +402,10 @@ private:
   // Hit collection name parameters
   G4String WCODCollectionName;
 
+  // WLS material name
+  bool isWLSFilled;
+  bool BuildCladding;
```
tdealtry: Please label as OD WLS cladding

include/WCSimDetectorMessenger.hh
```diff
@@ -45,6 +46,25 @@ class WCSimDetectorMessenger: public G4UImessenger
...
+  G4UIcmdWithAString* PMTODRadius;
+  // #PMTs and position
+  G4UIcmdWithAnInteger* PMTODperCellHorizontal;
+  G4UIcmdWithAnInteger* PMTODperCellVertical;
+  G4UIcmdWithADouble* PMTODPercentCoverage;
+  G4UIcmdWithADoubleAndUnit* ODPMTShift;
+  // isWLS constructed
+  G4UIcmdWithoutParameter* isWLSFilled;
+  // build cladding around WLS
+  G4UIcmdWithoutParameter* BuildCladding;
```
tdealtry: Same comment - please label as OD WLS cladding

src/WCSimDetectorMessenger.cc
```diff
@@ -107,6 +107,119 @@ WCSimDetectorMessenger::WCSimDetectorMessenger(WCSimDetectorConstruction* WCSimD
...
+  ODWLSPlatesLength->SetGuidance("Set OD WLS plates length (unit: cm mm).");
+  ODWLSPlatesLength->SetParameterName("ODWLSPlatesLength", true);
+  ODWLSPlatesLength->SetDefaultValue(60.);
+  ODWLSPlatesLength->SetUnitCategory("Length");
+  ODWLSPlatesLength->SetDefaultUnit("cm");
+  ODWLSPlatesLength->SetUnitCandidates("cm mm");
+
+  // Nb of OD PMT per cell HORIZONTALLY
+  PMTODperCellHorizontal = new G4UIcmdWithAnInteger("/WCSim/HyperKOD/PMTODperCellHorizontal", this);
+  PMTODperCellHorizontal->SetGuidance("Set number of OD PMT per cell HORIZONTALLY (unit: cm mm).");
```
tdealtry:
```suggestion
  PMTODperCellHorizontal->SetGuidance("Set number of OD PMT per cell HORIZONTALLY.");
```

src/WCSimDetectorMessenger.cc
```diff
@@ -107,6 +107,119 @@ WCSimDetectorMessenger::WCSimDetectorMessenger(WCSimDetectorConstruction* WCSimD
...
+  PMTODperCellHorizontal->SetParameterName("PMTODperCellHorizontal", true);
+  PMTODperCellHorizontal->SetDefaultValue(1);
+  PMTODperCellHorizontal->SetRange("PMTODperCellHorizontal>=0");
+
+  // Nb of OD PMT per cell HORIZONTALLY
+  PMTODperCellVertical = new G4UIcmdWithAnInteger("/WCSim/HyperKOD/PMTODperCellVertical", this);
+  PMTODperCellVertical->SetGuidance("Set number of OD PMT per cell VERTICALLY (unit: cm mm).");
```
tdealtry:
```suggestion
  PMTODperCellVertical->SetGuidance("Set number of OD PMT per cell VERTICALLY.");
```

src/WCSimDetectorMessenger.cc
```diff
@@ -107,6 +107,119 @@ WCSimDetectorMessenger::WCSimDetectorMessenger(WCSimDetectorConstruction* WCSimD
...
+  PMTODperCellVertical->SetParameterName("PMTODperCellVertical", true);
+  PMTODperCellVertical->SetDefaultValue(1);
+  PMTODperCellVertical->SetRange("PMTODperCellVertical>=0");
+
+  // Nb of OD PMT per cell HORIZONTALLY
+  PMTODPercentCoverage = new G4UIcmdWithADouble("/WCSim/HyperKOD/PMTODPercentCoverage", this);
+  PMTODPercentCoverage->SetGuidance("Set global OD photocoverage percentage (unit: cm mm).");
```
tdealtry:
```suggestion
  PMTODPercentCoverage->SetGuidance("Set global OD photocoverage percentage.");
```

src/WCSimEventAction.cc
```diff
@@ -867,8 +865,7 @@ void WCSimEventAction::FillRootEvent(G4int event_id,
   // First two tracks come from jhfNtuple, as they are special
 
   int k;
-  //Modify to add decay products
-  for (k=0;k<jhfNtuple.npar;k++) // should be just 2
+  for (k=0;k<2;k++) // should be just 2
```
tdealtry: Please revert this

src/WCSimEventAction.cc
```diff
@@ -982,8 +977,6 @@ void WCSimEventAction::FillRootEvent(G4int event_id,
 	parentType = -211;
       } else if (pionList.count(trj->GetParentID()) ) {
 	parentType = 211;
-      } else if (primaryList.count(trj->GetParentID()) ) {
-	parentType = 1;
```
tdealtry: Please revert this

src/WCSimEventAction.cc
```diff
@@ -908,7 +905,6 @@ void WCSimEventAction::FillRootEvent(G4int event_id,
   // same, april 7th 2005
   std::set<int> pionList;
   std::set<int> antipionList;
-  std::set<int> primaryList;
```
tdealtry: Please revert this

src/WCSimEventAction.cc
```diff
@@ -939,7 +935,6 @@ void WCSimEventAction::FillRootEvent(G4int event_id,
     if ( trj->GetPDGEncoding() == 211 ) pionList.insert(trj->GetTrackID());
     if ( trj->GetPDGEncoding() == -211 ) antipionList.insert(trj->GetTrackID());
        
-    if( trj->GetParentID() == 0 ) primaryList.insert(trj->GetTrackID());
```
tdealtry, line 942: Please revert this

src/WCSimPMTObject.cc
```diff
@@ -1965,3 +2011,46 @@ G4double PMT3inch::GetDarkRateConversionFactor(){
...
+  G4int iPt=0;
+  std::map<G4double, G4double>::iterator itr;
+  for(itr = mapQE.begin(); itr != mapQE.end(); itr++) {
+    gQE->SetPoint(iPt,itr->first,itr->second);
+    iPt++;
+  }
+}
+
+WCSimBasicPMTObject::~WCSimBasicPMTObject(){
+}
```
tdealtry, line 2062: Please cleanup the TGraph

src/WCSimPMTObject.cc
```diff
@@ -1965,3 +2011,46 @@ G4double PMT3inch::GetDarkRateConversionFactor(){
...
+
+WCSimBasicPMTObject::WCSimBasicPMTObject(std::vector<G4double> wl,std::vector<G4double> qe,G4double max){
+  wavelength=wl;
+  QE=qe;
+  maxQE=max;
+}
+
+void WCSimBasicPMTObject::DefineQEHist(std::map<G4double,G4double> mapQE){
+  gQE = new TGraph();
```
tdealtry:
```suggestion
  gQE = new TGraph(mapQE.size());
```
should be more efficient.
Please also check if the TGraph exists

include/WCSimPMTObject.hh
```diff
@@ -6,7 +6,8 @@
 #include "Randomize.hh"
 #include <map>
 #include <vector>
-
+#include <TH1F.h>
```
tdealtry:
```suggestion
```
not used

src/WCSimPMTObject.cc
```diff
@@ -1965,3 +2011,46 @@ G4double PMT3inch::GetDarkRateConversionFactor(){
...
+G4int PMT3inch::GetNbOfQEDefined(){
+  const G4int factor = 20;
+  return factor;
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////
+// COMBINED PMTS CLASS
+// Goal is to instantiate purely virtual WCSimPMTObject Classes to contained an extended
+//  QE information of several PMTs collection
+// Useful for applying QE in stacking action when several PMTs collection are defined
```
tdealtry, line 2024: - In the ID-only case, this combined class isn't used, and the QE values are taken from the PMT class directly
- In the ID+OD case, this combined class is used, and the QE values are taken from a combination of ID & OD PMTs QE values inside `WCSimDetectorConstruction::CreateCombinedPMTQE()`

My reading of this is that there's a potential for differences in the ID response between running and ID-only & and ID+OD geometry. This is of course something we'd like to avoid. Am I missing something?

src/WCSimPMTQE.cc
```diff
@@ -38,7 +38,7 @@ G4double WCSimDetectorConstruction::GetPMTQE(G4String CollectionName, G4double P
 
   // return 0 for wavelenght outside the range
   if (flag==1){
-    if (PhotonWavelength <= low_wl || PhotonWavelength >= high_wl || PhotonWavelength <=280 || PhotonWavelength >=660){
+    if (PhotonWavelength <= low_wl || PhotonWavelength >= high_wl ){
```
tdealtry: This is a potential cause of the difference seen from the Travis CI tests. Please revert for now so we can test

include/WCSimSteppingAction.hh
```diff
@@ -1,19 +1,35 @@
 #ifndef WCSimSteppingAction_h
 #define WCSimSteppingAction_h 1
 
+#include <TTree.h>
```
tdealtry:
```suggestion
```

src/WCSimSteppingAction.cc
```diff
@@ -13,6 +17,66 @@
...
+  double distance = absorbedPhotonPosition.mag();
+  return 0.2/(1+distance);
+}// End of EmittedPhotonDetectedProbability function.
+
+
+// This function will determine whether or not a photon will be detected by the PMT. This function
+// uses only the position of the photon's absorption with respect to the PMT (distance from PMT) and
+// lab measurements of the efficiency as a function of distance, taken at Edinburgh University.
+bool EmittedPhotonDetected(G4ThreeVector absorbedPhotonPosition){
+  TRandom3 *r = new TRandom3(time(0)); // This can be changed to whatever random number generator WCSim is using.
```
tdealtry: Creating a new `TRandom3` everytime this is called will give a big memory leak.
If you want to keep a separate random number generator just for this process, please create it in the constructor (and delete it in the destructor), and come up with an appropriate way to set the seed (e.g. in `WCSimRandomParameters::SetSeed()`)
Otherwise `G4UniformRand()` should work

src/WCSimSteppingAction.cc
```diff
@@ -13,6 +17,66 @@
 #include "G4SDManager.hh"
 #include "G4RunManager.hh"
 
+const double planck = 4.1357e-15; // ev.s
+const double lightspeed = 299792458e9; // nm/s
+
+// This function will determine the probability of an emitted photon will being detected by the PMT.
```
tdealtry: Please make clear that these are WLS-related (or even WLS+PMT-system-related) functions

src/WCSimSteppingAction.cc
```diff
@@ -27,17 +91,14 @@ void WCSimSteppingAction::UserSteppingAction(const G4Step* aStep)
...
-//    G4cout << " PROBLEM! " << theTrack->GetCreatorProcess()->GetProcessName() <<
-//  std::flush << G4endl;
-//  }
-  
+  // WLS
+  bool WLS=false;
+  WLS=true;
+  if(WLS){
+    CollectPhotonsInfoInWLSPlates(aStep);
+    // WLSPhysicsProcess(aStep);
```
tdealtry: Why is this commented? Is the method even required anymore?

src/WCSimSteppingAction.cc
```diff
@@ -27,17 +91,14 @@ void WCSimSteppingAction::UserSteppingAction(const G4Step* aStep)
...
-//  G4double vx = aMomentum.x();
-//  G4int ix = std::isnan(vx);
-//  if(ix != 0){
-//    G4cout << " PROBLEM! " << theTrack->GetCreatorProcess()->GetProcessName() <<
-//  std::flush << G4endl;
-//  }
-  
+  // WLS
+  bool WLS=false;
+  WLS=true;
```
tdealtry: A better way would be to set this to `true` only if there is WLS in the geometry

src/WCSimSteppingAction.cc
```diff
@@ -160,3 +221,228 @@ double WCSimSteppingAction::FieldLines(G4double /*x*/,G4double /*y*/,G4int /*coo
...
+//	phot->position->SetX(absorbedPhotonPosition->x());
+//	phot->position->SetY(absorbedPhotonPosition->y());
+//	phot->position->SetZ(absorbedPhotonPosition->z());
+
+  // Here we can use TF1 and the GetRandom() function to selected from the emission spectra.
+
+  // We don't know the spatial distribution of the emitted photon so we will just keep this distribution uniform
+  // over the surface of a sphere.
+
+  TRandom3 *r = new TRandom3(time(0)); // This can be changed to whatever random number generator WCSim is using.
```
tdealtry: Another memory leak. Please fix

src/WCSimStackingAction.cc
```diff
@@ -26,32 +26,40 @@ G4ClassificationOfNewTrack WCSimStackingAction::ClassifyNewTrack
...
 
   // Make sure it is an optical photon
   if( particleType == G4OpticalPhoton::OpticalPhotonDefinition() )
     {
       G4double photonWavelength = (2.0*M_PI*197.3)/(aTrack->GetTotalEnergy()/eV);
-      G4double ratio = 1./(1.0-0.25);
+      // MF : translated from skdetsim : better to increase the number of photons
+      // than to throw in a global factor  at Digitization time !
+      //G4double ratio = 1./(1.0-0.25);
+      G4double ratio = 0.5;
```
tdealtry: The ratio has been changed from 4/3 to 1/2. Why?
This could be the cause of the Travis CI differences. Please revert for now so we can check

src/WCSimStackingAction.cc
```diff
@@ -26,32 +26,40 @@ G4ClassificationOfNewTrack WCSimStackingAction::ClassifyNewTrack
...
-	  // than to throw in a global factor  at Digitization time !
-	  G4double ratio = 1./(1.0-0.25);
-	  // XQ: get the maximum QE and multiply it by the ratio
-	  // only work for the range between 240 nm and 660 nm for now 
-	  // Even with WLS
-	  G4double wavelengthQE = 0;
 	  if (DetConstruct->GetPMT_QE_Method() == 1){
-	    wavelengthQE  = DetConstruct->GetPMTQE(WCIDCollectionName,photonWavelength,1,240,660,ratio);
+	    if(!DetConstruct->GetIsCombinedPMTCollectionDefined())
+	      wavelengthQE  = DetConstruct->GetPMTQE(WCIDCollectionName,photonWavelength,1,200,700,ratio);
```
tdealtry:
```suggestion
	      wavelengthQE  = DetConstruct->GetPMTQE(WCIDCollectionName,photonWavelength,1,240,660,ratio);
```

src/WCSimStackingAction.cc
```diff
@@ -26,32 +26,40 @@ G4ClassificationOfNewTrack WCSimStackingAction::ClassifyNewTrack
...
+      //G4double ratio = 1./(1.0-0.25);
+      G4double ratio = 0.5;
+      // XQ: get the maximum QE and multiply it by the ratio
+      // only work for the range between 240 nm and 660 nm for now 
+      // Even with WLS
       G4double wavelengthQE = 0;
       if(aTrack->GetCreatorProcess()==NULL) {
-	wavelengthQE  = DetConstruct->GetPMTQE(WCIDCollectionName,photonWavelength,1,240,660,ratio);
+	if(!DetConstruct->GetIsCombinedPMTCollectionDefined())
+	  wavelengthQE  = DetConstruct->GetPMTQE(WCIDCollectionName,photonWavelength,1,200,700,ratio);
```
tdealtry:
```suggestion
	  wavelengthQE  = DetConstruct->GetPMTQE(WCIDCollectionName,photonWavelength,1,240,660,ratio);
```
Another change that could be in the Travis CI. And seemingly, this one doesn't affect the OD. So please revert

src/WCSimStackingAction.cc
```diff
@@ -26,32 +26,40 @@ G4ClassificationOfNewTrack WCSimStackingAction::ClassifyNewTrack
...
-	  G4double wavelengthQE = 0;
 	  if (DetConstruct->GetPMT_QE_Method() == 1){
-	    wavelengthQE  = DetConstruct->GetPMTQE(WCIDCollectionName,photonWavelength,1,240,660,ratio);
+	    if(!DetConstruct->GetIsCombinedPMTCollectionDefined())
+	      wavelengthQE  = DetConstruct->GetPMTQE(WCIDCollectionName,photonWavelength,1,200,700,ratio);
+	    else
+	      wavelengthQE  = DetConstruct->GetStackingPMTQE(photonWavelength,1,200,700,ratio);
 	  }else if (DetConstruct->GetPMT_QE_Method() == 2){
+	    if(!DetConstruct->GetIsCombinedPMTCollectionDefined())
+	      wavelengthQE  = DetConstruct->GetPMTQE(WCIDCollectionName,photonWavelength,0,200,700,ratio);
```
tdealtry:
```suggestion
	      wavelengthQE  = DetConstruct->GetPMTQE(WCIDCollectionName,photonWavelength,0,240,660,ratio);
```

src/WCSimStackingAction.cc
```diff
@@ -26,32 +26,40 @@ G4ClassificationOfNewTrack WCSimStackingAction::ClassifyNewTrack
...
+	    else
+	      wavelengthQE  = DetConstruct->GetStackingPMTQE(photonWavelength,0,200,700,ratio);
 	    wavelengthQE  = DetConstruct->GetPMTQE(WCIDCollectionName,photonWavelength,0,240,660,ratio);
```
tdealtry: Did you mean to delete this line?

src/WCSimDetectorConfigs.cc
```diff
@@ -243,15 +243,16 @@ void WCSimDetectorConstruction::SetHyperKWithODGeometry()
   WCPMTName           = PMT->GetPMTName();
   WCPMTExposeHeight   = PMT->GetExposeHeight();
   WCPMTRadius         = PMT->GetRadius();
-  WCIDDiameter          = 70.8*m; // = 74m - 2*(60cm ID wall + 1m OD)
-  WCIDHeight            = 54.8*m; // = 60m - 2*(60cm ID wall + 2m OD)
+  WCIDDiameter          = 64.8*m; // = 68m - 2*(60cm ID wall + 1m OD)(69m with 0.5m concrete each side)
+  WCIDHeight            = 65.8*m; // = 71m - 2*(60cm ID wall + 2m OD)(72m with 0.5m concrete on each side)
```
tdealtry, line 247: Are these the official "this is how big the HK tank will be" numbers? Do you have a reference for them?
(If these numbers are correct, I'll sort out the ID-only geometry. It should of course be the same as the ID part of this geometry)

src/WCSimDetectorConstruction.cc
```diff
@@ -297,3 +298,131 @@ void WCSimDetectorConstruction::SaveOptionsToOutput(WCSimRootOptions * wcopt)
...
+    maxQEVec.push_back(PMT[iPMT]->GetmaxQE());
+    NbOfWLBins.push_back(PMT[iPMT]->GetNbOfQEDefined());
+  }
+
+  // Concatenate WL vec and remove duplicate
+  std::map<G4double,G4double> QE;
+
+  // Recursive algorithm to set new QE for combined PMT collection
+  // F. Nova: define QE for a given wavelength as sqrt(QE1^2 + QE2^2 + ...), where each QEi is the value of QE for PMT type i
+  // replace it with maximum
```
tdealtry: Please clarify the comment - looks like `max()` rather than `sqrt()` is used now

src/WCSimDetectorConstruction.cc
```diff
@@ -297,3 +298,131 @@ void WCSimDetectorConstruction::SaveOptionsToOutput(WCSimRootOptions * wcopt)
...
+    for(std::map<G4double, G4double>::iterator it=QEMap[iCol].begin(); it!=QEMap[iCol].end(); ++it){
+      if(iCol+1<QEMap.size()){
+        std::map<G4double, G4double>::iterator foundWL = QEMap[iCol+1].find(it->first);
+        if(foundWL == QEMap[iCol+1].end()){
+          G4cout << "Undefined QE in next collection" << G4endl;
+          G4cout << "Will add it" << G4endl;
+          G4cout << " ### " << it->first << "nm : " << it->second << G4endl;
+          QE[it->first]=it->second;
+        } else {
+	  //	  max_QE = sqrt(it->second*it->second + foundWL->second*foundWL->second);
```
tdealtry:
```suggestion
```

sample-root-scripts/llib.C
```diff
@@ -0,0 +1,16 @@
+void llib(){
```
tdealtry: This file isn't required anymore. Please use `$WCSIMDIR/rootwc` instead to automatically load WCSim libraries & includes

sample-root-scripts/ODAnalysis.C
```diff
@@ -0,0 +1,329 @@
...
+// WCSim Includes
+#include "WCSimRootEvent.hh"
+#include "WCSimRootGeom.hh"
+
+// Defines
+#define PI 3.141592654
+
+
+/*
+* To run this script just type:  root -l -x llib.C 'ODAnalysis.C("wcsim.root", "ODHits.root", false)'
```
tdealtry:
```suggestion
* To run this script just type:  rootwc -l -x 'ODAnalysis.C("wcsim.root", "ODHits.root", false)' 
```

sample-root-scripts/ODAnalysis.C
```diff
@@ -0,0 +1,329 @@
...
+
+// Defines
+#define PI 3.141592654
+
+
+/*
+* To run this script just type:  root -l -x llib.C 'ODAnalysis.C("wcsim.root", "ODHits.root", false)' 
+* or replace wcsim.root and ODHits.root for your input and output filenames respectively
+* The final true can be changed to true to turn on verbosity.
+* The default: root -l -x llib.C ODAnalysis.C will run with the variables mentioned above
```
tdealtry:
```suggestion
* The default: root -l -x ODAnalysis.C will run with the variables mentioned above
```

sample-root-scripts/ODAnalysis.C
```diff
@@ -0,0 +1,329 @@
...
+	if (wcsimdirenv != NULL) {
+		gSystem->Load("$WCSIMDIR/libWCSimRoot.so");
+		gSystem->Load("$WCSIMDIR/libWCSimRoot.rootmap");
+                gSystem->Load("$WCSIMDIR/libWCSimRootDict_rdict.pcm");
+	} else {
+		std::cout << "Error: WCSIMDIR environment variable not set." << std::endl;
+	}
+
+
+}// End of loadlibs function
```
tdealtry:
```suggestion
```

sample-root-scripts/ODAnalysis.C
```diff
@@ -0,0 +1,329 @@
...
+
+void ODAnalysis( const char *inFileName = "wcsim.root", const char *outFileName = "ODHits.root", bool verbosity = 0){ 
+	
+//	loadlibs(); // Load the required libraries
```
tdealtry:
```suggestion
```

sample-root-scripts/ODAnalysis.C
```diff
@@ -0,0 +1,329 @@
...
+		std::cout << "Input file: " << inFileName << std::endl;
+	}
+	
+
+
+	// Get a pointer to the tree from the input file
+	TTree *wcsimTree = (TTree*) inFile->Get("wcsimT");
+	
+	// Get the number of events in the tree
+	long int nEvent = wcsimTree->GetEntries();
```
tdealtry:
```suggestion
	const long int nEvent = wcsimTree->GetEntries();
```

sample-root-scripts/ODAnalysis.C
```diff
@@ -0,0 +1,329 @@
...
+	TH2F *numHitPMTsVsPhi = new TH2F("numHitPMTsVsPhi", "numHitPMTsVsPhi", nBinsAngle , angleMin, angleMax, nBinsHits, hitsMin, hitsMax );
+	TH2F *totalRawHitsVsTheta = new TH2F("totalRawHitsVsTheta", "totalRawHitsVsTheta", nBinsAngle , angleMin, angleMax, nBinsHits, hitsMin, hitsMax );
+	TH2F *totalRawHitsVsPhi = new TH2F("totalRawHitsVsPhi", "totalRawHitsVsPhi", nBinsAngle , angleMin, angleMax, nBinsHits, hitsMin, hitsMax );
+	TH2F *totalDigiHitsVsTheta = new TH2F("totalDigiHitsVsTheta", "totalDigiHitsVsTheta", nBinsAngle , angleMin, angleMax, nBinsHits, hitsMin, hitsMax );
+	TH2F *totalDigiHitsVsPhi = new TH2F("totalDigiHitsVsPhi", "totalDigiHitsVsPhi", nBinsAngle , angleMin, angleMax, nBinsHits, hitsMin, hitsMax );
+
+
+
+
+	for (int ev = 0; ev < nEvent; ev++){ // Loop over events
```
tdealtry:
```suggestion
	for (long int ev = 0; ev < nEvent; ev++){ // Loop over events
```

sample-root-scripts/trigger_OD.C
```diff
@@ -0,0 +1,201 @@
+#include <stdio.h>     
+#include <stdlib.h>
+#include <libgen.h>
+
+void trigger_OD(char *filename=NULL) {
+  /* A simple script to plot aspects of phototube hits 
+   * 
+   * I like to run this macro as 
+   * $ root -l -x 'trigger_OD.C("OD.root")'
```
tdealtry:
```suggestion
   * $ rootwc -l -x 'trigger_OD.C("OD.root")'
```

sample-root-scripts/trigger_OD.C
```diff
@@ -0,0 +1,201 @@
...
+  gROOT->Reset();
+  char* wcsimdirenv;
+  wcsimdirenv = getenv ("WCSIMDIR");
+  if(wcsimdirenv !=  NULL){
+    gSystem->Load("${WCSIMDIR}/libWCSimRoot.so");
+    gSystem->Load("${WCSIMDIR}/libWCSimRoot.rootmap");
+    gSystem->Load("${WCSIMDIR}/src/WCSimRootDict_rdict.pcm");
+  }else{
+    std::cout << "Can't load WCSim ROOT dictionaries" << std::endl;
+  }
```
tdealtry:
```suggestion
```

sample-root-scripts/ODAnalysis.C
```diff
@@ -0,0 +1,329 @@
+// C++ Includes
```
tdealtry: From a quick scan, it looks like this file and `read_OD.C` are very similar, with the major differences being
- ODAnalysis saves plotted 2D histogram canvases, in addition to the TTree
- read_OD saves information from ID & OD, rather than just the OD
Is it possible to merge them into one macro? There could be `bool plot` and `bool saveID` arguments

tdealtry (changes requested):
Here's my other set

There's some ideas of what's going on with the Travis plots. Hopefully we can see agreement after some reversions, then talk about which changes are relevant for ID+OD, and which should be OD-only

macros/OD__10k_OD_PMTs__no_ID__no_dark_noise__1_GeV_horizontal_muon.mac
mahditaani: You are right, this is a redundant macro and will be removed.

src/WCSimTuningParameters.cc
mahditaani: Good spot! yes, I will implement this suggestion.

macros/OD.mac
mahditaani: Sure, done.

WCSim.mac
mahditaani: I have changed the code a little so it correctly points to the default file data/MuonFlux-HyperK-ThetaPhi.dat  and so I have added the comment to WCSim.mac as you have suggested.

src/WCSimPrimaryGeneratorMessenger.cc
mahditaani: Changed the default value so that it is correct. As you have suggested.

macros/OD.mac
mahditaani: This is achieved by stopping the ID PMTs from being built (using an if statement) but this is not yet changeable from the mac files so I have removed the feature and changed the filenames.

MuonFlux-HyperK-ThetaPhi.dat
mahditaani: I really like this suggestion. If there are other inputs that other groups are creating it would be useful to have a data folder. I will put the file in there for now.

src/WCSimPrimaryGeneratorAction.cc
mahditaani: You are right, the code was calling a single input file rather than a separate file for comics and vector data. I have fixed this see: 52e063abba1152a0a90cb1d0c7ab900b82f60f07.

src/WCSimPrimaryGeneratorAction.cc
mahditaani: Done d48090901ca1bd3c2a2ee4f5cf0328e77755cc66

src/WCSimWCSD.cc
mahditaani: Perhaps @federiconova could comment on this?

src/WCSimTuningParameters.cc
mahditaani: Done. See 12b24ff7c8d3984fb566ed476d87e0d474f29781

src/WCSimTuningMessenger.cc
mahditaani: Done. See 12b24ff7c8d3984fb566ed476d87e0d474f29781

include/WCSimTuningParameters.hh
mahditaani: Done. See 12b24ff7c8d3984fb566ed476d87e0d474f29781

include/WCSimTuningMessenger.hh
mahditaani: Done. See 12b24ff7c8d3984fb566ed476d87e0d474f29781

WCSim.cc
mahditaani: Nice! Done, see 2f8abe44d7a556aa7959c6c59330988786bd25ba

src/WCSimWCSD.cc
federiconova: Hi. About the argument, it seemed to me GetPMTQE always expects a collection name, not a volume name (see definition in WCSimPMTQE.cc and other calls in WCSimStackingAction.cc). Is that wrong?

About the wavelength, I guess enlarging the range is always fine? We might be interested in those extra photons.

Anyway please revert this changes if they seem problematic.

WCSim.mac
mahditaani: Done. See fb97cd96bc61ac50f1060fe6965a8289dd18324c

macros/OD.mac
mahditaani: I think it might be better to keep this as OD.mac because this is the default baseline design. We are currently simulating an additional 10k geometry which is why there is also an OD_10k.mac. What do you think?

include/WCSimPrimaryGeneratorAction.hh
mahditaani: You are right, it wasn't used. Removed it. See c8b9e4448a76a9f6b26a635d1b0680429cddb160

src/WCSimDetectorMessenger.cc
mahditaani: Yup, silly mistake. Good catch. Fixed: 3722b24dd5ba65e301dc282f56908ab9b549e09b

src/WCSimDetectorMessenger.cc
mahditaani: Yup, silly mistake. Good catch. Fixed: 3722b24dd5ba65e301dc282f56908ab9b549e09b

src/WCSimDetectorMessenger.cc
mahditaani: Yup, silly mistake. Good catch. Fixed: 3722b24dd5ba65e301dc282f56908ab9b549e09b

src/WCSimDetectorConfigs.cc
mahditaani: Yeah these numbers were shown several collaboration meetings ago and are the numbers now stated in the technical report.  I will note that changing to the new numbers caused some stuck tracks but I can't figure out where they are coming from (I couldn't find any overlaps).

src/WCSimDetectorConstruction.cc
mahditaani: Hi @federiconova do you have any thoughts on this?

src/WCSimDetectorConstruction.cc
mahditaani: Removed. See efaef16e74a05b14a5ced84158e3ae083a71ee8a and 2d485754aab0323152f3f7896280c07f24dd4eef

include/WCSimSteppingAction.hh
mahditaani: Removed. See efaef16e74a05b14a5ced84158e3ae083a71ee8a and 2d485754aab0323152f3f7896280c07f24dd4eef

include/WCSimPMTObject.hh
mahditaani: Removed. See efaef16e74a05b14a5ced84158e3ae083a71ee8a and 2d485754aab0323152f3f7896280c07f24dd4eef

src/WCSimDetectorConstruction.cc
federiconova: Hi, is this function used at all? I suspect it isn't.
I think initially it was defined as sqrt and then I replaced it with max, but eventually we are not even using it?

tdealtry:
Thanks for this @mahditaani & @federiconova I'll have a proper look tomorrow (github actions have sponged all my time today... but look to have something working now. Will close/open this PR to trigger a build)

include/jhfNtuple.h
mahditaani: Done. See 57f135ea8e0cabdff66650875e4f1271ca03dcf9

src/WCSimWCSD.cc
mahditaani: I will change the wavelengths back as this is probably what is causing the CI tests to fail.

sample-root-scripts/trigger_OD.C
mahditaani: I am unable to use this rootwc command after building with cmake. How is it supposed to be used?

src/WCSimPMTQE.cc
mahditaani: I guess this is an optimisation for ID PMTs but the OD will use WLS plates that can detect photons with less than 280nm wavelength. I will revert but this will need to be included for the OD to function correctly.

federiconova: That's right. But anyway, why is the test failing? If extending the possible wavelengths changes the results for the ID, doesn't that mean that they *should* be extended, because some photons are being missed?

tdealtry:
Doing another close/open to check if the permissions for uploading to the validation website are fixed

tdealtry:
Ok, so no idea what's happened there. It seems to have run the validation on WCSim/develop rather than mahditaani/prod/ODWithWLS. I think I've got a fix, so will do another close/open (but of course it could be that a close/open by me isn't the same as a push by you...)
On the plus side, the action is now much faster, now we're not simulating dark noise for a **long** time in the validation

tdealtry:
I'm happy the validation is correctly working on your branch now, but the output is still different to what's expected unfortunately... will delve next week

src/WCSimStackingAction.cc
mahditaani: Looking through the code, it seems that this just arbitrarily increases the QE. I am not sure I understand why this is even here. @federiconova if I remember correctly you had a reason for changing this. In any case, I will revert for now.

src/WCSimStackingAction.cc
mahditaani: Yes it would seem so!

src/WCSimStackingAction.cc
federiconova: ok, thanks for this work!

src/WCSimDetectorConstruction.cc
mahditaani: @federiconova on line 376 the std::max() function is used. Am I correct in saying that the new QE of a combined collection is the highest QE for that wavelength?

federiconova: Yes.

src/WCSimWLSProperties.cc
mahditaani: Done, see cbd79d33abcdca16a32d19fd8ae3294921acd240

src/WCSimWLSProperties.cc
mahditaani: Done, see cbd79d33abcdca16a32d19fd8ae3294921acd240

src/WCSimWLSProperties.cc
mahditaani: Done, see cbd79d33abcdca16a32d19fd8ae3294921acd240

sample-root-scripts/trigger_OD.C
mahditaani: Never mind, it is because it isn't automatically copied when building with cmake.

src/WCSimSteppingAction.cc
mahditaani: I think I will remove these functions in another pull request as they aren't used any more.

src/WCSimPMTObject.cc
mahditaani: @federiconova do you have any comments on this?

federiconova: Hi, sorry, I am happy to comment but what are the incriminated lines? I only see four commented lines.

src/WCSimWLSProperties.cc
mahditaani: Done, see cbd79d33abcdca16a32d19fd8ae3294921acd240

src/WCSimPMTObject.cc
mahditaani: I see that in `WCSimDetectorConfigs` we make use of the combined PMT QE for the OD configuration. Does this affect the QE of the ID PMTs at all?

src/WCSimConstructMaterials.cc
mahditaani: Yes that is a good suggestion. I have moved it.

src/WCSimPMTObject.cc
federiconova: Mahdi I think you had implemented these lines but they are actually never used. In DetectorConfigs, isCombinedPMTCollectionDefined=true. So then in StackingAction the code uses GetPMTQE instead of GetStackingPMTQE. So the WCSimBasicPMT object is then never used (in WCSimPMTQE). I feel it should be safe to just remove the BasicPMT class (to which these comments refer) and all references to it.

src/WCSimSteppingAction.cc
mahditaani: Yeah this process is no longer required but I will remove them in the next pull request.

src/WCSimWLSProperties.cc
mahditaani: You're right. Good spot.

src/WCSimWLSProperties.cc
mahditaani: You're right. Good spot.

macros/OD.mac
mahditaani: There are already two copies of each command underneath the tank and od parts. I guess the comments are confusing though.

mahditaani: I have modified this for clarity. See c2f0c1aa25d9c8886fb1674f386ccca2a4705f16

sample-root-scripts/ODAnalysis.C
mahditaani: You are right. I have removed the redundant macro.

WCSim.cc
```diff
@@ -1,3 +1,4 @@
+#include <TRandom3.h>
```
tdealtry:
```suggestion
```
Not required

macros/OD.mac
tdealtry: Ok that makes sense if that's the current baseline

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -111,6 +174,7 @@ WCSimPrimaryGeneratorAction::~WCSimPrimaryGeneratorAction()
              << " = " << _counterRock/(G4double)_counterCublic << G4endl;
   }
   inputFile.close();
+  inputCosmicsFile.close();
   delete particleGun;
   delete MyGPS;   //T. Akiri: Delete the GPS variable
   delete messenger;
```
tdealtry, line 180:
```suggestion
  delete messenger;
  delete hFluxCosmics;
  delete hEmeanCosmics;
```
Noitced a couple of histograms that should be cleared up in the destructor

src/WCSimRunAction.cc
tdealtry: Ok so I just run 10 events with OD.mac, and this is huge!
- Photon tree: 38 MB
- wcsimT tree: 1 MB

So we defintiely need to switch this off by default. I'd suggest something like 
- add a `bool` option to `WCSimRunActionMessenger` called `SaveWLSPhotonInfo`
- only create the `TTree` here if that's `true`
- only `Fill()` the tree in `WCSimSteppingAction.cc` if the tree is not `null`

It'd also be good to improve the description e.g.
```suggestion
  photonTree = new TTree("photons","Photons in WLS Tree");
```

src/WCSimEventAction.cc
```diff
@@ -1002,7 +999,7 @@ void WCSimEventAction::FillRootEvent(G4int event_id,
 	pdir[l]=mom[l];        // momentum-vector 
 	stop[l]=Stop[l]/cm; // stopping point 
 	start[l]=Start[l]/cm; // starting point 
-	//G4cout<<"part 2 start["<<l<<"]: "<< start[l] <<G4endl;
+	G4cout<<"part 2 start["<<l<<"]: "<< start[l] <<G4endl;
```
tdealtry: Please comment this out again (or any other way where it's off by default). It's very verbose!
```suggestion
	//G4cout<<"part 2 start["<<l<<"]: "<< start[l] <<G4endl;
```

src/WCSimEventAction.cc
```diff
@@ -880,7 +877,7 @@ void WCSimEventAction::FillRootEvent(G4int event_id,
       pdir[l]=jhfNtuple.pdir[k][l];
       stop[l]=jhfNtuple.stop[k][l];
       start[l]=jhfNtuple.start[k][l];
-      //G4cout<< "start[" << k << "][" << l <<"]: "<< jhfNtuple.start[k][l] <<G4endl;
+	G4cout<< "start[" << k << "][" << l <<"]: "<< jhfNtuple.start[k][l] <<G4endl;
```
tdealtry: Please comment this out again (or any other way where it's off by default). It's very verbose!
```suggestion
	//G4cout<< "start[" << k << "][" << l <<"]: "<< jhfNtuple.start[k][l] <<G4endl;
```

src/WCSimPMTObject.cc
tdealtry: Ok thanks for this. Removing the `WCSimBasicPMTObject` class sounds great

tdealtry (changes requested):
Thanks for all this @mahditaani 
I've had a skim through it all again, and just a few comments left (most trivial, couple are a bit more work)
Almost time to get this PR merged in!

src/WCSimPrimaryGeneratorAction.cc
mahditaani: Good spot! done, see 4fc0e54e2a93c7acb5fb2ac65c143299335982fc

src/WCSimRunAction.cc
mahditaani: @tdealtry thanks for your suggestions but actually this feature is no longer needed. It was initially used to create the wavelength shifting process but we have now moved over to a completely geant4 implementation. As such I have removed this and other unused code in 78f458ebf883fc3a6652906b4315fa3c4cfcbaf7 .

tdealtry:
Thanks for all this work @mahditaani and the rest of the OD group!

PR #307, opened 2020-12-21
Update for cmake compilation with ROOT6 (+ scripts to compile on sukap)

gpronost:
The current CMakeLists.txt is not compatible with ROOT6, I updated it so it is compatible with both ROO5.34 and ROOT6. In addition, I updated the old scripts available to compile on sukap and outside sukap.


env_sukap_ROOT5.sh
```diff
@@ -0,0 +1,35 @@
...
+############################
+# ROOT/CERN
+############################
+# Note: official ROOT 5.34 installation on sukap is not compatible with cmake
+#  Temporary use Guillaume Pronost's version. Should change in the future.
+#
+export ROOT_DIR=/home/pronost/software/root-5.34.38-build
+# Note: There is no official ROOT 6 installation 
+#  Temporary use Guillaume Pronost's version. Should change in the future.
+#export ROOT_DIR=/home/pronost/software/root-6.22.00-build
```
tdealtry:
```suggestion
```

env_sukap_ROOT6.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+############################
+# ROOT/CERN
+############################
+# Note: official ROOT 5.34 installation on sukap is not compatible with cmake
+#  Temporary use Guillaume Pronost's version. Should change in the future.
+#
+#export ROOT_DIR=/home/pronost/software/root-5.34.38-build
```
tdealtry:
```suggestion
```

tdealtry (changes requested):
Thanks for fixing this @gpronost (I'm still a `make` rather than `cmake` user, so not run into this yet)

I do have one comment. I'd prefer putting some of the shell scripts into a new folder e.g. `setup/` (definitely `SetupGitHub.sh`, `env_sukap*.sh`, `make_sukap*.sh`, potentially `make.sh` too), just to keep things a bit cleaner in the root directory


Also, a couple of thoughts on merging the `make*.sh` files (why am I being pedantic? It's the usual of having 3 files doing (almost) the same thing - they will go out of sync eventually)
First, to merge the sukap files
- If you set a variable like `$root_version_dir` in your `env_sukap` you can use it in the `build_directory` construction
- If you enforce a workflow of `source env.sh; make.sh` rather than just `make.sh`

Then to merge with the main `make.sh` it'd be
- In `env_sukap` add the `cmake` bin to the users path e.g. `PATH=/usr/local/cmake-3.18.1/bin/:$PATH` to make 
- Then you can just use `cmake` without the absolute path in `make.sh`

Then you could go a step further, and add an argument to the `env_sukap*.sh` that means you can switch between ROOT5/6

I don't mind if you use these suggestions or not (since I'm not a sukap user) but it's worth thinking about it

gpronost:
Hi Tom, I followed your comments and moved the scripts to the setup/ directory. I also merge the sukap scripts and the make scripts.

One point however: I kept the make script in the root directory. If moved, we will have to consider how to be sure  the script is executed from the correct directory. In addition, I think users will then see it more clearly if it stay there,

gpronost:
I also included the modifications made by Nick on WCSim hybrid for the CMakeList, there was a bug happening in some cases.

tdealtry (approved):
Looks great. Thanks @gpronost I'll merge when travis goes green

PR #312, opened 2021-04-30
[FIX] ROOT/CERN Randomizer initialization

gpronost:
Hi,

This is a fix it seems I forget to commit before: the issue is that ROOT/CERN Randomizer (used in the Rn generator for example) is not initialized properly and always keep the same value. Here, I set it to be the same value as the Geant4 Randomizer seed, allowing to change it at the same time, and so to have correct randomization (while keeping the possibility to reproduce the previous results).


tdealtry:
Thanks @gpronost

PR #318, opened 2021-09-02
Update HK tank size to HK TR numbers

tdealtry:
The HK ID size is out of date. This PR updates the geometry to match the TR

Note the validation will fail, since the geometry has changed. Will discuss this at the next HK software meeting, then, if approved, will remake the reference files & merge this in


tdealtry:
Closing this after #399

PR #325, opened 2021-12-29
Fix trivial compiler warnings

spradlin:
Eliminates the causes of most trivial compiler warnings when compiled with gcc 4.8.5, and fixes a couple of bugs indicated by some non-trivial compiler warnings.

See notes in spradlin/WCSim#9.

Most of the compiler warnings are variable name clashes, which were addressed by renaming one of the conflicting objects.

Bugs addressed include
- An incorrect use of the G4Colour rgb constructor in `src/WCSimConstructPMT.cc`, and
- Incomplete implementation of inline functions in `include/WCSimRunAction.hh` that allow termination without a return value

The remaining compiler warnings from `include/WCSimTrajectory.hh` indicate an unimplemented feature in `WCSimTrajectory::DrawTrajectory()` that is assumed by invocation in `WCSimEventAction::EndOfEventAction()`.

The remaining compiler warnings from `src/WCSimPrimaryGeneratorAction.cc` indicate potential mismatches in the cosmic muon flux between the version read from a file and the internal representation as a pair of `TH2D`s.

### DIFF
--- a/include/WCSimPMTObject.hh
+++ b/include/WCSimPMTObject.hh
@@ -299,27 +299,27 @@ class WCSimBasicPMTObject
   ~WCSimBasicPMTObject();
 
  private:
-  std::vector<G4double> QE;
-  std::vector<G4double> wavelength;
-  std::map<G4double,G4double> mapQE;
-  G4double  maxQE;
-  TGraph   *gQE = nullptr;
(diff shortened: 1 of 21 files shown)


spradlin:
Obviated by merge of https://github.com/WCSim/WCSim/pull/347.

PR #332, opened 2022-03-22
add comment to OD.mac

federiconova:
added comment in OD.mac file to recommend using sensitive detector option


tdealtry:
Thanks @federiconova

PR #336, opened 2022-07-12
Migrate to Geant4.10.2+

tdealtry:
In order to use any version of Geant4 since 4.10.2, a minor change is required to the WCSim code due to a method argument change
If you are simulating radioactivity, 4.10.3 is the minimum version you should use, due to known bugs in previous versions
The nuPRISM version already uses the new version

The sukap setup script has also been modified to use Geant4.10.5.p01, simply because this is a version that is already installed on sukap. I must warn that changes to the output of WCSim are only known to be small between Geant 4.10.1 & 4.10.3. There may be large changes between Geant 4.10.3 & 4.10.5

I will let this PR sit for a while, as it concerns changing the recommended version of Geant4, which is not a trivial change. A new version will be tagged too

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,9 +36,11 @@ More detailed information about the simulation is available in
 ## Current notes and how to build
 
 You should have a recent and working version of ROOT and GEANT4.
-(Known to work with GEANT 4.10.1p03 and ROOT v5.28.00)  You also need all of the G4
-data files including hadron xsecs etc.  Those are the only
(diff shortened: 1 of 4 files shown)


tdealtry:
Build failed in CI as expected - need to modify the docker base container with a newer Geant4 version

tdealtry:
Closing this as most things have been merged in

Deciding to keep G4.10.3 for WCSim's sukap setup script. Newer versions of G4 will be considered for GHOST

PR #342, opened 2022-09-02
Installation fixes

guiguem:
The jtNtuple.h wasn't copied during installation
Moved the WCSimConfig.cmake.in under make as suggested by Tom (thought I did this already...)


CMakeLists.txt
```diff
@@ -125,7 +125,7 @@ add_subdirectory (src)
 
 
 
-configure_file(${PROJECT_SOURCE_DIR}/WCSimConfig.cmake.in ${CMAKE_CURRENT_BINARY_DIR}/WCSimConfig.cmake @ONLY)
+configure_file(cmake/WCSimConfig.cmake.in ${CMAKE_CURRENT_BINARY_DIR}/WCSimConfig.cmake @ONLY)
```
```suggestion
configure_file(${PROJECT_SOURCE_DIR}/cmake/WCSimConfig.cmake.in ${CMAKE_CURRENT_BINARY_DIR}/WCSimConfig.cmake @ONLY)
```
Would it be better or worse to use the full path, rather than partial path here? (I don't know the answer, you're the CMake expert!)

tdealtry (commented):
Looks great. Thanks @guiguem. Just one question from me - let me know and I can merge

guiguem:
Hi Tom, yes thanks this is better!

PR #353, opened 2023-06-12
Merge WCTE to WCSim_v1.11.0

kmtsui:
Merge WCTE/WCSim into the latest WCSim/WCSim

~~Still need to modify the NoReplica construction method to include hybrid config and OD.~~

PR #355, opened 2023-07-08
IWCD update geom

kmtsui:
New IWCD geometry with updated detector dimension, and one geometry with OD

Also two debug:
1. Add `G4LogicalSkinSurface` for blacksheet
2. Reduce loop time for find trigger algorithm


pdeperio:
slides presented at [HK IWCD General Meeting](https://hkdbweb.in2p3.fr/doc/meetingfiles/A00011543/IWCD_geom_update_20230713.pdf)

tdealtry:
I'm looking at this, but tangentially - getting the CI working properly. I hope to have things working smoothly mid next week

kmtsui:
We are getting a new [feature](https://github.com/kmtsui/WCSim/pull/1) of storing the trajectory points for tracks that cross the ID/OD boundaries. Maybe will issue a new PR after this one is finished?

kmtsui:
BTW, there is gap between the OD barrel tyvek and OD cap tyvek. Is this expected or a bug?

tdealtry:
The CI is working now, mostly fine. But unfortunately I can't see how to get your branch using the new CI file myself. Can you do something like a `git pull && git push` (I think only changes are to `.github/workflows`, so there shouldn't be any conflicts)

tdealtry:
> BTW, there is gap between the OD barrel tyvek and OD cap tyvek. Is this expected or a bug?

I found this issue in the HK FD, and @spradlin fixed it. So it is unexpected in the current code...
[Although technically I found that there was a gap that let light through. It may be that not every piece was sealed up? If you could produce some visualisation you can post here, for both IWCD & HKFD geometries that'd be great - I've not worked out a setup that lets me do this yet...]

kmtsui:
image

This is a detector visualization for IWCD, where you can see a gap between the barrel inner tyvek (yellow) and the top tyvek.

At [here](https://github.com/kmtsui/WCSim/blob/IWCD/update_geom/src/WCSimConstructCylinder.cc#L1058), the cap is just defined as tubs which does not match the [barrel polyhedra](https://github.com/kmtsui/WCSim/blob/IWCD/update_geom/src/WCSimConstructCylinder.cc#L2175). Also [RingZ](https://github.com/kmtsui/WCSim/blob/IWCD/update_geom/src/WCSimConstructCylinder.cc#L2181) only goes until blacksheet not the OD cap position.

spradlin:
@tdealtry The geometry error that i fixed was a gap in the ID blacksheet.  The ID blacksheet should be light-tight.  If the OD tyvek is also supposed to seal, then this is definitely an error, and one that would have been missed by previous checks.

spradlin:
Have the new geometries been checked with `WCSim_Geometry_Overlaps_CHECK=ON`?  In local testing, i am seeing some ovelaps and a trip of one of the 'IMPOSSIBLE GEOMETRY' checks for both IWCD_mPMT and IWCD_mPMT_WithOD.

tdealtry:
Ahh so yes this is a thing that needs fixing for all OD geometries. Good spot! It would be very helpful if you can attempt a fix @kmtsui & share visualisations of both the IWCD & HK FD with OD geometries

Good point about the `WCSim_Geometry_Overlaps_CHECK=ON`. I'll add this to the CI

kmtsui:
Yes I am aware of the `IMPOSSIBLE GEOMETRY` warning from the barrel annulus ring blacksheet. It doesn't cause any fatal error so I just let it lives. 

In https://github.com/WCSim/WCSim/pull/353, I did some extra checks to ensure a sequential z arrays (e.g. [here](https://github.com/kmtsui/WCSim/blob/merge_WCTE_to_v1.11.0/src/WCSimConstructCylinder.cc#L1507)). Do you have other suggestions to resolve the warning/bug?

spradlin:
> In #353, I did some extra checks to ensure a sequential z arrays (e.g. [here](https://github.com/kmtsui/WCSim/blob/merge_WCTE_to_v1.11.0/src/WCSimConstructCylinder.cc#L1507)). Do you have other suggestions to resolve the warning/bug?

Hmm, IMPOSSIBLE GEOMETRY is a warning that the overall dimensions are incompatible.  It should be fixed by making the definition in WCSimDetectorConfigs.cc self-consistent.  The kind of dimension tampering in https://github.com/kmtsui/WCSim/blob/merge_WCTE_to_v1.11.0/src/WCSimConstructCylinder.cc#L1507 usually has unintended consequences that cannot fix an impossible geometry definition.

The way that the parts of the endcap assembly are intended to fit are shown in an attachment to https://github.com/WCSim/WCSim/pull/137, https://github.com/WCSim/WCSim/files/60297/OverlapBug.pdf.

spradlin:
I think that the issue is that one of the historical 1mm fudge factors needs to be taken into account.  See the 1mm subtracted in the [computation of `innerAnnulusRadius`](https://github.com/kmtsui/WCSim/blob/faa1b20d1c4d88b5034ebb981b75951aad2fc07a/src/WCSimConstructCylinder.cc#L99-L110).

Adding this 1mm offset to [`WCBarrelPMTOffset`](https://github.com/kmtsui/WCSim/blob/faa1b20d1c4d88b5034ebb981b75951aad2fc07a/src/WCSimDetectorConfigs.cc#L1247) corrects both the IMPOSSIBLE GEOMETRY warning and two of the three overlaps in my local testing.

I am sorry that this factor is something that was introduced after the diagrams in https://github.com/WCSim/WCSim/pull/137.

spradlin:
It looks like the overlap that i am seeing between `WCPMTsupport2` and its containing volume arise because
[the dimensions of `WCPMTsupport2` are hardcoded](https://github.com/kmtsui/WCSim/blob/faa1b20d1c4d88b5034ebb981b75951aad2fc07a/src/WCSimConstructPMT.cc#L415-L417)!

kmtsui:
> I think that the issue is that one of the historical 1mm fudge factors needs to be taken into account. See the 1mm subtracted in the [computation of innerAnnulusRadius](https://github.com/kmtsui/WCSim/blob/faa1b20d1c4d88b5034ebb981b75951aad2fc07a/src/WCSimConstructCylinder.cc#L99-L110).

I arrived at the same confusion. Adding the 1mm fudge factor should solve the problem.

kmtsui:
> It looks like the overlap that i am seeing between WCPMTsupport2 and its containing volume arise because
[the dimensions of WCPMTsupport2 are hardcoded](https://github.com/kmtsui/WCSim/blob/faa1b20d1c4d88b5034ebb981b75951aad2fc07a/src/WCSimConstructPMT.cc#L415-L417)!

We tried to resolve the reflector overlap warning by redefining the geometry in https://github.com/WCSim/WCSim/pull/353
Now the warning should be gone, but I agree we should have a better automation/construction process.

kmtsui:
> Ahh so yes this is a thing that needs fixing for all OD geometries. Good spot! It would be very helpful if you can attempt a fix @kmtsui & share visualisations of both the IWCD & HK FD with OD geometries

In commit https://github.com/WCSim/WCSim/pull/355/commits/f8b1773aac43f35fa30857e3a51bf2e7395495ae, the cap tube is turned into polygon, and the missing tyvek in the Top/BotCapAssembly is added back. The code is a bit clumsy so any suggestion is welcome.

IWCD visualization (ID PMTs are not placed to speed up)
image
HK visualization
image

tdealtry:
This looks great. Thank you!

I would say that it is essential to not have any overlap warnings, as whilst you can ignore overlaps, overlaps will mess up the physics response in some undefined way (could be benign, could be horrible). So thank you for fixing these warnings

What is running of the CI looks to be running successfully. The two caveats are
- anything with an mPMT in isn't working. This is because this line is missing from your code https://github.com/WCSim/WCSim/blob/develop/cmake/this_wcsim.sh.in#L8 (it means it can't find the mPMT file `mPMTconfig_19_nuPrism_3ring.txt`) So please fix that line (or better yet, pull in everything from develop - as I mentioned, it's mostly just the CI files)
- the text file comparisons are complaining, but I think that's just a logic issue (reporting failed when not). I've just fixed that, so should report success successfully next time it's run

Sorry the CI is still in the teething stage. We're getting there though

tdealtry:
I've checked the code, it looks good. And the "failing CI" is for known reasons
- You've fixed the stuck track warnings "Optical photon is killed because of missing refractive index". I'm very grateful for this!
  - There is a small increase in "GeomNav1002" warnings in the SK geometry. For example [here](https://wcsim.github.io/Validation/b794d713f5b811dca289e2e1d81253be7e8c547d/15/wcsim_sk_SensitiveDetector_Only_NoTrigger_gun500MeVem_xdir_dark0_n1000_bad.diff.txt) but up to 10% more (~30) of those warnings for a reduction of 10,000s of the others is a small price to pay
- There are geometry overlap warnings in the old nuPRISM geometry. Which is fine, because that is superceeded. I'll produce a new set of references today

PR #366, opened 2023-08-03
Revert "Revert "IWCD update geom""

tdealtry:
Reverts WCSim/WCSim#364

The CI is finally working, so lets revert the revert, to restore the #355 PR

### DIFF
--- a/WCSim.mac
+++ b/WCSim.mac
@@ -35,6 +35,13 @@
 #/control/execute macros/mPMT_nuPrism1.mac
 #/control/execute macros/mPMT_nuPrism2.mac
 
+# This is the updated IWCD geometry as of 20230630
+#/WCSim/WCgeom IWCD_mPMT
+#/WCSim/Construct
+
+# This is the updated IWCD geometry as of 20230630, plus OD
+#/WCSim/WCgeom IWCD_mPMT_WithOD
+#/WCSim/Construct
 
 #Note that whilst you can change the number of PMTs
 # in the HK FD ID using these photocoverage parameters,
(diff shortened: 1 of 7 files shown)


tdealtry:
Ok so despite everything being red, I'm happy to accept it

- You've fixed the stuck track warnings "Optical photon is killed because of missing refractive index". I'm very grateful for this!
  - There is a small increase in "GeomNav1002" warnings in the SK geometry. For example [here](https://wcsim.github.io/Validation/615ed33681ff98bee24903493b453fc9c76244fa/15/wcsim_sk_SensitiveDetector_Only_NoTrigger_gun500MeVem_xdir_dark0_n1000_bad.diff.txt) but up to 10% more (~30) of those warnings for a reduction of 10,000s of the others is a small price to pay
- There are geometry overlap warnings in the old nuPRISM geometry. Which is fine, because that is superceeded.
- Whilst there are differences in number of tracks/hits/digits
  - The blacksheet skin being added fixes the optical photon warnings, so must also effect the number of photons(->hits->digits)
  - These are mostly at the ~1% level, traditionally what we use to say things are consistent
  - Things are more different with the "ODbarrelIn" particle guns for the 20" & mPMTs seen [here](https://wcsim.github.io/Validation/615ed33681ff98bee24903493b453fc9c76244fa/index.html). But we know that these weren't perfect before (difference in hit rate when starting at +x or +y) and despite the means changing, the assymettries are actually slightly reduced. So another win.

Thank you very much @kmtsui. Sorry again for the delay

I'll tag to produce a new set of references today, using the new IWCD geometries

PR #371, opened 2023-08-24
Resolve linkage issue

guiguem:
It seems that in more modern compilers, the PROJECT_LIBRARIES variables is not set properly. So we set it more explicitely here


tdealtry:
Thanks!

PR #374, opened 2023-09-07
Rearrange optional compilation flag

gondiaz:
To install WCSim dependent packages the `WCSIM_CHECK_GEOMETRY_OVERLAPS` need to be accesible.
This is done by setting the variable in the `WCSimConfig.cmake` via its `WCSimConfig.cmake.in` template when WCSim is installed.
To do that, we need to move the definition of `WCSim_Geometry_Overlaps_CHECK` to the higher hierarchy CMakeLists file from the its previous definition at the `src` subdirectory.


src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -123,53 +123,55 @@ WCSimPrimaryGeneratorAction::WCSimPrimaryGeneratorAction(
...
-      cosThetaMean=(atof(token[2]));
-      cosThetaMin=(atof(token[3]));
-      cosThetaMax=(atof(token[4]));
-      phiMean=(atof(token[5]));
-      phiMin=(atof(token[6]));
-      phiMax=(atof(token[7]));
-      */
-      flux=(atof(token[8]));
-      Emean=(atof(token[9]));
+  if (useCosmics) {
```
guiguem: Do we agree this will be not run all the time (unless one set the useCosmics variable to true)?

CMakeLists.txt
```diff
@@ -90,9 +90,18 @@ if(WCSim_DEBUG_COMPILE_FLAG STREQUAL ON)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g")
 endif()
 
+##############
+# WCSim compilation options
+##############
+# Cution: setting a new default value to this variable requires the recompilation of the full package
```
```suggestion
# Caution: setting a new default value to this variable requires the recompilation of the full package
```
Typo

src/WCSimPrimaryGeneratorAction.cc
tdealtry: Good spot @guiguem In practice this is never true, because `useCosmics` is by default `false` and only ever becomes true via the messenger (mac command `/mygen/generator cosmics`), which happens **after** this constructor

I think this is a nice idea (I'm always confused as to why it produces the extra root file), but this block needs to be moved somewhere else. For example to, `GeneratePrimaries()` (or better, a function called from there, because `GeneratePrimaries()` is already a massive unreadable function...) with a flag to say whether this block has been run (so it only gets run once)

CMakeLists.txt
```diff
@@ -152,6 +161,23 @@ pbuilder_install_files(${CMAKE_INSTALL_PREFIX}/bin/rootwc-files
...
+        ${PROJECT_SOURCE_DIR}/mPMT-configfiles/mPMTconfig_33_50.txt
+        ${PROJECT_SOURCE_DIR}/mPMT-configfiles/mPMTconfig_33_50_1.txt
+        ${PROJECT_SOURCE_DIR}/mPMT-configfiles/mPMTconfig_34_22.5_1.txt
+        ${PROJECT_SOURCE_DIR}/mPMT-configfiles/mPMTconfig_34_22.5_2.txt
+        ${PROJECT_SOURCE_DIR}/mPMT-configfiles/mPMTconfig_34_22_1.txt
+        ${PROJECT_SOURCE_DIR}/mPMT-configfiles/mPMTconfig_34_22_2.txt
+        ${PROJECT_SOURCE_DIR}/mPMT-configfiles/mPMTconfig_37_HK.txt
+        ${PROJECT_SOURCE_DIR}/mPMT-configfiles/mPMTconfig_Position_WCTE.txt
+        ${PROJECT_SOURCE_DIR}/mPMT-configfiles/mPMTconfig_Position_WCTE_RotateBarrelHalfTower.txt
+        )
```
tdealtry, line 187: General question for the CMakeLists.txt @guiguem Is there a reason why installing files is done on a per file basis, rather than just installing everything in a folder? The latter is possible - see for example [this answer](https://stackoverflow.com/a/67883070)

tdealtry (changes requested):
Thanks for this @gondiaz Just a few comments

CMakeLists.txt
guiguem: Right, I don't have a strong preference, I tend to be explicit about the files we install, so we are conscious about the potential clustering we are putting (this also allows to remove a file more easily that using pattern-based file installation; this is especially true for source and header file; maybe here for config file, this is not something we want to do). But this is really a personal preference; I am fine if we use something else!

tdealtry: Yeah this is a tricky case for config files. If people forget to add/remove them for `CMakeLists.txt` for compiled code, it's obvious. Not so for config files. But as you say, it's easier to add/remove them as required if done explicitly like this.

Happy to keep it as it is for WCSim [mostly because I haven't formed a preference yet], and maybe think again for GHOST & other Tool packages

guiguem:
I just added the WCSimCore library into the list of produced libraries. This should ease the importation and linkage of these two libraries into subsequent libraries via CMake

tdealtry:
Thanks for the new commits @gondiaz. From your side, are there any extra things you want to put into this PR?

From my side, there's 3 outstanding things I'd like before I can merge this (don't worry, 2 are trivial)
- Typo in CMakeLists.txt. See the suggestion above
- Can you add `.vscode/` to the `.gitignore`? This will stop people accidentally commiting such files that you're removing again
- I've verifyed that the cosmics generator is broken, because the cosmics file is never loaded & the histograms never created (i.e. the `if (useCosmics)` block is never run). See the conversation above for why, and for a suggested way around this

src/WCSimPrimaryGeneratorAction.cc
gondiaz: Thanks @guiguem and @tdealtry for the catch. 
I added this part of the code into the `GeneratePrimaries()` throught the `Create_cosmics_histogram` function.

gondiaz:
Done, as for the `.vscode` directory it was already included in the `.gitignore`.

tdealtry (approved):

tdealtry:
The auto checks are happy, and I verfied that the cosmics are happy, so I'm happy. Thanks for this @gondiaz!

PR #375, opened 2023-09-11
Proper numbering and orientations of PMTs in mPMT for WCTE

kmtsui:
To get the numbering and orientations correct for WCTE mPMTs. May need to further change the module dimensions later.
image


tdealtry:
Thanks @kmtsui This is [modifying all mPMT geometries](https://wcsim.github.io/Validation/898dddbb62225125b22d834ce2fe43150df895b6/index.html) not just WCTE. Can you explain a bit more as to what changed and why?

I would presume the difference for all geometries is in [WCSimMultiPMTParameterisation.cc](https://github.com/WCSim/WCSim/pull/375/files#diff-100fd9a0379edae38b2d3e55408c76784f104fdc5acf3c9b34f8800c209072e6)

kmtsui:
To change the ordering, a new [mPMTconfig ](https://github.com/WCSim/WCSim/pull/375/files#diff-e36a0af1a50f4f5cb13c34e900c53233ce7492ec41fc3210bd37f5e994a49ba8) is used. But then there is an extra offset in the ring because [copy number](https://github.com/WCSim/WCSim/blob/898dddbb62225125b22d834ce2fe43150df895b6/src/WCSimMultiPMTParameterisation.cc#L70) is directly used. That why I subtracted the number of PMTs from the previous rings.

Looking at the output, the difference in geometry positions is flipping between -0.000 and 0.000, so it should be numerical precision in the subtraction. Let me change the way of subtraction.

kmtsui:
okay, the difference is still there for some geometries, probably because in the original code, phi is sometimes added over 2pi.

tdealtry:
Ok I see. Happy for you to merge it when you're happy with it, since the changes are small (< 1% level)

I just draw your attention to [this](https://wcsim.github.io/Validation/c1ec455c98d6cb5378762d91a226c1b71790bc40/23/geofile_NuPRISMBeamTest_16cShort_mPMT.txt.diff.txt) showing that the WCTE detector size has been modified by this PR. Is this expected?
(the FD/IWCD/SK geometries, thankfully, haven't changed size)

kmtsui:
Looking at `WCSimConstructGeometryTables.cc`, the detector size variables are calculated directly from the PMT positions instead of the mPMT module. Since the PMT numbering/angle are changed, these size variables are changed accordingly. Personally I prefer a better way to define the detector radius and height, maybe directly from `WCIDDiameter` and `WCIDHeight`?

Anyway I think we are good to merge it.

PR #376, opened 2023-09-14
Make having an invalid PMT position file fatal.

tdealtry:
Otherwise, people won't see the error message lost in the `cout` stream, and they'll assume everything is fine

@kmtsui can you please approve, or convince me why this PR a bad idea?


kmtsui:
Making it fatal is not a problem, but with your change I think it is not actually exiting, but just printing out error?

kmtsui (approved):

tdealtry:
Doh! This shows why PRs are better than direct commits, even for simple things... Thanks for the review!

kmtsui (approved):

PR #379, opened 2023-10-10
Datatable event generator

JackFannon:
# Datatable event generator
## Description
An event generator that takes datatable files as input:

| Index in event | Particle PDG | Energy (MeV) | x | y | z | x dir | y dir | z dir | time |
|-|-|-|-|-|-|-|-|-|-|

where particles are defined in single lines containing their PDG code, energy, position, direction, and time in the event.

## UI Commands
The following commands can be used to select the datatable generator and pass it a primary file:
```
/mygen/generator/datatable
/mygen/vecfile <FILENAME>
```

## Example
To define a positron and a neutron in a single event the datatable file would look as follows:
```
0   -11  12.345  0.000  0.000  0.000  0.000  0.000  -1.000  0.000
1  2112   3.456  0.000  0.000  0.000  1.000  0.000   0.000. 0.000
```
note that particles in the same event are not required to have the same position or direction.

## Misc
Fixed a bug in `PrimaryGeneratorAction.hh`. Some `if` statements were checking as:
```
...
 else if (choice="microsecond")
...
```
instead of:
```
...
 else if (choice=="microsecond")
...
```


src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -905,8 +907,92 @@ void WCSimPrimaryGeneratorAction::GeneratePrimaries(G4Event* anEvent)
...
+      }
+
+      // Buffer to convert between string and other variables
+      std::istringstream buffer(line);
+
+      // Load information into local variables
+      buffer >> index >> pdgid >> ene >> pos[0] >> pos[1] >> pos[2] >> dir[0] >>
+          dir[1] >> dir[2] >> t;
+
+      if (index == 0 && firstParticle == false) {
```
tdealtry, line 972:
```suggestion
      // We've finished reading all the particles constituing the event n we're generating
      // and we've also read the first line of event n+1
      // So rewind to the start of event n+1, for when we call this function again for the next event
      if (index == 0 && firstParticle == false) {
```

Just checking I understand what you're doing. And I think putting a comment in to explain it is a good idea. So please modify the comment for accuracy!

tdealtry (changes requested):
This looks good. Thanks @JackFannon! You're missing a few things though
- Need to add the option to the messenger class, so you can actually turn it on/off (maybe you forgot to commit a few changes?)
- Add an example of how to use it to WCSim.mac, commented out like the others
- In the WCSimPrimaryGeneratorAction constructor, you should initialise `useDataTableEvt = false`
- What happens when the file is exhausted before the number of requested events via `beamOn` in the .mac file is reached? See [here for how an example of how to handle it](https://github.com/WCSim/WCSim/blob/develop/src/WCSimPrimaryGeneratorAction.cc#L252)

Then there's this. I know spallation can do crazy massive things, so not sure if specifically not filling this information is actually best. And anyway, each primary track you produce *should* be saved in the output file. If this isn't filled, then it should be explicitly mentioned that it isn't filled in the .mac file _documentation_
- You're not filling some information e.g. `vtxs`, so the e.g. vertex information won't be available in the output WCSimRootEvent
- What happens with massive events? WCSim has some fixed length arrays (yes, I know...) that will fail. See [here for an example of how to handle it](https://github.com/WCSim/WCSim/blob/develop/src/WCSimPrimaryGeneratorAction.cc#L377) (actually probably better to exit rather than die, so people can't not notice it)

tdealtry:
Can you also do a git pull from WCSim/WCSim develop? The validation is failing, and it looks similar to how #375 failed (the validation reference files were updated after that merge)

(Why this is happening? I guess the CI isn't setup to get the fully merged result of the PR into develop. Not sure how to properly do that... but hopefully asking you to do a `git pull` isn't too much to ask)

src/WCSimPrimaryGeneratorAction.cc
JackFannon: Hi @tdealtry! Thanks for the feedback and help.

Yeah, you were right. I was missing a couple of commits that actually contained the definitions in the messenger class and a comment explaining the rewind behaviour in the generator.

I've added everything else: check for if the generator reaches the end of the datatable before `BeamOn` is reached, check for the maximum number of vertices etc.

I think that I've also sorted out setting the information that's written to the output file, though I may have missed some. I'd appreciate it if you could double-check that if you get the chance, please.

Also, sorry the commits are such a mess; I've got an auto-formatter installed that had made **many** _**many**_ changes to a lot of files in the same commits as actual work.

The pull from `WCSim/WCSim/develop` has also been done and _should_ be sorted now.

Let me know if there are any other issues!

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -910,8 +913,116 @@ void WCSimPrimaryGeneratorAction::GeneratePrimaries(G4Event* anEvent)
...
+      lastLinePos = inputFile.tellg();
+
+      // Set values to save to output
+      SetBeamEnergy(ene, nParticles);
+      SetBeamDir(dir, nParticles);
+      SetBeamPDG(pdgid, nParticles);
+      SetVtxs(nParticles, pos);
+
+      nParticles++;
+
```
tdealtry, line 988:
```suggestion
SetNvtxs(nParticles);
```
I think this should be here, so if the last line of the file is read as an event you want to generate, `nvtxs` is set correctly

Am I getting the logic right?

WCSim.mac
```diff
@@ -266,6 +266,13 @@
 #/gps/direction 1 0 0
 #/gps/position 0 0 0
 
+## Or you can use the datatable option
+## This is similar to the nuance format option, though this uses datatable formatted files.
+## Datatables are formatted as follows:
+## | Index of particle in event | PDG code | Energy | X | Y | Z | Px | Py | Pz | Time |
```
tdealtry, line 272:
```suggestion
## | Index of particle in event | PDG code | Energy | X | Y | Z | Px | Py | Pz | Time |
## Note that the column separators "|" are for clarity in describing the format, and should not be included in the file
## Note also that comments are allowed in the data table file by starting lines with a "#"
```

tdealtry (changes requested):
Thanks @JackFannon! Only a couple of minor things left

P.S. Well done for spotting that time unit choice bug!

src/WCSimPrimaryGeneratorAction.cc
JackFannon: I see what you mean, though would this not call `SetNvtxs(nParticles)` every time the loop is run? So it would be called for every particle in the event.

I think it might actually make better sense outside of the while loop after all the particles have been looped over. This would then catch all of the events as long as `MAX_N_VERTICES` isn't reached.

tdealtry: Yeah that's a better fix thanks

tdealtry (approved):

tdealtry:
Thanks for the contrubution @JackFannon

PR #380, opened 2023-10-10
Adds alternative LI generator to read profiles from db

ekneale:
LI generator which reads in injector parameters by injector ID from json db file for collimator, diffuser and laser ball simulation. 

Placeholder light injectors (esp. OD collimators and diffusers) in data/LightInjectors.json

Example usage: macros/LIGen.mac
Example for reading output sample-root-scripts/read_LIGen_OD_output.C

For OD light injectors, it's currently recommended to comment out #define ACTIVATE_IDPMTS in src/WCSimConstructCylinder.cc (line 71), pending a change to the existing geometry.


macros/LIGen.mac
```diff
@@ -0,0 +1,116 @@
...
+/WCSim/SavePi0 false
+
+#choose the Trigger & Digitizer type (and options)
+/DAQ/Digitizer SKI
+#/DAQ/Trigger NDigits
+/DAQ/Trigger NoTrigger
+#/DAQ/RelativeHitTime true
+
+#grab the other DAQ options (thresholds, timing windows, etc.)
+/control/execute /data2/kneale/hk-software/wcsim_od_calib/macros/daq.mac
```
tdealtry: General comment for this file - can you remove the personal paths? The preference is to assume you're running in `$WCSIMDIR` or `$WCSIM_BUILD_DIR`

macros/LIGen.mac
```diff
@@ -0,0 +1,116 @@
...
+#/control/execute /home/kneale/hk-software/wcsim_od_calib/macros/visOGLQT-OD.mac
+#/control/execute /home/kneale/hk-software/wcsim_od_calib/macros/visRayTracer.mac
+
+
+/mygen/generator lightinjector
+/mygen/nphotons 10000
+/mygen/injectorType collimator
+/mygen/injectorIdx 1
+#/mygen/injectorFile somefile.json
+#/mygen/photonMode 1
```
tdealtry: We document via mac file comments in WCSim. Why? Tradition. So could you say something about each option. Making some specific requests too
```suggestion
/mygen/generator lightinjector
# This is nphotons per LI device? Or total?
/mygen/nphotons 10000
# Does changing this to diffuser and keeping Idx set to 1 cause the collimator to turn into a diffuser in the same position? Or is the Idx numbering for diffusers and collimators independent?
/mygen/injectorType collimator
# Explicitly say that you can only generate from 1 LI device at a time
/mygen/injectorIdx 1
#/mygen/injectorFile somefile.json
# Describe what each photon mode is
#/mygen/photonMode 1
```

macros/LIGen.mac
```diff
@@ -0,0 +1,116 @@
...
+
+## change the name of the output root file, default = wcsim.root
+
+/WCSimIO/RootFile /home/kneale/hk-software/wcsim_devel/wcsim_LIGen.root
+
+## set a timer running on WCSimRunAction
+#/WCSimIO/Timer false
+
+/run/beamOn 10000
```
tdealtry, line 128: This is the same number as nphotons. I presume this produces 10k events, each containing 10k photons (rather than 10k events, each containing 1 photon). But since the numbers are identical, best add a comment to answer the question

sample-root-scripts/read_LIGen_OD_output.C
```diff
@@ -0,0 +1,335 @@
...
+#include "WCSimRootEvent.hh"
+#include "WCSimRootGeom.hh"
+
+// Defines
+#define PI 3.141592654
+#define topPMT 19
+/*
+*How to run:
+*
+* enter in the terminal $WCSIMDIR/rootwc/rootwc -l 'read_OD.C("WCSim.root","outputFile.root",false)' to run the code
```
tdealtry: Please update the how to run with the correct filename

Also it would be nice to have a quick "this is what this macro does" comment

sample-root-scripts/read_LIGen_OD_output.C
```diff
@@ -0,0 +1,335 @@
...
+
+	// Get a pointer to the tree from the input file
+	TTree *wcsimTree = (TTree*) inFile->Get("wcsimT");
+
+	// Get the number of events in the tree
+	long int nEvent = wcsimTree->GetEntries();
+	if (verbosity) { std::cout << "Number of events: "<< nEvent << std::endl;}
+
+	// Create a WCSimRootEvent to put stuff from the tree in
+	WCSimRootEvent *wcsimRootID = new WCSimRootEvent();
```
tdealtry, line 76: Sorry to ask, but will be useful when ID LI's are starting to be used
Can you add the mPMT stuff too? This is saved in a third branch `wcsimrootevent2`

sample-root-scripts/read_LIGen_OD_output.C
tdealtry, line 1: - Thank you for adding an analysis macro. It's a rare treat, and something we are really lacking in general in WCSim

data/LightInjectors.json
tdealtry, line 1: - Can you add OD to the "comment" for each device? I think you're intending to use the same system for ID LI too?
- `Photon_Positions.csv` is listed. Can you add this to the data folder (& also rename it to `LI_OD_laser_ball_photon_positions.csv` or similar)

src/WCSimPrimaryGeneratorMessenger.cc
```diff
@@ -130,6 +175,10 @@ WCSimPrimaryGeneratorMessenger::~WCSimPrimaryGeneratorMessenger()
   delete injectorTimeCmd;
   delete openingAngleCmd;
   delete injectorWavelengthCmd;
+  delete lightInjectorCmd;
+  delete lightInjectorIdxCmd;
+  delete lightInjectorNPhotonsCmd;
+  delete lightInjectorFilenameCmd;
```
tdealtry, line 181:
```suggestion
  delete lightInjectorFilenameCmd;
  delete lightInjectorModeCmd;
```
Think you're missing one

src/WCSimLIGen.cc
```diff
@@ -0,0 +1,282 @@
...
+        db = wcsimdir + "LightInjectors.json";
+    }
+    else {
+        db = wcsimdir + injectorFilename;
+    }
+    
+    ifstream fJson(db.c_str());
+    
+    if (!fJson) {
+        G4cout << "LIGen: [ERROR] light injector db " << db << " not found, reverting to generic db" << G4endl;
```
tdealtry:
```suggestion
        G4cerr << "LIGen: [ERROR] light injector db " << db << " not found, reverting to generic db" << G4endl;
```

But I'm not sure this actually is "reverting" to anything, as `fJson` is read into `buffer` whether this is true or false

src/WCSimLIGen.cc
```diff
@@ -0,0 +1,282 @@
...
+    }
+    else {
+        LoadProfilePDF();
+    }
+}
+
+void WCSimLIGen::LoadPhotonList(){
+
+    string photonsFile = wcsimdir + photonsFilename;
+    cout << photonsFile << endl;
```
tdealtry: Please use `G4cout` `G4err` and `G4endl` exclusively (there can be some weird mangling of printout if these are mixed with `std::cout` etc)

src/WCSimLIGen.cc
```diff
@@ -0,0 +1,282 @@
...
+
+#include "json.hpp"
+
+#include "G4ParticleGun.hh"
+#include "G4PhysicalConstants.hh"
+#include "G4OpticalPhoton.hh"
+#include "G4RandomDirection.hh"
+#include "G4SystemOfUnits.hh"
+
+using namespace std;
```
tdealtry:
```suggestion
```
No thanks (HK coding convention recommendation. It'll help you remove the `cout`s too)

src/WCSimLIGen.cc
```diff
@@ -0,0 +1,282 @@
...
+    }
+}
+
+void WCSimLIGen::LoadPhotonList(){
+
+    string photonsFile = wcsimdir + photonsFilename;
+    cout << photonsFile << endl; 
+    ifstream photondata(photonsFile.c_str());
+    if ( ! photondata) {
+        cout << "WCSimLIGen [ERROR]: Failed to open " << photonsFilename << endl;
```
tdealtry: Again, seems like you're just carrying on, despite the file not opening

src/WCSimLIGen.cc
```diff
@@ -0,0 +1,282 @@
...
+    // myMessenger, myDetector
+    photonMode = 0;
+    wcsimdir = string(getenv("WCSIMDIR"))+"/data/";
+} 
+
+void WCSimLIGen::SetPhotonMode(G4bool photonmode){
+   
+    if (photonmode){
+        photonMode=1;
+    }
```
tdealtry, line 45:
```suggestion
    }
    else
      photonMode = 0;
```

src/WCSimLIGen.cc
```diff
@@ -0,0 +1,282 @@
...
+    cout << "Profile filled." << endl;
+}
+
+
+
+void WCSimLIGen::GeneratePhotons(G4Event* anEvent,G4int nphotons){
+
+    // Use particle gun to generate a point source
+    G4ParticleGun *myLIGun = new G4ParticleGun();
```
tdealtry: Memory leak. Can you move this to `Initalise()` and rememeber to `delete` it in the destructor (or `Finalise()`)

src/WCSimLIGen.cc
```diff
@@ -0,0 +1,282 @@
...
+    energy = PhotonEnergyFromWavelength(injectorWavelength);
+
+    if (photonMode){
+        // Get the position and direction from the photon list
+        // TODO subdivide into an equal number of photons per pulse 
+        // to limit the number of particles in G4Event
+        // for (int iphoton=0;iphoton<nphotons;iphoton++){
+        // have an index which is set in the constructor
+        // that way we can pick up the list where we left off last event
+    //    for (int iphoton=0;iphoton<myPhotons.size();iphoton++){
```
tdealtry: Has this been done? And so can be removed?
Just a warning that if you do do this, the pileup of "true hits" into "digitised hits" won't be handled correctly

src/WCSimLIGen.cc
```diff
@@ -0,0 +1,282 @@
...
+        for (int iphoton=0;iphoton<nphotons;iphoton++){
+            // Generate random time for this photon in 20 ns pulse window
+            G4double time = G4RandFlat::shoot(20.0,40.0)*ns;
```
tdealtry: Make sure to document the pulse width assumption more than just here (e.g. in the mac file)

Better may be to put this into the json? (naively I'd be suprised if the collimator/diffuser/ball have the same time profile)

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -128,6 +129,12 @@ WCSimPrimaryGeneratorAction::WCSimPrimaryGeneratorAction(
   openangle = 0.;
   wavelength = 435.;
 
+  //Light injector generator variables (LIGen)
+  LIGen = 0;
+  injectorType = "";
+  injectorIdx = "";
+  injectorFilename = "";
+
```
tdealtry, line 141: I think you're missing initialising of `photonMode = 0`

(It's fine not to initialise if it's given as an option in the mac file. But if it's not, I don't think it is correctly set to the default value. But it's always a complicated flow the messengers, so I may be wrong)

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -749,6 +756,60 @@ void WCSimPrimaryGeneratorAction::GeneratePrimaries(G4Event* anEvent)
...
+        G4ThreeVector vtx = LIGen->GetInjectorPosition();
+        G4ThreeVector dir = LIGen->GetInjectorDirection();
+        G4int pdg = 0;
+        G4double E = LIGen->GetPhotonEnergy();
+        LIGen->GeneratePhotons(anEvent,nphotons);
+
+        // Save the mc properties for validation
+        // Limit of <900 photons due to size of array in jhfntuple
+        /*
```
tdealtry: Can this block be deleted?

I would expect all LI photons you generate to be in the output file, in the track store. Is that not the case? Is there a benefit of saving this information too? (Naively I'd presume all photons have the same start point, energy, PDG... it's just the direction that's different photon by photon - and that info [is in WCSimRootTrack](https://github.com/WCSim/WCSim/blob/develop/include/WCSimRootEvent.hh#L38))

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -749,6 +756,60 @@ void WCSimPrimaryGeneratorAction::GeneratePrimaries(G4Event* anEvent)
...
     }
+
+  else if (useLightInjectorEvt)
+    {
+
+        // Initialise the light injector once per sim. 
+        // This will get LI settings (position, direction, etc) from the db
+        // and produce a 3D histogram of the LI profile.
+        if ( !LIGen ) {
+            LIGen = new WCSimLIGen();
```
tdealtry, line 772: Please add a `if(LIGen) delete LIGen` to the destructor

src/WCSimPrimaryGeneratorMessenger.cc
```diff
@@ -82,7 +82,52 @@ WCSimPrimaryGeneratorMessenger::WCSimPrimaryGeneratorMessenger(WCSimPrimaryGener
...
+  lightInjectorCmd->SetGuidance(" injectorType : collimator, diffuser or laserball");
+  lightInjectorCmd->SetParameterName("injectorType",true);
+  lightInjectorCmd->SetCandidates("collimator diffuser laserball");
+  lightInjectorCmd->SetDefaultValue("collimator");
+
+  lightInjectorIdxCmd = new G4UIcmdWithAString("/mygen/injectorIdx",this);
+  lightInjectorIdxCmd->SetGuidance("Set the ID number of the light injector you want to use");
+  lightInjectorIdxCmd->SetGuidance("[usage] /mygen/injectorIdx index");
+  lightInjectorIdxCmd->SetGuidance("see data/lightInjectors.json for indices for now");
+  lightInjectorIdxCmd->SetGuidance(" injectorIdx : 0,1,2, ... ");
```
tdealtry: Not sure what the most susinct solution is, but this made me think you can set multiple injectors to be active by using a comma separated string
Maybe `injectorIdx : INT`?

tdealtry (changes requested):
Hi @ekneale. Thanks for this.

A few comments inline. And one here
- Can you add the macro, and the data files to `CMakeLists.txt`?

src/WCSimLIGen.cc
ekneale: Following this up, it seems that at least for the diffuser and collimator the photon times are within 1 ns and on that basis there is no plan to measure the time profiles. So I have changed WCSimLIGen.cc to generate random times in a 1 ns window, and added a comment in the LIGen.mac. If the laser ball has a different time profile I will move it to the json at some point in the future.

src/WCSimLIGen.cc
ekneale: I haven't done this yet but have removed the comments for now.

sample-root-scripts/read_LIGen_OD_output.C
ekneale: I've added a minimal example (as for the ID PMTs) for now, and can look to extract more detail once we start to use ID LI if that's sufficient?

ekneale:
Hi @tdealtry, thanks for reviewing this. I believe I've now made all of the changes you requested.

macros/LIGen.mac
tdealtry:
```suggestion
/control/execute macros/daq.mac
```
Think this was lost in the flurry of comments. This is the kind of thing I meant (there's a few places in the file)

sample-root-scripts/read_LIGen_OD_output.C
tdealtry: Yeah that's great thanks

src/WCSimLIGen.cc
```diff
@@ -0,0 +1,279 @@
...
+    //myDetector = myDC;
+    //myMessenger = new WCSimLIGenMessenger(this);
+    this->Initialise();
+
+}
+
+WCSimLIGen::~WCSimLIGen(){
+
+    // things to delete
+    delete myLIGun;
```
tdealtry, line 27: Missed this (less bad) leak last time
```suggestion
    delete myLIGun;
    if(hProfile) delete hProfile;
```

src/WCSimLIGen.cc
tdealtry: Sounds good to me

tdealtry:
Thanks @ekneale Just a couple of small things left

tdealtry:
Just to explain to everyone why [the validation is breaking for OD hits](https://wcsim.github.io/Validation/68eaee839aa749d0e95473237e749e7f3e718a59/index.html) (It scared me at first)
It's your change to `WCSimEventAction` to apply the `RelativeHitTime` to OD PMTs (in the same way it is applied to ID 20" & mPMT PMTs)
This is a bug fix, so is working as expected

macros/LIGen.mac
ekneale: Argh sorry I'll fix these now.

ekneale:
Hopefully all sorted now @tdealtry

tdealtry (approved):

tdealtry:
Sorry @ekneale some conflicts to resolve after #379...

ekneale:
Hopefully good to go now @tdealtry - it seems like the failed tests are still just related to the bug fix?

tdealtry:
Yeah that's right

Thanks for the contribution @ekneale

PR #382, opened 2023-10-12
Fix for #381

### LINKED ISSUE #381: Default value of /DAQ/RelativeHitTime
Should the `RelativeHitTime` variable in `WCSimEventActionn` be given a default `false` value? 
In my local environment, if I don't use the `/DAQ/RelativeHitTime` macro command, the `RelativeHitTime` variable is given a non-zero (true) value. 
Is this behavior expected?

tdealtry:
This is a fix for #381, more in keeping in how we (mostly) handle default mac options in WCSim
@kmtsui please test - if it doesn't work, doing the fix you suggested will certainly work


kmtsui (approved):
Working well!

PR #383, opened 2023-10-13
Inverse beta decay (DSNB) generator

JackFannon:
# IBD Generator
### General usage
Generates inverse beta decay events with random neutrino energies, within the energy range provided by a flux file, weighted by the flux and cross section at that energy. Two particles are generated per event: a positron and a neutron.

Input for the generator is a flux database and a model name. See below for more information. 

An example of how to use the generator is in `WCSim.mac`.

Events are generated isotropically as this would be expected for the diffuse supernova neutrino background. 

### Database format
The file `./data/DSNBFluxes.json` contains the energy and flux vectors used in the weighting of events. The file has the following format for each model:
```
        {
            "name": "Tabrizi NH",
            "energy": [0.01, ..., 100],
            "flux": [0.00766989, ..., 2.05945e-06]
        },
```

New models should be relatively easy to add. The UI command is set to allow any input, though the IBDGen class will catch it if an unknown model is added. To add a model all that is required is an energy vector, flux vector, and name. The model is selected in the macro file using the name in the database. The energy vector and flux vector must have the same size.

I'm open to other suggestions of how to structure the database since the `.json` file contains **very** long lines and can be a pain to check syntax in, specifically commas. 

## Current implemented models
- Nakazato+15 normal hierarchy minimal oscillation
- Nakazato+15 inverted hierarchy maximal oscillation
- Horiuchi+21 normal hierarchy (highest prediction implemented so far, two other more optimistic ones are available (Kaplinghat and Totani)
- Hartmann+97 (Lowest prediction implemented so far)
- Flat spectra - All fluxes set at $1.0 \ \mathrm{cm}^{-2}\mathrm{MeV}^{-1}\mathrm{s}^{-1}$

**_Other models are available._**


JackFannon:
I'm unsure as to why this is failing some of the checks. This is up to date with the latest develop branch commits.

data/DSNBFluxes.json
tdealtry, line 1: I suggest adding some fields like `url` (meaning the place you got the data from) and `reference` (meaning e.g. paper or conference talk)
You of course know what every model is right now by the `name`. But its best to future proof this kind of data

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -955,6 +957,51 @@ void WCSimPrimaryGeneratorAction::GeneratePrimaries(G4Event* anEvent)
       	*/
       }
 
+    } else if (useIBDEvt) {
+        // Generate IBD events from an input spectrum
+        WCSimIBDGen *IBDGen = new WCSimIBDGen(myDetector);
```
tdealtry: There's a memory leak here, with this created every event and never `delete`d

I think doing something similar to the light injector code (where you have a `WCSimIBDGen *` in the class private data members, and check whether it's a `nullptr` or not to do stuff with)

src/WCSimPrimaryGeneratorMessenger.cc
```diff
@@ -82,7 +82,16 @@ WCSimPrimaryGeneratorMessenger::WCSimPrimaryGeneratorMessenger(WCSimPrimaryGener
...
+
+  ibdspectrumCmd = new G4UIcmdWithAString("/mygen/ibd_spectrum",this);
+  ibdspectrumCmd->SetGuidance("Select the IBD spectrum file.");
+  ibdspectrumCmd->SetParameterName("ibd_spectrum", true);
+  ibdspectrumCmd->SetDefaultValue("data/DSNBFluxes.json");
+
+  ibdmodelCmd = new G4UIcmdWithAString("/mygen/ibd_model",this);
+  ibdmodelCmd->SetGuidance("Select the IBD model.");
+  ibdmodelCmd->SetParameterName("ibd_model", true);
+  ibdmodelCmd->SetDefaultValue("Flat");
```
tdealtry, line 95:
```suggestion
  ibdmodelCmd->SetDefaultValue("Flat");
  SetNewValue(ibdmodelCmd, "Flat");
```

src/WCSimPrimaryGeneratorMessenger.cc
```diff
@@ -82,7 +82,16 @@ WCSimPrimaryGeneratorMessenger::WCSimPrimaryGeneratorMessenger(WCSimPrimaryGener
   injectorWavelengthCmd->SetGuidance("Wavelength of the injector laser in nm");
   injectorWavelengthCmd->SetParameterName("injector_wavelength",true);
   injectorWavelengthCmd->SetDefaultValue(435.);
-   
+
+  ibdspectrumCmd = new G4UIcmdWithAString("/mygen/ibd_spectrum",this);
+  ibdspectrumCmd->SetGuidance("Select the IBD spectrum file.");
+  ibdspectrumCmd->SetParameterName("ibd_spectrum", true);
+  ibdspectrumCmd->SetDefaultValue("data/DSNBFluxes.json");
```
tdealtry: I don't know what the purpose of `SetDefaultValue()` actually is, but it doesn't do what you think it does. In order to set it, you need this extra call
```suggestion
  ibdspectrumCmd->SetDefaultValue("data/DSNBFluxes.json");
  SetNewValue(ibdspectrumCmd, "data/DSNBFluxes.json");
```

src/WCSimIBDGen.cc
```diff
@@ -0,0 +1,258 @@
...
+    for (const auto &model : data["models"]) {
+        if (model["name"].get<string>() == model_name) {
+            G4cout << "IBDGen: [INFO] using model " << model_name << G4endl;
+
+            energy = model["energy"].get<std::vector<float>>();
+            flux = model["flux"].get<std::vector<float>>();
+
+            G4cout << "IBDGen: [INFO] spectrum read from database" << G4endl;
+
+            // Check if the flux vector and energy vector have the same size
```
tdealtry, line 60: You should add another check, that the energy is monotonically increasing

src/WCSimIBDGen.cc
```diff
@@ -0,0 +1,258 @@
...
+        }
+    }
+
+    // If energy is larger than the maximum energy in the spectrum, return the final flux value
+    return flux_vec.back();
+}
+
+G4ThreeVector WCSimIBDGen::GenRandomPosition() {
+    // Generate random neutrino position
+    // Pick random position in detector
```
tdealtry: Can you be clear about what "detector" means here? Is it the cuboidal world volume? The cylindrical extent of the tank? Just inside the ID? Something else?

src/WCSimIBDGen.cc
```diff
@@ -0,0 +1,258 @@
...
+    // Generate a random number between -1 and 1
+    double x_nu = myDetector->GetGeo_Dm(0) * (-1.0 + 2.0 * G4UniformRand());
+    double y_nu = myDetector->GetGeo_Dm(1) * (-1.0 + 2.0 * G4UniformRand());
```
tdealtry: There's no check to make sure the `x` & `y` combination translate into a valid `r` (assuming that's what you want. I think that's what you want?)

src/WCSimIBDGen.cc
```diff
@@ -0,0 +1,258 @@
...
+        G4double xs_weight = CrossSection(rand_ene, rand_cos_theta);
+
+        // Flux at rand_ene
+        G4double flux_weight = InterpolateSpectrum(energy, flux, rand_ene);
+
+        passed = (xs_test < xs_weight * flux_weight);
+    }
+}
+
+double WCSimIBDGen::CrossSection(double e_nu, double cos_theta) {
```
tdealtry, line 200: Can you add some comments to say which paper(s) this cross section came from?

src/WCSimIBDGen.cc
```diff
@@ -0,0 +1,258 @@
...
+
+    cross_section *= sigma_0 / 2;
+
+    // Convert from MeV^{-2} to mm^2 (native units for GEANT4)
+    cross_section *= CLHEP::hbarc * CLHEP::hbarc;
+
+    return cross_section;
+}
+
+double WCSimIBDGen::PositronEnergy(double e_nu, double cos_theta) {
```
tdealtry, line 265: Can you add some comments to say which paper(s) this correction came from?

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -955,6 +957,51 @@ void WCSimPrimaryGeneratorAction::GeneratePrimaries(G4Event* anEvent)
...
+
+        // Generate neutron
+        particleGun->SetParticlePosition(vtx);
+        particleGun->SetParticleDefinition(particleTable->FindParticle("neutron"));
+        particleGun->SetParticleEnergy(neutron.getT());
+        particleGun->SetParticleMomentumDirection(neutron.getV());
+        particleGun->SetParticleTime(0.);
+
+        // Generate positron
+        particleGun->GeneratePrimaryVertex(anEvent);
```
```suggestion
        particleGun->GeneratePrimaryVertex(anEvent);
        
        // Generate positron
```

tdealtry (changes requested):
Yeah that was my fault (the automated PR to update the reference files on a new tag failed) - fixed now, and your stuff is coming up green as expected

Thanks for this @JackFannon A few requests (mainly citing your work), but looks great

src/WCSimPrimaryGeneratorMessenger.cc
JackFannon: Changed the name of this command from `ibdspectrumCmd` to `ibdDatabaseCmd` to better reflect the function of the command. Have also added `SetNewValue` line for this.

src/WCSimPrimaryGeneratorAction.cc
JackFannon: Thanks for noticing this! Sorted to follow the same logic that `LIGen` uses.

src/WCSimIBDGen.cc
JackFannon: Added a comment about this referring to the inner detector.

src/WCSimIBDGen.cc
JackFannon: Added `while` loop to ensure that positions are generated inside the ID

src/WCSimIBDGen.cc
JackFannon: Added a check for this

src/WCSimIBDGen.cc
JackFannon: Comment added 👍

src/WCSimIBDGen.cc
JackFannon: Comment added 👍

data/DSNBFluxes.json
JackFannon: I have references for all the data, though I only have a proper source for the [Nakazato model](http://asphwww.ph.noda.tus.ac.jp/srn/). The other sets of data came from previous DSNB analyses done for SK-IV and SK-VI. I'm not sure where they generated the spectra, the only references I can find point to the papers describing the original models.

I've emailed Harada-san who wrote the last SK-VI DSNB analysis to see if they have a source for the files.

Maybe it would just be better to remove all but the Nakazato models in the case that I can't find a source for the data. What do you think?

CMakeLists.txt
```diff
@@ -106,9 +106,10 @@ set( WCSIM_CHECK_GEOMETRY_OVERLAPS $<IF:$<BOOL:${WCSim_Geometry_Overlaps_CHECK}>
 
 pbuilder_install_files(${CMAKE_INSTALL_PREFIX}/data
         ${PROJECT_SOURCE_DIR}/data/MuonFlux-HyperK-ThetaPhi.dat
-	${PROJECT_SOURCE_DIR}/data/biasprofile.dat
-	${PROJECT_SOURCE_DIR}/data/LightInjectors.json
-	${PROJECT_SOURCE_DIR}/data/LI_OD_laser_ball_photon_positions.csv
+	      ${PROJECT_SOURCE_DIR}/data/biasprofile.dat
+        ${PROJECT_SOURCE_DIR}/data/DSNBFluxes.json
+      	${PROJECT_SOURCE_DIR}/data/LightInjectors.json
+      	${PROJECT_SOURCE_DIR}/data/LI_OD_laser_ball_photon_positions.csv
```
JackFannon: Make whitespace consistent
```suggestion
	${PROJECT_SOURCE_DIR}/data/biasprofile.dat
        ${PROJECT_SOURCE_DIR}/data/DSNBFluxes.json
      	${PROJECT_SOURCE_DIR}/data/LightInjectors.json
      	${PROJECT_SOURCE_DIR}/data/LI_OD_laser_ball_photon_positions.csv
```

data/DSNBFluxes.json
tdealtry: This was really a bookkeeping request. So I think there's no need to delete models. Just leave a json field open for them to be filled later in a small PR, when you get the info

JackFannon: Great. By the sounds of it, SK requested the data from the various paper authors. I'll do the same and I'll put in a PR when I hear back.

Thanks for reviewing this!

tdealtry (approved):

tdealtry:
Thanks for this @JackFannon

JackFannon:
Thanks for reviewing this @tdealtry !

PR #384, opened 2023-11-02
Feature/mPMT LED generator

kmtsui:
New primary generator for mPMT LED
- Dynamically load PMT positions after detector construction, then calculate source position and direction based on mPMT id and LED id
- Allow short change in source direction for uncertainty estimation
- Source position and direction are saved as vertex and beam direction (first particle track direction) in output

Plus a small bug fix of saving multiple vertices information in radon generator


WCSim.mac
```diff
@@ -350,6 +350,22 @@
 #/gps/direction 1 0 0
 #/gps/position 0 0 0
 
+# Or inject photon from mPMT LED
+#/mygen/generator mPMT-LED
+# Use PMTid (from 1 to totalNum_mPMTs) to specify source mPMT
+# Use LEDid (from 0 to 11) to specify LED position, theta and phi directional change from nominal direction
```
tdealtry, line 358: Naive question here. I thought the LED mPMT proposal was to replace 5 (of the 19) 3" PMTs each with 8 LEDs = 40 LEDs in total. Why the number 12 here?

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -1372,7 +1366,173 @@ void WCSimPrimaryGeneratorAction::GeneratePrimaries(G4Event* anEvent)
...
+      //G4cout<<"Dir = "<<LEDdir.x()<<" "<<LEDdir.y()<<" "<<LEDdir.z()<<G4endl;
+
+      // G4Transform3D tubeTransform = myDetector->GetTubeTransform(mPMTLEDId);
+      // // Get tube orientation vector
+      // //G4Vector3D nullOrient = G4Vector3D(0,0,1);
+      // G4Vector3D pmtOrientation = tubeTransform * nullOrient;
+      // G4double distToPMT = 5*cm;
+      // G4double xpos = tubeTransform.getTranslation().getX() + pmtOrientation.x()*distToPMT;
+      // G4double ypos = tubeTransform.getTranslation().getY() + pmtOrientation.y()*distToPMT;
+      // G4double zpos = tubeTransform.getTranslation().getZ() + pmtOrientation.z()*distToPMT;
```
tdealtry: I think this can be deleted? e.g. `zpos` is already defined above

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -1372,7 +1366,173 @@ void WCSimPrimaryGeneratorAction::GeneratePrimaries(G4Event* anEvent)
...
+
+      //MyGPS->ClearAll();
+      //MyGPS->AddaSource(1.);
+      G4ParticleDefinition* pd = particleTable->FindParticle("opticalphoton");
+      MyGPS->SetParticleDefinition(pd);
+      MyGPS->GetCurrentSource()->GetEneDist()->SetEnergyDisType("Mono");
```
tdealtry: Can you do this kind of stuff only for the first event?

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -1372,7 +1366,173 @@ void WCSimPrimaryGeneratorAction::GeneratePrimaries(G4Event* anEvent)
...
+      G4ThreeVector dirx = dirz.orthogonal();
+      G4ThreeVector diry = dirz.cross(dirx);
+      MyGPS->GetCurrentSource()->GetAngDist()->DefineAngRefAxes("angref1",dirx);
+      MyGPS->GetCurrentSource()->GetAngDist()->DefineAngRefAxes("angref2",diry);
+      // MyGPS->GetCurrentSource()->GetAngDist()->SetAngDistType("iso"); // isotropic emission for now
+      // MyGPS->GetCurrentSource()->GetAngDist()->SetMinTheta(0.);
+      // MyGPS->GetCurrentSource()->GetAngDist()->SetMaxTheta(15*deg);
+      // MyGPS->GetCurrentSource()->GetAngDist()->SetMinPhi(0.);
+      // MyGPS->GetCurrentSource()->GetAngDist()->SetMaxPhi(360*deg);
+      // MyGPS->GetCurrentSource()->SetNumberOfParticles(1000);
```
tdealtry: Think you're doing this via mac file now, so please remove

tdealtry (changes requested):
Thanks @kmtsui Looks good. Few minor things, while we wait on me to fix the CI

tdealtry:
Turns out the CI is actually working 🥳 So just those couple of minor comments to address before merging

WCSim.mac
kmtsui: This is another set of calibration LEDs which will be placed between the small PMTs

kmtsui: The LEDs are the green dots in the figure.
image

src/WCSimPrimaryGeneratorAction.cc
kmtsui: Done

src/WCSimPrimaryGeneratorAction.cc
kmtsui: Done

WCSim.mac
tdealtry: Ahh I see, thanks. The difference between "mPMT LED" and "LED-mPMT". Not confusing at all... 
They're the names we use in the collaboration, so we have to use them. But what we can do is be explicit to avoid confusion. Something like
```
# Or inject photon from mPMT LED
# (Note that these are LEDs positioned in the standard mPMT modules.
#  They are not LEDs in the LED-mPMT modules)
```

tdealtry:
Thanks @kmtsui I've just set the CI off again, so happy for you to merge if it comes up green, and when you resolve the last comment. Thanks

tdealtry:
Ahh I always forget this. Rerunning a CI job uses the old, rather than new, CI workflow. You'll have to push the fix to the comment to set it running.

WCSim.mac
kmtsui: done

tdealtry (approved):

tdealtry:
Thanks @kmtsui

PR #385, opened 2023-11-02
Feature/PMT placement from table input

kmtsui:
Revise the non-replica PMT placement method using table input
The table includes

1. PMT type: type 1 or 2 in hybrid configuration
2. mPMTid: this changes the default ordering scheme which is based on section and copyNo
3. Section: 0 = barrel, 1 = top cap, 2 = top border ring, 3 = bottom cap, 4 = bottom border ring
4. X,Y,Z positions in cm
5. PMT Orientation: this is just for reference
6. Rotation: Z-rotation of the whole mPMT module in deg, so that the 19 small PMTs are in correct orientation
7. Use: disable the placement if equal to 0

With the `Section` variable, PMTs are placed in the correct logical/physical volumes. For barrel PMTs, the XY positions are used to calculate phi coordinate for placement, identify the correct phi-column and apply correction such that the mPMT is always touching the barrel wall.

An example table input for WCTE is available at `data/mPMT_Position_WCTE.txt`


data/mPMT_Position_WCTE.txt
```diff
@@ -0,0 +1,114 @@
+PMT position table for custom PMT placement. The direction is just for referenec but not used.
```
tdealtry:
```suggestion
PMT position table for custom PMT placement. The direction is just for reference but not used.
```

tdealtry (commented):
Thanks for this. All looks good. I just need to fix the CI, then can merge

tdealtry:
Turns out the CI is actually working 🥳 So please just accept this typo, then go ahead & merge

kmtsui:
Somehow the CI is broken again?
`/opt/Validation//Generate/daq_readfilemain: error while loading shared libraries: libG4Tree.so: cannot open shared object file: No such file or directory`

tdealtry:
CI is working again now. But since all you've done is fix a typo since it was working, I'll bypass running the CI on this again & just go ahead & merge. Thanks @kmtsui

PR #389, opened 2023-11-09
Update to IBD Cross Section and Selection Algorithm

JackFannon:
# Updates
## Cross section
The cross section calculation for the IBD generator previously used a simplified version of the Strummia-Vissani calculation. Now it uses the fully expanded version. This has been compared to the SNTools cross section calculation and has a maximum deviation of 1.2%. See the slides presented to the [HK UK PSC meeting on 31/10/2023](https://hkdbweb.in2p3.fr/meetingplanner.htm?meetingid=A00031301&segname=4876196) for more information. This plot shows a comparison between the SNTools cross section and the one implemented for this generator:
image

## Selection algorithm
The IBD generator uses rejection sampling to weight events by the product of the cross section and the flux. In the previous PR, the search area was defined by the product of the max cross section and the max flux. This overestimated the height of the distribution quite a bit. This has been updated to calculate the max of the cross section * flux over the energy range, giving a much smaller sampling area. This plot shows the uniformly distributed sample points (green for accepted points and pink for rejected points) and the true values for the product of the cross section and the flux (blue):
image
Before this fix the uniform distribution was ~4 times taller and the sampling algorithm was ~0.3 % efficient. This has been improved to ~18 % efficiency. There's further scope to improve this efficiency, though it may not be worth spending time on. To sample 100000 events the algorithm takes around 12 seconds and WCSim takes much more time to actually simulate the events.

## Usage
Class usage has changed slightly. The spectrum is now read in the initialiser. So
```c++
myIBDGenerator = new WCSimIBDGen(ibd_database, ibd_model, myDetector);
```
now creates a `WCSimIBDGen` object and reads the vector of flux and energy for the `ibd_model` specified from the `ibd_database`, all in one go. This also calculates the maximum value of cross section * flux. Only two other calls are now required in `PrimaryGeneratorAction` to generate an event. These are `GenRandomPosition` and `GenEvent` to generate the vertices, directions and momenta.

# Bug fixes
## Position generation
Previously generating positions between +- ID height instead of +- (ID height)/2
## Cross section calculation
There was a bug in the cross section calculation where sometimes the cross section would return as negative. This has been fixed in the complete rehaul of the calculation.
## Database bugs
Some small bugs in the model database. A couple of missing commas and fluxes in the energy vectors.


src/WCSimIBDGen.cc
```diff
@@ -104,6 +111,37 @@ double WCSimIBDGen::InterpolateSpectrum(std::vector<float> ener_vec, std::vector
...
+
+    for (float ene : energy) {
+        // Calculate the cross section for forward scattered positrons and backward scattered positrons. One of
+        // these values will always represent the maximum cross section across the energy range (forward at high
+        // energies and backward at low energies)
+        double x_sec_forward = dSigmaBydCosTheta(ene, 1.0);
+        double x_sec_backward = dSigmaBydCosTheta(ene, -1.0);
+
+        // Get the flux at ene
+        double flux_at_ene = InterpolateSpectrum(energy, flux, ene);
```
tdealtry: Is there a good reason why `InterpolateSpectrum()` takes arguments for `ener_vec` & `flux_vec`?
The only places `InterpolateSpectrum()` is called, member variables `energy` & `flux` are passed.
So why not just use the `energy` and `flux` data members in that method, and remove them as arguments?
If there is a good reason for the two extra arguments, please change it to passing the vectors by (const) reference

tdealtry (changes requested):
Thanks for this @JackFannon. One code request inline. One plot request - can you show something to confirm that the cross section is consistent with the cross section from another place (compare with the old version, sntools, the paper - whatever is easiest)

FYI the CI will actually work next time you push

src/WCSimIBDGen.cc
JackFannon: Thanks for the comments!

I've added the plot of the comparison between this cross section calculation and the SNTools one to the PR description.

There wasn't really a good reason for passing vectors to the interpolation function. I think the original thinking was that the function might also be used to interpolate other vectors but this didn't end up being the case. I've updated this to use the member variables.

tdealtry (approved):

tdealtry:
Thanks @JackFannon

JackFannon:
Thanks for reviewing this @tdealtry

PR #390, opened 2023-11-10
Updates to saving true particle tracks 

nickwp:
See [this PR](https://github.com/WCTE/WCSim/pull/14) from the WCTE fork for full discussion and details.

This changes how true particle tracks are chosen to be saved.

Previously, this logic was spread in two places: the EndOfEventAction and the TrackingAction. The TrackingAction marks events to be saved, while EndOfEventAction chooses some of these to save (while flat tree format saves all marked to be saved).

Now, all tracks marked to be saved by TrackingAction are saved. Optionally, it also saves tracks if they are in the chain of particles from primary particles until (but not including) photons that produce Cherenkov hits. (Cherenkov hit producing photons have their details already saved in the WCSimRootCherenkovHitTime class - later this could be removed and just save the track, but for now avoiding breaking any analysis code relying on that).

The option to save all tracks involved in producing hits can be turned on or off with /Tracking/saveHitProducingTracks

The logic for marking tracks to be saved is also simplified:
Anything in the ParticleList
Anything in the ProcessList
These lists are left the same as they were before, but can be added to by mac file (no change).

Here are benchmarks for 100 x 500 MeV electrons with the new version (with and without the saveHitProducingTracks) and old version:
**saveHitProducingTracks true**
User time (seconds): 153.71
Maximum resident set size (kbytes): 155,280
File size (kbytes): 39,072
**saveHitProducingTracks false**
User time (seconds): 156.01
Maximum resident set size (kbytes): 155,228
File size (kbytes): 38,918
**old version**
User time (seconds): 151.38
Maximum resident set size (kbytes): 204,488
File size (kbytes): 38,197

So only about 2% increase in file size, negligible change to memory usage and processing time.


kmtsui:
The first `TEST_SK_5MeVem_centre_xdir` test completely failed with seg fault. Anything special in this geometry?

nickwp:
hmm, in my tests I didn't get any segfaults and geometry should be unchanged. If I can get the mac file used for the check, I'll see if I can figure out what's wrong

tdealtry:
Thanks @nickwp The simple one is [here](https://github.com/tdealtry/Validation/blob/master/Generate/macReference/wcsim_test_seed20230628.mac)

Looking at [the validation plots](https://wcsim.github.io/Validation/d5aea2c061172703d470ea749cdac041dda65a99/index.html), everything is _failing_ but this is to be expected - there are different tracks saved

Also there are more mac files failing, also via seg faults
- `wcsim_sk_*`
- `wcsim_hybrid_*`

So basically it seems like SK & FD ID only geometries are seg faulting, while FD ID+OD, IWCD w/ OD, IWCD w/o OD are fine. Intriguing

nickwp:
This seems to be the worst kind of bug... Initially I couldn't reproduce the segfault at all. But then running all the different test mac files I did see it occur once, but for a different one to the CI tests. I tried changing random seed and it still seemed to segfault. but then one time it didn't.
In the end, I tried running a bunch of times with identical configuration (including identical random seed) and it turned out that most of the time it succeeded but occassionally segfaults. When it segfaults, it's always when processing the very first event.
So I guess it's an edge-case bug that occurs only sometimes, depending on some random memory access. When it works, it's probably by accident that it accesses some memory that it shouldn't, but just happens to be coincidentally ok to access sometimes.
I'll continue trying to figure out the underlying cause.

nickwp:
Found and (hopefully) fixed the bug. It was an edge case where for the fvery first trajectory created in a run, the pointer to its parent trajectory would be initialised randomly instead of set to 0. Usually this didn't matter since it would get initialised to 0 anyway, but occasionally got randomly some other value, which would cause the segfault in the rare situation where it actually attempts to access it early on. Hopefully the CI confirms it works now.

tdealtry:
Nice one @nickwp Thanks for this
I'll issue another release after this merge, since the default output has changed (I reiterate it is just the tracks that have been saved - all hit output is identical)

PR #392, opened 2023-11-20
Fix bug for photon tracking

nickwp:
One-line fix for bug that apparently crept in when renaming the "ParentID" variables to "ParentSavedTrackID" for true cherenkov hits. In the version recenently merged in PR #390 the track ID being saved for a hit was mistakenly changed to the ID of the photon track, but it should be the ID of the parent of the photon track, since that is the track which is actually saved in the output.

I've tested by generating some electron events and checking that now, for all true non-dark-noise hits, the track with the ID given in WCSimRootCherenkovHitTime class's "ParentSavedTrackID" is actually a track that is saved in the output (dark noise has parent -1, as before).


nickwp:
The failing CI appears to be the same track information that changed in #390. Do some reference plots need updating?

tdealtry:
Thanks @nickwp

PR #393, opened 2023-11-21
Make jobOptions.mac usable for WCSimOpticalPhysics

kmtsui:
Define `WCSimOpticalPhysics` in `WCSimPhysicsListFactory` constructor, then we can use macro commands in `jobOptions.mac` to control `WCSimOpticalPhysics` options


tdealtry:
Thanks @kmtsui

PR #394, opened 2023-11-28
Update iwcd geom

kmtsui:
Change in IWCD ID default diameter, and minor bug fix in GPS generator verbose


tdealtry:
Thanks for this. CI is behaving as expected - namely that the IWCD+OD checks are failing, but everything else is passing

Can you update the IWCD ID only geometry too? Or, if that geometry is not really intended to be used anymore, add a call to a new `IWCDGeometryNoLongerUsed()` function (similar to `HKGeometryNoLongerUsed()`) that'll force people to think hard about whether they should be using the old geo (but is still there for backwards compatability)

When you do this, I can merge & create a new release (that'll create new IWCD reference plots...)

kmtsui:
Thanks for the comment, I indeed forgot the ID-only geometry. Still need to keep the option alive for a while

tdealtry:
Great - IWCD ID only now coming up red as expected

Thanks @kmtsui

PR #395, opened 2023-11-29
Install missing *.hpp files

gondiaz:
This PR adds the *.hpp files into the install directory.

As today it is relevant for the json.hpp file. 

Since this file was not installed, WCSim dependent code needing needing this file (such as WCSimFQTuner) is using the system's (for Linux installs) json.hpp which may differ.

This fixs allows the WCSimFQTuner to be installed in MacOS systems without extra changes in the CMakeLists.


tdealtry:
Nice one. Thanks @gondiaz

PR #396, opened 2023-11-30
Fix PMT overlap in IWCD geom

kmtsui:
Properly set `WCCapEdgeLimit` such that cap PMT will not overlap with edge volume. 
Also change `WCPMTODPercentCoverage` to keep the same number of OD PMT.


tdealtry:
Thanks for this @kmtsui It's certainly fixed the "Optical photon is killed because of missing refractive index" issues
Before I merge, just want to confirm that you're happy with the number of PMT change
```
< Type 1 ID PMT number & size       7980    4.00
---
> Type 1 ID PMT number & size       7676    4.00
4c4
< OD PMT number & size               364   40.00
---
> OD PMT number & size               404   40.00
```

kmtsui:
Yes this gives 404 mPMTs and OD PMTs which is the correct number.

PR #397, opened 2023-12-06
Feature/photon history

kmtsui:
Toggle WCSim to save photon scattering and reflection history when `-DWCSIM_SAVE_PHOTON_HISTORY_FLAG=ON` is set in `cmake` configuration. This feature is mostly for calibration studies when we want to know the photon propagation histories with respect to different water qualities and surface reflectivity.

Major changes include:
1. Save `trackID` per photon in `WCSimWCHit` and `WCSimWCDigi`
2. Save the number of scattering and reflection experienced by a photon in `WCSimTrajectory`. In `WCSimTrajectory::AppendStep()`, check whether a Rayleigh/Mie scattering happens, or whether a boundary reflection occurs and save the boundary surface type (`kOtherS,kBlackSheetS,kReflectorS, kPhotocathodeS` defined in `WCSimEnumerations`)
3. Construct new `WCSimRootCherenkovHitHistory` class to save the photon scattering and reflection histories. Users can access this class in a way similar to `WCSimRootCherenkovHitTime`, thus get the scattering and reflection histories per Cherenkov photon. An example usage is also documented in `sample-root-scripts/sample_readfile.C`.
4. In `WCSimEventAction::EndOfEventAction()`, retrieve photon histories from `WCSimTrajectory` container and save in output root file.
5. The photon process check and photon histories saving only occur when `-DWCSIM_SAVE_PHOTON_HISTORY_FLAG=ON` is set in `cmake` configuration. This will enforce `WCSimTrajectory` to be stored in `WCSimTrackingAction::fpTrackingManager` for every photon, overriding setup from `/Tracking/fractionOpticalPhotonsToDraw`.

Bug fix:
1. In `WCSimWCTriggerBase::AlgNDigits`, if the digi hit time range is smaller than trigger window, the find trigger loop is invalid. Pushed a fix to enforce `window_start_time<=window_end_time` in all case

### DIFF
--- a/include/WCSimRootEvent.hh
+++ b/include/WCSimRootEvent.hh
@@ -170,17 +170,19 @@ class WCSimRootCherenkovHitHistory : public TObject {
 
 private:
   
-  Int_t   fScat;
-  std::vector<Int_t> fReflec;
+  Int_t   fNRayScat;
+  Int_t   fNMieScat;
+  std::vector<ReflectionSurface_t> fReflec;
 
 public:
   WCSimRootCherenkovHitHistory() {}
-  WCSimRootCherenkovHitHistory(Int_t scat, std::vector<Int_t> refle);
+  WCSimRootCherenkovHitHistory(Int_t nRayScat, Int_t nMieScat, std::vector<ReflectionSurface_t> refle);
   virtual ~WCSimRootCherenkovHitHistory() { }
   bool CompareAllVariables(const WCSimRootCherenkovHitHistory * c) const;
 
-  Int_t     GetScatter() const { return fScat; }
-  std::vector<Int_t> GetReflection() const { return fReflec; }
+  Int_t     GetNRayScatters() const { return fNRayScat; } // Get the number of Rayleigh scattering a photon experienced
+  Int_t     GetNMieScatters() const { return fNMieScat; } // Get the number of Mie scattering a photon experienced
+  std::vector<ReflectionSurface_t> GetReflectionSurfaces() const { return fReflec; } //  Get the vector of reflection surfaces a photon experienced
 
   ClassDef(WCSimRootCherenkovHitHistory,1)
 };
@@ -504,8 +506,9 @@ public:
 					   std::vector<TVector3>  photonEndPos,
 					   std::vector<TVector3>  photonStartDir,
 					   std::vector<TVector3>  photonEndDir);
-  WCSimRootCherenkovHitHistory   *AddCherenkovHitHistory(Int_t scat,
-					   std::vector<Int_t> reflec);
+  WCSimRootCherenkovHitHistory   *AddCherenkovHitHistory(Int_t nRayScat,
+             Int_t nMieScat,
+					   std::vector<ReflectionSurface_t> reflec);
   TClonesArray        *GetCherenkovHits() const {return fCherenkovHits;}
   TClonesArray        *GetCherenkovHitTimes() const {return fCherenkovHitTimes;}
   TClonesArray        *GetCherenkovHitHistories() const {return fCherenkovHitHistories;}
--- a/include/WCSimRootEvent.hh
+++ b/include/WCSimRootEvent.hh
@@ -165,6 +165,26 @@ public:
   ClassDef(WCSimRootCherenkovHitTime,2)
 };
 
+// Scattering and reflection history for each Cherenkov hit (photon)
+class WCSimRootCherenkovHitHistory : public TObject {
+
+private:
+  
+  Int_t   fScat;
+  std::vector<Int_t> fReflec;
+
+public:
+  WCSimRootCherenkovHitHistory() {}
+  WCSimRootCherenkovHitHistory(Int_t scat, std::vector<Int_t> refle);
+  virtual ~WCSimRootCherenkovHitHistory() { }
+  bool CompareAllVariables(const WCSimRootCherenkovHitHistory * c) const;
+
+  Int_t     GetScatter() const { return fScat; }
+  std::vector<Int_t> GetReflection() const { return fReflec; }
+
+  ClassDef(WCSimRootCherenkovHitHistory,1)
+};
+
 
 //////////////////////////////////////////////////////////////////////////
 
@@ -362,6 +382,8 @@ private:
   Int_t                fCherenkovHitCounter;
   Int_t                fNcherenkovhittimes;      // Number of hits in the array
   TClonesArray         *fCherenkovHitTimes;      //-> Array of WCSimRootCherenkovHits
+  Int_t                fNcherenkovhithistories;      // Number of hits in the array
+  TClonesArray         *fCherenkovHitHistories;  //-> Array of WCSimRootCherenkovHitHistories
 
   Int_t                fNumDigitizedTubes;  // Number of digitized tubes
   Int_t                fNcherenkovdigihits;  // Number of digihits in the array
@@ -440,6 +462,7 @@ public:
   Int_t               GetNtrack_slots()       const {return fTracks->GetLast() + 1; } //don't use fNtrack_slots directly as it doesn't take into account automatic TClonesArray shortening when tracks at start/end are removed
   Int_t               GetNcherenkovhits()     const {return fNcherenkovhits; }
   Int_t               GetNcherenkovhittimes() const {return fNcherenkovhittimes;}
+  Int_t               GetNcherenkovhithistories() const {return fNcherenkovhithistories;}
   Int_t               GetNcherenkovdigihits() const {return fNcherenkovdigihits;}
   Int_t               GetNcherenkovdigihits_slots() const {return fCherenkovDigiHits->GetLast() + 1; } //don't use fNcherenkovdigihits_slots directly as it doesn't take into account automatic TClonesArray shortening when digits at start/end are removed
   Float_t             GetSumQ()               const { return fSumQ;}
@@ -481,8 +504,11 @@ public:
 					   std::vector<TVector3>  photonEndPos,
 					   std::vector<TVector3>  photonStartDir,
 					   std::vector<TVector3>  photonEndDir);
+  WCSimRootCherenkovHitHistory   *AddCherenkovHitHistory(Int_t scat,
+					   std::vector<Int_t> reflec);
   TClonesArray        *GetCherenkovHits() const {return fCherenkovHits;}
   TClonesArray        *GetCherenkovHitTimes() const {return fCherenkovHitTimes;}
+  TClonesArray        *GetCherenkovHitHistories() const {return fCherenkovHitHistories;}
 
   WCSimRootCherenkovDigiHit   *AddCherenkovDigiHit(Double_t q,
 						   Double_t t,
@@ -497,7 +523,7 @@ public:
 
   TClonesArray	      *GetCaptures() const {return fCaptures;}
 
-  ClassDef(WCSimRootTrigger,5) //WCSimRootEvent structure
+  ClassDef(WCSimRootTrigger,6) //WCSimRootEvent structure
 };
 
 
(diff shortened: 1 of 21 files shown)


sample-root-scripts/sample_readfile.C
```diff
@@ -253,6 +259,17 @@ int sample_readfile(const char *filename="../wcsim.root", bool verbose=false)
...
 		    <<", parent TrackID: "<<thehittimeobject->GetParentID()<<";";
 	      }
+#ifdef WCSIM_SAVE_PHOTON_HISTORY
+        // use the same index as WCSimRootCherenkovHitTime
+        WCSimRootCherenkovHitHistory *thehithistoryobject =  dynamic_cast<WCSimRootCherenkovHitHistory*>(historyArray->At(thephotonsid));
+	      if(thehithistoryobject)
+        {
+          // Number of scattering, and types of reflection surface (WCSimEnumerations)
+          cout<<" Scattering: "<<thehithistoryobject->GetScatter()<<", Reflection: ";
+          for (auto r: thehithistoryobject->GetReflection()) cout<<r<<" ";
```
tdealtry: Maybe nicer to use `EnumAsString()` here?
More verbose, but clearer

include/WCSimRootEvent.hh
```diff
@@ -165,6 +165,26 @@ public:
...
+  Int_t   fScat;
+  std::vector<Int_t> fReflec;
+
+public:
+  WCSimRootCherenkovHitHistory() {}
+  WCSimRootCherenkovHitHistory(Int_t scat, std::vector<Int_t> refle);
+  virtual ~WCSimRootCherenkovHitHistory() { }
+  bool CompareAllVariables(const WCSimRootCherenkovHitHistory * c) const;
+
+  Int_t     GetScatter() const { return fScat; }
```
tdealtry: Does this return the parent Cherenkov photon? The number of scatters this Cherenkov photon has experienced? Something else?
Need a comment here to clarify, and potentially a more informative method name

include/WCSimRootEvent.hh
```diff
@@ -165,6 +165,26 @@ public:
   ClassDef(WCSimRootCherenkovHitTime,2)
 };
 
+// Scattering and reflection history for each Cherenkov hit (photon)
+class WCSimRootCherenkovHitHistory : public TObject {
+
+private:
+  
+  Int_t   fScat;
+  std::vector<Int_t> fReflec;
```
tdealtry: Surely this is what we want?
```suggestion
  std::vector<ReflectionSurface_t> fReflec;
```
and similarly elsewhere in this class

include/WCSimRootEvent.hh
```diff
@@ -165,6 +165,26 @@ public:
...
+
+public:
+  WCSimRootCherenkovHitHistory() {}
+  WCSimRootCherenkovHitHistory(Int_t scat, std::vector<Int_t> refle);
+  virtual ~WCSimRootCherenkovHitHistory() { }
+  bool CompareAllVariables(const WCSimRootCherenkovHitHistory * c) const;
+
+  Int_t     GetScatter() const { return fScat; }
+  std::vector<Int_t> GetReflection() const { return fReflec; }
```
tdealtry: Could call it `GetReflectionSurfaces()`
(at the very least, it should be plural)

include/WCSimTrajectory.hh
```diff
@@ -112,6 +113,12 @@ public: // with description
     return bTypes;
   }
 
+// Functions to set/get photon history
+  inline void AddPhotonScatter(G4int val) { pScatter += val; }
+  inline void AddPhotonReflection(G4int val) { pReflec.push_back(val); }
```
tdealtry: Why not use `ReflectionSurface_t` for these functions? Similarly for the member variables

src/WCSimTrajectory.cc
```diff
@@ -226,10 +261,40 @@ void WCSimTrajectory::AppendStep(const G4Step* aStep)
...
+  // Add photon history
+  if (PDGEncoding==0)
+  {
+    const G4VProcess* pds = thePostPoint->GetProcessDefinedStep();
+    //G4cout<<"Having optical photon in AppendStep "<<pds->GetProcessName()<<G4endl;
+    if ( pds->GetProcessType() == fOptical )
+    {
+      if ( pds->GetProcessSubType() == fOpRayleigh || pds->GetProcessSubType() == fOpMieHG )
+      {
+        AddPhotonScatter(1);
```
tdealtry: Naive question, but would it ever be useful to know how many of each scattering process (rather than the combined total)?
Now would be the perfect time to split the numbers up

include/WCSimWCHit.hh
```diff
@@ -31,7 +31,7 @@ class WCSimWCHit : public G4VHit
  public:
   
   void SetTubeID       (G4int tube)                 { tubeID = tube; };
-  void SetTrackID      (G4int track)                { trackID = track; };
+  void SetTrackID      (G4int track)                { trackID.push_back(track); };
```
tdealtry: Please rename to something like `AddTrackID()` to make it clear that it's pushing back, rather than setting the only value

tdealtry (changes requested):
Thanks @kmtsui. A few inline comments, mostly about naming of things. Also
- Please add a comment to WCSim.mac to say that `/Tracking/fractionOpticalPhotonsToDraw` is ignored when `-DWCSIM_SAVE_PHOTON_HISTORY_FLAG=ON` is set (and 100% of photons are saved).
- The CI runs successfully, which is good news! I would like to see something comparing results with `-DWCSIM_SAVE_PHOTON_HISTORY_FLAG` set to ON and OFF though, to ensure nothing odd is happening

tdealtry:
Good spot on the random number fix. This shows why we should put in more than 1 random number generator in GHOST...

Can you pull in from develop? It looks like your IWCD geom fix isn't accounted for in the CI.

kmtsui:
It seems that photon tracks with WLS in OD PMTs are recorded in the output root files, I am trying to remove them...
And somehow when there are too many tracks, the [CI](https://github.com/WCSim/WCSim/actions/runs/7148770243/job/19470222190?pr=397#step:11:9363) is failing. Any thoughts?

tdealtry:
It's when it's running [the WCSim file -> simpler tree convertor](https://github.com/WCSim/Validation/blob/master/Generate/daq_readfilemain.C) I guess its running out of memory
Looking [here](https://wcsim.github.io/Validation/061d6447e4cc91fc896fed70a7e0a97ec5370f7f/2/2/index.html) we see that it gets through the 20" ok, and fails for mPMTs at ~event 170 of 200. Each call of the convertor does 20" then mPMT then OD, so there's some kind of memory leak going on - I've applied a fix (2 actually - 1] to cleanup some memory at the end of the function, 2] as a more foolproof thing by changing how the script is called - one PMT type per process)

kmtsui:
CI broken due to [typo ](https://github.com/WCSim/Validation/blob/fa5f3ea25f52c06a54f3e4b76cbaf0f75417aa0d/Generate/Generate.sh#L4) `pmtype` <->`pmttype`

tdealtry:
Doh! Fixed and rerunning on your latest commit

kmtsui:
Thanks. The CI checks are finished and looking good. In the end, I need to change the track saving flags such that they are not applied to optical photons. But the side effect is that photon tracks will not be saved at all when we want to use the `ProcessList` or `ParticleList` control. 
I am think whether it is good to roll back to [d3ee4f0](https://github.com/WCSim/WCSim/pull/397/commits/d3ee4f03e997afa53fc2d3afcd1745489940b28d) and accept the large number of additional tracks as known feature.

tdealtry:
All green is always good!

I've lost track a little on precisely what's going on...

- [The CI](https://wcsim.github.io/Validation/061d6447e4cc91fc896fed70a7e0a97ec5370f7f/index.html) on 061d644 (effectively d3ee4f0) is all green for everything but geometries with ODs, where the number of tracks is increasing. In IWCD, the increase a small (up to 20%) but for HK FD it's large (3000%+).
  - Need to grab the file size information from the CI really, to see if this is something to be worrying about. But it's potentially worrying.
- Since the CI is green for all other geometries, this suggests it's extra optical photons that are being saved related to something in the OD. Likely the WLS plates
  - Why is the WLS special? Is it something to do with extra wavelength shifting processes generating two levels of optical photon tracks? (Something like Cherenkov photon -> absorption in WLS (gets a != 0 track ID) -> remmission = new optical photon)
- It seems like `-DSAVE_PHOTON_HISTORY=ON` has been the default for quite a few CI runs. Why then has the tracking information not changed for all detectors? Since (among other things) this effectively sets `/Tracking/fractionOpticalPhotonsToDraw` to 100%

I suppose if we need some slightly different logic for `-DSAVE_PHOTON_HISTORY` = ON and OFF, in order to give the functionality & keep file size in check, we can do. Since the checks do show that things are identical, other than extra OD tracks, when things are ON. We can have OFF be the size light mode, with the fine control with `ProcessList` or `ParticleList` control options

kmtsui:
The large increase in the number of tracks being saved is due to:
1. [All photon trajectories are stored](https://github.com/kmtsui/WCSim/blob/feature/photon_history/src/WCSimTrackingAction.cc#L73) (equivalent to `/Tracking/fractionOpticalPhotonsToDraw 100`) 
2. [Optical photons from WLS are saved ](https://github.com/kmtsui/WCSim/blob/feature/photon_history/src/WCSimTrackingAction.cc#L135)because of the "OpWLS" being added to [ProcessList](https://github.com/kmtsui/WCSim/blob/feature/photon_history/src/WCSimTrackingAction.cc#L29)
3. [Parent photon tracks of the WLS photons are saved](https://github.com/kmtsui/WCSim/blob/feature/photon_history/src/WCSimTrackingAction.cc#L172C35-L172C57) becaue of the `saveHitProducingTracks` flag

Only the OD events have such WLS processes, so we see differences only in those OD geometries. We will see the same behavior if we use the `/Tracking/fractionOpticalPhotonsToDraw 100` command instead of `-DSAVE_PHOTON_HISTORY=ON`

If we want to keep the `/Tracking/fractionOpticalPhotonsToDraw` usage, it will need more an extra flag to mark whether the photon trajectories are coming from `/Tracking/fractionOpticalPhotonsToDraw` or `-DSAVE_PHOTON_HISTORY=ON`, or do another random number throw in the `PostUserTrackingAction()`

kmtsui:
I have added an extra flag to mark whether to save the photon track in ROOT. Now `/Tracking/fractionOpticalPhotonsToDraw` has the same behavior no matter `-DSAVE_PHOTON_HISTORY` is `ON` or `OFF`. Accordingly `ProcessList` and `ParticleList` controls should also work properly.

tdealtry:
Thanks for the explanation & the careful work with this @kmtsui 
I'm happy to merge after you set the default for `WCSIM_SAVE_PHOTON_HISTORY_FLAG` to `OFF`

tdealtry (approved):

tdealtry:
Thanks @kmtsui

gondiaz:
Some methods here were modified, for example in `WCSimWCHit`, `SetTrackID` which is now `SetTrackID` and `GetTrackID()` to `GetTrackID(int i)`.

This changes directly affect [WCSimFQTuner](https://github.com/fiTQun/WCSimFQTuner), so I would suggest that from now on, before merging anything into `WCSim` we should also port the changes to the tuner (which could also be part of WCSim though). I was not following the details of the new functionality, so could you check for it?

kmtsui:
@gondiaz The change in `WCSimWCHit` and `WCSimWCDigi` is to save all photon track ID (instead of the just first one) for a PMT with multiple photon hit. I don't have write access to [WCSimFQTuner](https://github.com/fiTQun/WCSimFQTuner) so cannot update it.
The fact that WCSimFQTuner lives under another (private) organization makes it difficult to maintain. We would need the admin there to help with the integration.

PR #398, opened 2023-12-16
Cosmic flux fix

arturof:
Cosmic muons were always crossing exactly through the centre of the detector. Added code so they cross a random point uniformly distributed in the detector volume.

Fixed bug where cosmic histograms content was shifted by 1 bin and another one where the GetBinContent function was incorrectly used to find the muon mean energy.

Run was crashing because cosmic muons were generated outside the world volume and when deleting the cosmic histograms in the destructor. Both issues are fixed now.


tdealtry:
Looks great. Thanks @arturof 
Since the CI doesn't check cosmics automatically, I wonder if you can post a plot (or few) to show this is working as expected. I presume you've made some already for a talk?

arturof:
Yes, I have a few plots.

WCSim converts the input cosmic flux text file (data/MuonFlux-HyperK-ThetaPhi.dat by default) in histograms that are saved in a file named cosmicflux.root. The histograms are the flux and mean energy as a function theta and phi angles. The last bin of the histograms were empty because the first bin in the text file is 0 instead of 1 as in ROOT convention. I fixed that and they look like this now:

hFluxCosmics
hEmeanCosmics

There was also a problem finding the cosmic muon energy from the mean energy histogram. The muon energy distribution didn't make sense but it looks like expected now:

wcsim_HK_cosmics_histos_genE

I'm assuming the input text file is correct. There are no muons below 100 GeV in the plot and that's compatible with the text file. I'm guessing that was made on purpose to study high energy cosmic muons. I don't know who created the input text file.

Lastly, all muons were generated passing exactly through the centre of the detector. As a consequence, there were no Cherenkov photons from muons starting in a gap along Z = 0. The photon start position looked like this after generating 300 cosmic muons:

wcsim_HK_cosmics_run_start_rz

The code I added generates muons passing through a random point in the ID volume. The same distribution doesn't have the gap anymore (1000 muons):

wcsim_HK_cosmics_merged_histos_start_rz

tdealtry:
All looks great. Thanks @arturof

PR #401, opened 2024-01-17
Allow multiple runs

nickwp:
Geant4 allows multiple runs with different generator settings by having multiple /run/beamOn commands and changing generator settings between them. This is useful if e.g. you want to simulate some events at one source position followed by a some events at another position, etc., without needing to execute WCSim once for each position. You can just set the mac file to do them sequentially.

But this doesn't work in WCSim because the EndOfRunAction recreates (overwrites) the output root file, so only the final run gets saved.
To fix this, the StartOfRunAction can check the run number and either (re)create the output file for the first run, or open it to append for subsequent runs. Also, when writing each event, the tree itself is filled rather than filling individual branches, because when filling branches it was then required to specify manually the number of events in the tree. It's better to just fill the tree so the number of events is automatically correct, rather than tracking it through multiple runs and specifying it manually.

Tested with a macro file ending with
```
/gps/direction -1 0 0
/run/beamOn 2
/gps/direction 1 0 0
/run/beamOn 3
```
Checked that the output had five events in total, the first two events simulated tracks with direction (-1, 0, 0) and the final three simulated tracks with directions (1, 0, 0).

### DIFF
--- a/src/WCSimRunAction.cc
+++ b/src/WCSimRunAction.cc
@@ -168,7 +168,7 @@ void WCSimRunAction::BeginOfRunAction(const G4Run* aRun)
       }
       else{
           TFile *hfile = new TFile(rootname.c_str(), "UPDATE");
-          WCSimTree = (*TTree) hfile->Get("wcsimT");
+          WCSimTree = (TTree*) hfile->Get("wcsimT");
           wcsimrooteventbranch = WCSimTree->GetBranch("wcsimrootevent");
           wcsimrooteventbranch2 = WCSimTree->GetBranch("wcsimrootevent2");
           wcsimrooteventbranch_OD = WCSimTree->GetBranch("wcsimrootevent_OD");
--- a/src/WCSimRunAction.cc
+++ b/src/WCSimRunAction.cc
@@ -47,7 +47,7 @@ WCSimRunAction::~WCSimRunAction()
 
 }
 
-void WCSimRunAction::BeginOfRunAction(const G4Run* /*aRun*/)
+void WCSimRunAction::BeginOfRunAction(const G4Run* aRun)
 {
 
   fSettingsOutputTree = NULL;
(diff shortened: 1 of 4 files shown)


tdealtry:
Hi @nickwp thanks for this. Looks good.
The only question I have is about documenting this feature, as it could produce unexpected results e.g.
- The options tree won't have a record of how many events were in each run
- Attempt to change something that needs to be set before any runs have taken place (e.g. a tuning parameter like scattering) and things won't actually change
  - Geometry changes are something that *may* fit into this category, so I don't think it's as simple as saying you can change anything in the `WCSim.mac` file, but not in the `jobOptions.mac` or `tuning_parameters.mac`

So some suggested wording (please add to the bottom of `WCSim.mac`)

```
# It is possible to perform multiple runs of WCSim in a single process
# The benefit of this is to reduce the time spent on WCSim initialisation
# in a use case e.g. simulating events at multiple source positions
# In order to do this, you can repeat the /run/beamOn command e.g.
# /gun/direction -1 0 0
# /run/beamOn 2
# /gun/direction 1 0 0
# /run/beamOn 3
# However, be warned that you cannot change all options between runs
# - All parameters in tuning_parameters.mac and jobOptions.mac changed will have no effect
# - Some parameters in WCSim.mac may not have an effect (e.g. geometry)
# - It is up to you to perform the bookkeeping of which events were run with which options
# It is therefore your responsibility to confirm that the options you are changing work as expected in multi-run mode
```

src/WCSimRunAction.cc
```diff
@@ -514,9 +527,10 @@ void WCSimRunAction::EndOfRunAction(const G4Run*)
     // Close the Root file at the end of the run
     TFile* hfile = WCSimTree->GetCurrentFile();
     hfile->cd();
-    optionsTree->Fill();
-    optionsTree->Write();
-    
+    if(aRun->GetRunID() == 0) {
```
tdealtry: Is it possible to make this be filled on every run? Things don't change in your example, but if someone e.g. changed the dark rate, it would be useful

Hopefully ROOT's compression algorithm is sensible and the overhead for having multiple identical entries in a tree is small

nickwp: Yes, can do this. Initially I decided against doing so because it would not be clear which options tree entries corresponded to which events. But on further investigation I found that the WCSimRootEvent has an EventHeader with a run number. It's currently hard coded to be always 0, but I'll change it to correspond to the actual run number and therefore correspond to the relevent options tree entry.

nickwp:
Thanks Tom

I've updated it to fill the options tree for each run, and to track which events are in which run using the already existing (but never properly implemented, until now) run number in the event header. This run number corresponds to the entry in the options tree for a given event, as well as simply specifying which /run/beamOn command produced the event.

I added the suggested documentation, with additional documentation about the options tree entries and run number in the event header.

src/WCSimRunAction.cc
tdealtry: That's even better. Thanks Nick!

tdealtry:
Thanks @nickwp I'll merge this on green

PR #402, opened 2024-01-18
Avoid using wrong mPMT id in replica placement

kmtsui:
When `/WCSim/PMT/ReplicaPlacement true` is used with `/WCSim/PMT/PositionFile`, the mPMT id is read from the wrong replica number. An extra protection is added to avoid such behavior.


tdealtry:
Thanks @kmtsui

PR #405, opened 2024-01-29
Update WCTE tank height and diameter and mPMT positions

nickwp:
These updates make changes to the WCTE tank size and mPMT positioning using the NoReplica mPMT placement method, to match the actual design of WCTE. The design is as extracted by Dean Karlen and described [here](https://wcte.hyperk.ca/wg/integration/conventions) and [here](https://github.com/WCTE/Geometry).

See [this presentation](https://wcte.hyperk.ca/wg/simulation-and-analysis/meetings/2024/20240129/meeting/wcte-geometry-in-wcsim) for the details of the new values for WCIDRadius, WCIDHeight, and the mPMT positions.

A test run of a hundred 500 MeV electrons did not indicate any geometry overlaps and had zero stuck track exceptions.

Visualisation also suggests it look ok:
WCTE_newgeom
WCTE_newgeom_top

As an extra check, I looked at the position of the central PMT of the end-cap mPMTs and barrel PMTs as output in the WCSimRootGeom:

For end-cap mPMTs, their central PMT's centre is at +/- 1332.44 cm in the WCSim coordinate system. The PMT's glass sphere is 50mm with 1mm thickness, so the distance between opposite end-cap PMT photocathodes is 2*(1332.44-(50-1)) = 2566.88mm. This matches the design values, where the mPMT backplates are at -1105.48 mm and +1955 mm in Dean's coordinate system, and the distance from mPMT backplate to central PMT photocathode is 246.8 mm, so the distance between opposite PMT photocathodes is 1105.48+1955-2*(246.8) = 2566.88 mm.

For barrel mPMTs, their central PMT's centre is at a radius of 1513.28 mm, so the central PMT's photocathode is at 1513.28-(50-1) = 1464.28 mm. This matches the design where the mPMT backplates are at a radius of 1711.083 mm so the central PMT photocathodes are at 1711.08-246.8 = 1464.28 mm.


kmtsui:
The mPMT module numbering is not updated in this PR, so it will stay the same as I last checked the WCTE solidworks model. We can either revert back to the original scheme, or update/confirm that it strictly follows the design scheme. @nickwp Which way do you prefer?

kmtsui:
@tdealtry BTW the WCSim validation macro does not include `/WCSim/PMT/PositionFile data/mPMT_Position_WCTE.txt` before `/WCSim/Construct` so the new PMT positions are not reflected in the validation.

nickwp:
> The mPMT module numbering is not updated in this PR, so it will stay the same as I last checked the WCTE solidworks model. We can either revert back to the original scheme, or update/confirm that it strictly follows the design scheme. @nickwp Which way do you prefer?

I think for mPMT module numbering, both Dean's ordering and the ordering in the SolidWorks file are both somewhat arbitrary and hasn't really been discussed or agreed yet. Thomas Lindner mentioned to me that he had some queries about it and wanted to discuss it. I asked Dean myself and he says he came up with his one to mimic the ordering of the PMTs within mPMTs.

So in this PR I tried to make minimal changes to WCSim, since I assumed that would be least likely to cause any problem for any existing analysis or code. Do you know if the existing numbering is what people will be using already in other code? Or did it change at some point recently?

I think I'm happy with just using whatever numbering causes the least trouble for people using WCSim for now, and then we can change it once we decide on a final numbering agreed with the other working groups.

tdealtry:
> @tdealtry BTW the WCSim validation macro does not include `/WCSim/PMT/PositionFile data/mPMT_Position_WCTE.txt` before `/WCSim/Construct` so the new PMT positions are not reflected in the validation.

Yeah I think the validation files have become outdated. I recommend that a WCTE person handles them, as you're the experts!

So in order to do that
- Update [this file](https://github.com/WCSim/Validation/blob/master/Generate/macReference/wcsim_wcte_SensitiveDetector_Only_NoTrigger_gun500MeVem_xdir_dark0_n800_seed20230628.mac#L8) (and the other WCTE ones)
- I'll merge that WCSim/Validation PR
  - I'll also update the HK FD geometries to use the realistic ones (It's been on my mind, I just keep forgetting)
- I'll merge this new geometry PR (when PMT ID discussion resolved)
- I'll issue a new WCSim release, which will ensure the validation reference files are updated

Whilst we're on the subject, if the 3 physics validations for WCTE (5 MeV electrons, 500 MeV electrons, 500 MeV muons) are not optimal for WCTE, please do change them too. In order to do this
- update the validation mac file(s)
  - you may need to add a file in the `physics/` mac file directory
  - rename the mac files to reflect the new setup. Also rename the output root filename
  - Remember to change the number of events are required (set based on run time, anywhere from a couple of minutes to a couple of hours)
- update [tests.txt](https://github.com/WCSim/Validation/blob/master/tests.txt) to point to the new file names (mac full filename & root filename stub), and give it a sensible updated name (that'll appear on the validation webpage)
- If you're changing the total number of tests defined
  - Change the WCSim/WCSim github action files to let it know about the new jobs (note that the ordering is different: 0 to N-1 for [reference file creation](https://github.com/WCSim/WCSim/blob/develop/.github/workflows/on_tag.yml#L112), 1 to N for [validation checks](https://github.com/WCSim/WCSim/blob/develop/.github/workflows/c-cpp.yml#L127))

kmtsui:
> So in this PR I tried to make minimal changes to WCSim, since I assumed that would be least likely to cause any problem for any existing analysis or code. Do you know if the existing numbering is what people will be using already in other code? Or did it change at some point recently?

I am not aware of anyone using the existing numbering yet. If the numbering scheme is not fixed yet, I prefer using the old conventions. In this case, the numbering should go from 1 to 32 for section 0, 49 to 69 for section 1, 33 to 48 for section 2, 86 to 106 for section3, 70 to 85 for section 4.

nickwp:
> I am not aware of anyone using the existing numbering yet. If the numbering scheme is not fixed yet, I prefer using the old conventions. In this case, the numbering should go from 1 to 32 for section 0, 49 to 69 for section 1, 33 to 48 for section 2, 86 to 106 for section3, 70 to 85 for section 4.

In this case I've committed a change that reverts the numbering to the scheme described here. I note that this scheme goes from 1 to 106, whereas the other goes from 0 to 105, but I assume that is as expected.

nickwp:
> Yeah I think the validation files have become outdated. I recommend that a WCTE person handles them, as you're the experts!

I've made a PR [here](https://github.com/WCSim/Validation/pull/26) updating the three WCTE validation macro files.

For now I think the 5 MeV electrons, 500 MeV electrons, 500 MeV muons are fine. I'm sure they work well for the purposes of ensuring there's no unexpected consequences to WCSim changes for WCTE.

tdealtry:
Thanks @nickwp 
I just bring up one question - have you verified that there are no unexpected off-by-one errors in using `1 to N` (rather than `0 to N-1`) numbering? This kind of thing has bitten on WCSim in the past

I'm testing the new CI on #407, but will then come to this

nickwp:
>  I just bring up one question - have you verified that there are no unexpected off-by-one errors in using 1 to N (rather than 0 to N-1) numbering? This kind of thing has bitten on WCSim in the past

So my understanding from Ka Ming is that the numbering this PR reverts to (`1 to N`) is actually the original nubering that WCSim used for a long time.

A PR a few months back (#385) changed the mPMT numbering to `0 to N-1` and as far as we're aware that has not actually been adopted yet in any WCTE software/analysis, so it's safer to revert to the old `1 to N` numbering for now.

Once WCTE has decided what to actually use in future, we can reintroduce the numbering of #385 (or whatever other numbering is chosen) and at that point should check thoroughly for any issues it causes downstream.

tdealtry:
Sounds good. Merging now. Release soon

PR #408, opened 2024-02-01
Revise cap logical volume definition, and properly implement CDS into WCTE geometry

kmtsui:
**Item 1**
After the update of WCTE geometry dimensions and PMT positions (https://github.com/WCSim/WCSim/pull/408), it is found that some cap PMTs are placed outside the mother volume `logicWCCap`(white volume in the bottom endcap).
image
To avoid further complications in future update, within the no-replica construction method, I decided to completely remove the `logicWCBarrelBorderRing` and `logicWCCap` volumes inside `logicCapAssembly`, and directly place the blacksheet and PMTs (and OD Tyvek if any) in `logicCapAssembly`.  This should make the code cleaner without affecting the actual functionality (the ring and cap volumes were designed to accommodate the replica placement).

**Item 2**
The CDS in a mechanical system to deploy calibration sources in WCTE. In normal run configuration, it is located close to the top endcap, and two of the mPMT modules are removed to allow the placement. To use the CDS in simulation, first we need to download the CAD model from WCTE webpage and set the input file path, then disable the two top cap mPMTs to in the mPMT position file. The instructions are written in `macros/WCTE.mac`. For now we do not enable the CDS placement by default, so it will not change the validation process. I think we want to understand its effect in simulation first and find a proper way to validate its placement and properties. The figure below shows the current placement (few more top cap PMTs are removed for clearer visualization)
imageimage


nickwp:
> After the update of WCTE geometry dimensions and PMT positions (#408), it is found that some cap PMTs are placed outside the mother volume `logicWCCap`(white volume in the bottom endcap).

Sorry about missing this, and thanks for spotting it. Checking my logs I had a typo when trying to enable the overlap checking, and failed to notice the CMake warning `Manually-specified variables were not used by the project: WCSim_Geometry_overlaps_CHECK`.

Note that even with that incorrectly set, it still does one overlap check (on the `pmt` placement) so I also didn't notice it wasn't fully checking overlaps. I'm a bit surprised it didn't give any stuck track exceptions which usually are very abundant with geometry overlaps, but perhaps in this particular case it doesn't cause that problem.

Given that I failed to properly perform this check, it might be worthwhile trying to implement an additional CI check that does so automatically? It could perform a run with`-DWCSim_Geometry_Overlaps_CHECK=ON`, run WCSim building the geometry, grep for any `*** G4Exception : GeomVol1002` in the output, and fail if it finds any? I'll see if I can figure out how to implement that.

Anyway, your solution for this overlap seems sensible, I'll try to check it more thoroughly as soon as I can.

tdealtry:
Thanks @kmtsui I'm happy with this from the HK side (it doesn't effect the FD/IWCD geometries) so you / @nickwp can merge when you're happy
Do just check you're happy with the [PMT position changes](https://wcsim.github.io/Validation/cbe604da7deaca15966643395f8e05932cecb0ed/21/geofile_NuPRISMBeamTest_16cShort_mPMT.txt.diff.txt) e.g. mPMT:106/3":19 changes from global PMT ID 2014 to 1710

@nickwp actually similar checks are already done & WCSim is compiled with the extra geo checks by default in the CI. See [this loop](https://github.com/WCSim/Validation/blob/master/MakeReference.py#L62) for the 2 overlap related checks we currently run (more can be added as necessary, of course - please put in a PR to add the extras. This becomes a red failure in place of the green "Num stuck track diff pass" on the webpage when there's a failure.

There's actually also [a check](https://github.com/WCSim/Validation/blob/master/Tests.sh#L362) for impossible geomety printout too, which again causes red/failure in place of "Geometry warnings pass". This is probably the best place for the new check you're suggesting (since it's a geo check, rather than a track check). Please PR in Validation. Thanks!

kmtsui:
> Do just check you're happy with the [PMT position changes](https://wcsim.github.io/Validation/cbe604da7deaca15966643395f8e05932cecb0ed/21/geofile_NuPRISMBeamTest_16cShort_mPMT.txt.diff.txt) e.g. mPMT:106/3":19 changes from global PMT ID 2014 to 1710

Thanks for pointing this out. Just rearrange the placement code to get back the old ordering.

tdealtry:
Great. Thanks @kmtsui Release coming soon

PR #409, opened 2024-02-01
Update GNUmakefile

kmtsui:
Add back the missing class in old fashion GNUmakefile

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -60,9 +60,9 @@ all: rootcint shared libWCSim.a lib bin
 
 ROOTSO    := libWCSimRoot.so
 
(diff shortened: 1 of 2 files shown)


tdealtry:
Thanks @kmtsui 

I'm going to merge this without waiting for the full CI, since it doesn't effect the CI build (that uses cmake) & this Makefile build is not generally recommended (everyone on HK & IWCD should use cmake)

Note that the CI is failing on WCTE because you've not pulled from #405 yet, so this is using the outdated WCTE PMT file. In future, please pull from WCSim/WCSim:develop to avoid this

PR #413, opened 2024-02-19
rootracker input file units

arturof:
WCSim was crashing after printing several times that the event was being generated outside the detector. Problem was fixed by setting position and energy units in GeV and cm using CLHEP/Geant4 convention.

Tested the code with rootracker vector files used for the MC production documented here: https://wiki.hyperk.org/do/view/HyperK/MCforFeb. They can also be found in sukap at: /disk1/disk02/usr6/yonenaga/production/vector/

After this change, WCSim doesn't crash and the true neutrino energy and vertex position distributions extracted from the output ntuple look as expected.

check_units_neg_genE

check_units_neg_genX


tdealtry:
Thanks for the fix @arturof

PR #416, opened 2024-02-26
Make all mPMT rotations consistent for WCTE

nickwp:
In the current input file for mPMT positions and rotations for WCTE, the mPMTs on the barrel are rotated by -90 degrees, while those on the endcap are rotated by a mix of -90, 0, +90 and 180 degrees:
Untitled
This appears to have come from an old solidworks design that is not realistic.

Since the mPMTs have 6-way symmetry, a +/-90 degree rotation of neighbouring mPMT breaks the symmetry, so avoiding that would be preferred. The symmetry is useful for ML reconstruction, which makes assumptions about each mPMT having PMTs in equivalent locations.

The actual mPMT rotations for the real detector haven't been finalised yet, nd affect the cable routing plan, but discussing with Thomas Lindner and Oliver Jeremy it should be possible to make the mPMT orientations consistent (at least without 90 degree rotations).

This PR simply sets all rotations to -90 degrees to make everything consistent with the barrel mPMTs.
The endcaps now have consistent mPMT orientations:
Untitled-1

I also checked that after this change, it works with the ML code that assumes consistency (where the same check failed for events generated with the current/old version).


nickwp:
The CI is expected to have some failures (only for WCTE) because the locations of some PMTs will have changed.

nickwp:
The CI had the expected differences. But it also identified five additional killed photons:

>  Optical photon is killed because of missing refractive index in either Aluminum or SilGel (transition from reflectorWCPMT to pmt) : could also be caused by Overlaps with volumes with logicalBoundaries.

Rotating the mPMT shouldn't cause this, and overlap checks found nothing, so it might just be some rare situation that happened to appear.

Neither @kmtsui nor I can reproduce it locally, but @kmtsui suggested uncommenting [this line](https://github.com/WCSim/WCSim/blob/abebcedc6216a942325484de65b1633d05b5b6fb/src/WCSimConstructMaterials.cc#L1487)

nickwp:
The above fix didn't seem to change anything, the CI logs still have the same killed photons and in fact everything is identical, so the uncommented lines seemed to have no effect.

tdealtry:
This is confusing. I've had a look through the construct materials & can't see anything obvious that's missing.

Did you fail reproducing the error
- With [the file the CI uses](https://github.com/WCSim/Validation/blob/master/Generate/macReference/wcsim_wcte_SensitiveDetector_Only_NoTrigger_gun500MeVem_xdir_dark0_n800_seed20230628.mac)?
- Inside a container?
- Compiled [with debug & geometry overlap checks on](https://github.com/WCSim/Validation/blob/master/build.sh)? (I really really hope that this won't cause differences...)

If you can reproduce it locally, then I recommend trying to up the verbosity & running again. e.g. [this line](https://github.com/WCSim/WCSim/blob/develop/WCSim.mac#L4) to get a little more information about where the dodgy tracks are

PR #417, opened 2024-02-27
Fix unique copyNo bug in pmt placement

kmtsui:
When non-replica placement is used without the PMT position table, the copyNo in endcap PMT placement is not assigned properly and some are duplicated. This makes the hit registration problematic. Now the bug is fixed.

Note that the default WCTE macro already includes the PMT position table, so this change will not affect the default behavior.


tdealtry:
Great thanks @kmtsui

PR #418, opened 2024-02-28
Unify mPMT orientations and set short absorption length for mPMT reflector aluminium to workaround missing refractive index

kmtsui:
Attempt to fix the missing refractive index issue in PR https://github.com/WCSim/WCSim/pull/416, but the refractive index table defined in `WCSimConstructMaterials.cc` is sometimes less than 1, which causes photon speed greater than 1 and aborts the event.
```
Starting event 462
  G4ParticleChange::CheckIt    : the velocity is greater than c_light  !!
  Velocity:  1.28523
opticalphoton E=2.49769e-06 pos=1.51383, -0.14082, 0.019856
      -----------------------------------------------
        G4ParticleChange Information  
      -----------------------------------------------
        # of 2ndaries       :                    0
      -----------------------------------------------
        Energy Deposit (MeV):                    0
        Non-ionizing Energy Deposit (MeV):                    0
        Track Status        :                Alive
        True Path Length (mm) :                0.019
        Stepping Control      :                    0
    First Step In the voulme  : 
    Last Step In the voulme  : 
        Mass (GeV)   :                    0
        Charge (eplus)   :                    0
        MagneticMoment   :                    0
                :  =                    0*[e hbar]/[2 m]
        Position - x (mm)   :             1.51e+03
        Position - y (mm)   :                 -141
        Position - z (mm)   :                 19.9
        Time (ns)           :             4.93e-05
        Proper Time (ns)    :                    0
        Momentum Direct - x :               -0.799
        Momentum Direct - y :                0.153
        Momentum Direct - z :                0.582
        Kinetic Energy (MeV):              2.5e-06
        Velocity  (/c):                 1.29
        Polarization - x    :               -0.571
        Polarization - y    :                0.113
        Polarization - z    :               -0.813

-------- EEEE ------- G4Exception-START -------- EEEE -------
*** G4Exception : TRACK003
      issued by : G4ParticleChange::CheckIt
momentum, energy, and/or time was illegal
*** Event Must Be Aborted ***
-------- EEEE -------- G4Exception-END --------- EEEE -------
```
Then I tried to give a very short absorption length in Al so that any photon traveling in Al will be killed. Now no missing refractive error is seen, but it kills some tracks and thus changes the RNG sequence.
Apart from WCTE, IWCD validation results are a bit off due to random fluctuations. HKFD are not affected though, because there are no reflectors in those mPMTs.

My hypothesis is that there are rare(?) occasions where photons are traveling inside Al, so the original missing refractive index error is due to photon moving from Al to SilGel. Either we can use this short absorption length, or we need a proper treatment of refractive index.

@tdealtry @nickwp Any thought?


tdealtry:
Thanks for looking into the @kmtsui 
(Just to let you know for future reference, editing a message then @-ing me doesn't notify me, so I missed this. Please add comments)

The proper way to fix this is to get the refractive index for Al. Hopefully this is findable in the literature somewhere

If this can't be found, then this hack is probably fine. In this instance, do make sure to add a clear comment to the `.cc` file to warn that this is a horrible hack, and that it should be fixed by adding a realistic refractive index for Al.

kmtsui:
The Al refractive index table defined in `WCSimConstructMaterials.cc` is actually from a seemingly legit [source](http://refractiveindex.info/?shelf=3d&book=metals&page=aluminium). The root cause for this problem is somehow a photon is produced and move inside the Al material. Maybe the electron moves through the reflector and generates Cherenkov photons?
It should be fine to use a short absorption length and Al indeed has a short one, but I will put a comment in .cc

tdealtry:
Ok fair enough (I still find it weird that a legimate source is breaking special relativity). I'm happy with this PR to be merged now - I'll let you merge when you're ready

nickwp:
I think the reason for this unphysical refraction is that the source provides a complex refractive index and it seems that only the real part was implemented. The G4 userguide has [an explanation](https://geant4-userdoc.web.cern.ch/UsersGuides/ForApplicationDeveloper/html/TrackingAndPhysics/physicsProcess.html?#complex-index-of-refraction) of complex refraction indices. It seems to suggest that the complex refractive index will allow G4 to calculate reflectivity from that (and presubably refraction too). But it doesn't use it to calculate absorption so we still need to define that. So if we really want to accurately model photons that propagate inside the aluminium, maybe we need to do it this way. But most likely we don't really care about photons propagating inside the aluminium since they should basically always just be absorbed.

kmtsui:
@nickwp I think we are ready, right?

nickwp:
I only just had time to more carefully check the validation logs and plots.

I see a lot of stuck track exceptions in IWCD, but that's the same as before (well it's now slightly increased for IWCD and slightly decreased for IWCDOD but roughly similar number of errors. So I think it's beyond this PR to fix those.

The other differences appear to be RNG differences as you mentioned.

It's difficult to look through the very long logs, but I didn't find anything unexpected searching for "exception" etc.

I'll rename this PR to make it clear it also includes the mPMT orientation changes, since it absorbed the changes from #416. So that PR can be closed and this one merged.

PR #424, opened 2024-03-25
PMT placement behind blacksheet

kmtsui:
This PR adds a new feature to allow the PMT volume to be placed behind the blacksheet, which means photons can enter the PMT volume even it is behind the blacksheet and properly propagate without being killed. This will be necessary when we implement the realistic mPMT model to properly track the photon reflection inside. The ID (blacksheet) dimensions are updated to reflect the more accurate (but not yet final) WCTE design.
Besides there is  a minor bug fix to install the missing file `data/mPMT_Position_WCTE_CDS.txt`

The placement logic works by:

1.  define `pmt_blacksheet_offset` which specifies how much the PMT logical volume is placed behind the blacksheet
2. properly define world dimensions to accommodate the intruding PMTs
3. Place the PMTs with offset + Use `G4MultiUnion` to combine all PMT `G4VSolid`
4. Subtract the PMT overlapping volume from blacksheet, finally place the blacksheet (and CDS) in proper positions.

Figures below show how the intruding PMTs look like in event display.
Side view of detector edge. White lines show the blacksheet volume
image
Top view
image


kmtsui:
hmm this geant4 is old and not compiled with `G4GEOM_USE_USOLIDS`, need to find workaround...
Or recompile geant4 in container?

tdealtry:
This looks great @kmtsui 

As a minimum change, can you see if swapping `G4MultiUnion` to `G4UMultiUnion` makes this compile & run?

From the [G4.10 release notes](https://geant4.web.cern.ch/download/release-notes/notes-v10.0.0.html#4.) you can
- Use a `UShape` on a case by case basis with `G4GEOM_USE_USOLIDS=OFF`
- Or use `UShape` for everything with `G4GEOM_USE_USOLIDS=ON`

Basically I'm worried that swapping all `Shape`s to `UShape`s will change output significantly. Of course, if someone would like to test this out, then please let me know & I can explain how to run the validation outside the CI

kmtsui:
I tried using `G4UMultiUnion` but it is useless without `G4GEOM_USE_USOLIDS`. Now try to import and define the class directly in WCSim...

tdealtry:
Ahh it's a shame that the simple way didn't work.

Hopefully the G4 in WCSim works. If not, you'll have seen the email I just sent to see if it's easy to setup a container with the new G4 flag, which will be a good backup
(If none of these work, it'll have to be run the CI locally)

nickwp:
Somehow I feel a bit uncomfortable hacking together different Geant versions (i.e. copying G4MultiUnion from Geant4.10.5 and using it with Geant4.10.3). Personally I feel that if we need this feature from a slightly newer Geant4 version, then unless there is a good reason to stick with 4.10.3 (which is six and a half years old now) we should probably put the work in to upgrade Geant4.

But I realise fully upgrading to a new Geant4 version could be a lot of work. Do you know what exactly is involved / required for upgrading Geant4 (even just one minor version, from 4.10.3 to 4.10.4, where G4MultiUnion was introduced)?

In a couple of weeks I should hopefully have time to volunteer to work on that. Besides, there might also be Geant bug fixes, performance improvements, or improved physics models / data from the past 6 years that we might benefit from.

kmtsui:
I do agree to move to more updated G4 if possible, but there could be lots of works validating the output and understanding the physics model differences. If there is any update plan we could possibly help.

For now the hack seems to work. Event display and geofile output look the same, no overlapping warning is seen. 
There are differences in wcsimrootevent plots but it is expected due to change in ID geometry (but why the CI checks are all green, the relevant plots are red on validation page)

tdealtry:
I'm not averse to updating G4 version, if someone has the time to do it. It will be nice to pickup some extra bug fixes, as well as reduce the hackiness. But ultimately, if this PR works for WCTE (and doesn't break HK) - which is what the CI shows, we could use it

The way this was done previously (4.10.1->4.10.3 when we did the merge with nuPRISM/WCSim) was comparing output for both versions, and doing the usual "are nhits & ndigits within around 1% different" (+ hit/digit time checks) for a variety of particle guns. My suggestion for how to do this would be
- add some extra particle guns to span a bit more physics in WCSim/Validation
- create a new WCSim/WCSim github action that we can run on button push on develop using 2 containers
  - the G4.10.3 one
  - the G4.10.5 one (I'm chatting with Mathieu about making this alternative container - will add you to CC @nickwp)

In terms of this PR, if this is the way we want to go, I don't see the following words - I think these need to be added somewhere ([point 2 here](https://geant4.web.cern.ch//download/license.html)) Probably to each of the copied G4 files.

> This product includes software developed by Members of the Geant4 Collaboration (http://cern.ch/geant4).

On the CI, @tjones79 & I tested & activated some new scripts last Wednesday. We made the mistake of forgetting to test the case where there is a failure in the CI.... I think I know the fix - will chat with Tom & hopefully fix it today

kmtsui:
Acknowledgement added, though I personally feel that the disclaimer in each file is already self-explanatory lol

tdealtry:
Agreed! But best to do what the license says

So this is ready (complete with fixed red CI, as expected). The question is then
1. Merge this as is
2. Upgrade G4 version & strip out the copy of the G4 files in this PR

@kmtsui @nickwp what are your preferences?

My feeling is, if there is enough person power to do this quickly enough for WCTE's needs, do 2. If not, do 1.

nickwp:
As much as I'd like to say that we could upgrade the G4 version quickly enough, it's maybe not worth the risk of delaying this too far if we do not get it done quickly. And the extra work to make a new PR with the copied G4 files removed once we do upgrade is minimal compared to just doing it in this PR.

The only worry is that without motivation to do it we will just delay much further the G4 upgrade, but overall I gues it's probably not worth worrying about since the features of this PR are actually needed sooner rather than later.

tdealtry:
Merging this now. Thanks @kmtsui 

I'm going to let the CI run (to check the CI fix works correctly for the errors we've been seeing for the past week over Easter). Then will issue a new release (to update the reference plots for WCTE)

PR #427, opened 2024-03-25
Output tree update

tdealtry:
- Turn off filling of true track/trajectory information in the 2nd (i.e. mPMTs in HK FD hybrid ID) & OD PMT types. This should save some file size
- Add option to select which PMTs' tree to analyse in `sample_readfile.C`
- Keep all tracks in the 0th trigger. Otherwise it is very confusing


tdealtry:
Actually this saves less file size than I thought because previously true track information was [only saved for PMT types that exist in the geometry](https://github.com/WCSim/WCSim/blob/develop/src/WCSimEventAction.cc#L965)

But this has made me notice that the CI for HK FD (both ID only & ID+OD) has produced no hits outside the 20" PMTs since v1.12.7... after much stress & confusion, I see this is a simple path error in the new set of WCSim/Validation mac files - basically the default SK geometry has been used because the new "simple" way to change geometries I setup failed
Really annoying that this was a warning not an error from geant...

I'm going to push something trivial here now, to check the fix in WCSim/Validation works. Expect a new WCSim release (= new reference files) tomorrow

PR #428, opened 2024-03-29
Fix photon history saving under new FillRootEvent structure

kmtsui:
The photon history tree filling is broken by https://github.com/WCSim/WCSim/pull/427
This is the minimal fix to make it work again


kmtsui:
CI errors come from `"Optical photon is killed because of missing refractive index"` counting, seems to be the same as previous tag run https://github.com/WCSim/WCSim/actions/runs/8448519973

tdealtry:
Thanks @kmtsui

PR #430, opened 2024-04-09
Update G4MultiUnion and G4Voxelizer

kmtsui:
When testing the new mPMT model to be used in WCTE, there were unknown overlapping warnings related to `G4MultiUnion`. After some investigation, I was pointed to a [bug report](https://bugzilla-geant4.kek.jp/show_bug.cgi?id=2301) which suggested the "coordinates of bounding box of each node" were calculated incorrectly. Only one line has to be fixed, but the bug fix was implemented in geant4 v10.7.2, so I updated the whole custom class anyway, and the warnings were gone. 

Again it shows a geant4 version update is probably favorable for picking up bug fixes and new functionalities, but manpower is always limited (；一_一)

### DIFF
--- a/include/G4MultiUnion_v1072.hh
+++ b/include/G4MultiUnion_v1072.hh
@@ -29,21 +29,20 @@
 // GEANT 4 class header file
 //
 // 
-// G4MultiUnion_v1051
+// G4MultiUnion_v1072
 //
 // Class description:
 //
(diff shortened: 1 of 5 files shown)


tdealtry:
Thanks @kmtsui 

As mentioned before, if someone kind find some time to validate, happy to update G4 version properly

PR #431, opened 2024-04-15
Iwcd geom update

kmtsui:
Update IWCD geometry based on [discussion](https://hkdbweb.in2p3.fr/doc/meetingfiles/A00039782/IWCD_240411_RyosukeAkutsu_mPMTLayout.pdf) from integration meeting.

Nominal configuration is Option B, which has 2 * 12 * 11 barrel mPMTs, and 52*2 cap mPMTs. Options A and C are backup if Option B does not give good photo-coverage, but they require position table inputs, and OD cap PMT numbers are probably not right. PMT positions are approximate anyway, will be updated when mechanical design is available.
image
image
image


tdealtry:
Looks good. Thanks @kmtsui 

A minor question - would it be useful to retain the ability to simulate the old detector design?

kmtsui:
There are "older" designs before this, but this is also the first one with OD so is somewhat special. I will probably just make a `SetIWCD_WithOD_Geometry_Old()` for simplicity.

tdealtry:
Sure thing. If you'd like to merge this now & re-add the old geometry later (but soon!) you can - if fitqun tuning folk are desperate for this new geometry

Otherwise, it "feels better" to wait for this old geometry to be available before merging

As usual, geometry change = CI change = will issue a new release after the merge

kmtsui:
Just pushed the old iwcd geometry option. Will let the CI run before merge.

PR #433, opened 2024-04-23
NuHepMC Reader and Generator

JackFannon:
A generator to read the [NuHepMC file standard](https://arxiv.org/pdf/1912.08005.pdf) that is based on the [HepMC3 file format](https://gitlab.cern.ch/hepmc/HepMC3).

The generator is selected in the macro file via `/mygen/generator hepmc3`; the file is specified using `/mygen/hepmc3file <FILEPATH>`. By default, the generator will read vector positions from the file, though in the case that isotropic positions need to be generated the command `/mygen/hepmc3PositionGen true` can be added to the macro file.

This now requires that there is a version of HepMC3 installed. Instructions for this can be found on the [HepMC3 GitLab page](https://gitlab.cern.ch/hepmc/HepMC3). Installing using a package manager is the easiest and most foolproof way of doing this.


CMakeLists.txt
```diff
@@ -33,11 +33,19 @@ pbuilder_prepare_project()
 set(CMAKE_CXX_STANDARD 14) # Enable c++14 standard
 
 
+########
+# HepMC3
+########
+find_package(hepmc3 REQUIRED)
+if (NOT hepmc3_FOUND)
+    message(FATAL "Unable to find HepMC3")
+endif (NOT hepmc3_FOUND)
```
```suggestion
find_package(HepMC3 REQUIRED)
if (NOT HepMC3_FOUND)
    message(FATAL "Unable to find HepMC3")
endif (NOT HepMC3_FOUND)
```

This was required in the hk-software container we use for WCSim to work. This is after doing 
```
sudo yum install epel-release
sudo yum install HepMC3 HepMC3-devel HepMC3-search HepMC3-search-devel HepMC3-interfaces-devel HepMC3-doc
```

Can you test in your setup if this works? If it doesn't, then different versions and/or installations methods of `HepMC3` provide different capitalisations of these files, which is fun...
```
    hepmc3Config.cmake
    hepmc3-config.cmake
```

I'm sure we can work around using `if` in this file, but not pleasant

FYI I installed `HepMC3-3.2.7-3` on `yum` today

cmake/CPM.cmake
tdealtry, line 1: Not used, please delete

include/WCSimPrimaryGeneratorAction.hh
```diff
@@ -107,7 +108,8 @@ private:
   G4bool   useRadioactiveEvt; // F. Nova: Radioactive flag
   G4bool   useRadonEvt; // G. Pronost: Radon flag
   G4bool   useLightInjectorEvt; // L. Kneale injector with profile from db
-  G4bool   useMPMTledEvt;
+  G4bool   useMPMTledEvt;r
```
tdealtry:
```suggestion
  G4bool   useMPMTledEvt;
```
Typo!

src/WCSimPrimaryGeneratorMessenger.cc
```diff
@@ -44,6 +45,20 @@ WCSimPrimaryGeneratorMessenger::WCSimPrimaryGeneratorMessenger(WCSimPrimaryGener
...
   fileNameCmdCosmics->SetDefaultValue("data/MuonFlux-HyperK-ThetaPhi.dat");
 
+  hepmc3fileNameCmd= new G4UIcmdWithAString("/mygen/hepmc3file",this);
+  hepmc3fileNameCmd->SetGuidance("Select the file of HepMC3.");
+  hepmc3fileNameCmd->SetGuidance(" Enter the file name of the HepMC3 file");
+  hepmc3fileNameCmd->SetParameterName("fileName",true);
+  hepmc3fileNameCmd->SetDefaultValue("inputhepmc3file");
+
+  hepmc3positionGenCmd = new G4UIcmdWithABool("/mygen/hepmc3positionGen",this);
+  hepmc3positionGenCmd->SetGuidance("Set to generate isotope positions or read from file.");
```
```suggestion
  hepmc3positionGenCmd->SetGuidance("Set to generate isotropic positions or read from file.");
```
typo

src/WCSimPrimaryGeneratorMessenger.cc
```diff
@@ -44,6 +45,20 @@ WCSimPrimaryGeneratorMessenger::WCSimPrimaryGeneratorMessenger(WCSimPrimaryGener
...
+  hepmc3positionGenCmd->SetGuidance("true : generate positions, false : read from file");
```
```suggestion
  hepmc3positionGenCmd->SetGuidance("true : generate positions randomly, false : read from file");
```

Also... generate positions where? Tank + rock? Entire water volume tank? ID only? ID FV?

src/WCSimPrimaryGeneratorMessenger.cc
```diff
@@ -44,6 +45,20 @@ WCSimPrimaryGeneratorMessenger::WCSimPrimaryGeneratorMessenger(WCSimPrimaryGener
   fileNameCmdCosmics->SetParameterName("fileName",true);
   fileNameCmdCosmics->SetDefaultValue("data/MuonFlux-HyperK-ThetaPhi.dat");
 
+  hepmc3fileNameCmd= new G4UIcmdWithAString("/mygen/hepmc3file",this);
+  hepmc3fileNameCmd->SetGuidance("Select the file of HepMC3.");
+  hepmc3fileNameCmd->SetGuidance(" Enter the file name of the HepMC3 file");
+  hepmc3fileNameCmd->SetParameterName("fileName",true);
+  hepmc3fileNameCmd->SetDefaultValue("inputhepmc3file");
+
+  hepmc3positionGenCmd = new G4UIcmdWithABool("/mygen/hepmc3positionGen",this);
```
tdealtry: I don't think I like the `.mac`-visible name for this - it's not clear enough what the difference between true/false is. I feel like it needs the word random in it. Maybe `/mygen/hepmc3positionGenRandomly`

Alternatively, could call it `/mygen/hepmc3positionGenMode` and use `0: use positions from the hepMC file; 1: generate positions randomly in X volume`

Suggestions welcome!

src/WCSimPrimaryGeneratorAction.cc
tdealtry, line 1: Please add an `else if` inside `SaveOptionsToOutput()` in order to save the input nuhepmc file used in the output file. Very useful for bookkeeping

tdealtry, line 1: Please add an `else if` inside `GetGeneratorTypeString()`

tdealtry, line 1: Please initialise `useHepMC3Evt = false` inside the constructor

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -1552,6 +1558,63 @@ void WCSimPrimaryGeneratorAction::GeneratePrimaries(G4Event* anEvent)
...
 
+    } else if (useHepMC3Evt) {
+
+      G4cout << "Using HepMC3 event" << G4endl;
+      // Check if the WCSimNuHepMC3Reader object has been initiaited yet
+      if (!hepmc3_reader) {
+        hepmc3_reader = new WCSimNuHepMC3Reader(hepmc3_filename, myDetector);
+      }
+
+      // Get the next event from the reader
```
tdealtry, line 1568: For my understanding, can one nuhepmc event include multiple neutrino interactions (vertices)?

I'm going to assume no in my next few comments

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -1552,6 +1558,63 @@ void WCSimPrimaryGeneratorAction::GeneratePrimaries(G4Event* anEvent)
...
+        // If particle has status 4 then it is a beam particle. This needs writing out, but not simulating
+        if (part->status() == 4) {
+          // Get direction (momentum) and normalise
+          G4ThreeVector dir(part->momentum().px(), part->momentum().py(), part->momentum().pz());
+          dir.set(dir.x() / dir.mag(), dir.y() / dir.mag(), dir.z() / dir.mag());
+
+          // Set write outs
+          SetBeamPDG(part->pdg_id(), 0);
+          SetBeamEnergy(part->momentum().e());
+          SetBeamDir(dir);
```
```suggestion
          SetBeamEnergy(part->momentum().e(), 0);
          SetBeamDir(dir, 0);
```
You explicitly put it in the 0th verex for `SetBeamPDG()` - do the same for the others

(of course if there are multiple vertices, the 0 should be replaced with a vertex counter...)

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -1552,6 +1558,63 @@ void WCSimPrimaryGeneratorAction::GeneratePrimaries(G4Event* anEvent)
...
+        }
+
+        // If the particle status is 1 then the particle needs simulating and writing out.
+        int n_final_state = 0;
```
tdealtry: This is initialised in the wrong place, since it resets every time through the `part:` for loop

But... I don't think it's actually needed.

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -1552,6 +1558,63 @@ void WCSimPrimaryGeneratorAction::GeneratePrimaries(G4Event* anEvent)
...
+          n_final_state++;
+
+          // Get direction (momentum) and normalise
+          G4ThreeVector dir(part->momentum().px(), part->momentum().py(), part->momentum().pz());
+          dir.set(dir.x() / dir.mag(), dir.y() / dir.mag(), dir.z() / dir.mag());
+
+          // Set write outs
+          SetBeamPDG(part->pdg_id(), n_final_state);
+          SetBeamEnergy(part->momentum().e());
+          SetBeamDir(dir);
```
tdealtry: These are outgoing particles, not incoming beam particles. See e.g. `muline` for how to set things in this context (it's different to particle gun)
```suggestion
```
After this, do please check that the output file does have the appropriate outgoing primary particles in it

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -1552,6 +1558,63 @@ void WCSimPrimaryGeneratorAction::GeneratePrimaries(G4Event* anEvent)
...
+      // Loop over the particles
+      for (HepMC3::GenParticlePtr part : hepmc3_reader->event.particles()) {
+
+        // Skip nuclear remnants
+        if (part->pid() == 2009900000) {
+          continue;
+        }
+
+        // If particle has status 4 then it is a beam particle. This needs writing out, but not simulating
+        if (part->status() == 4) {
```
tdealtry, line 1583: you check status
4: beam
1: outgoing

Is there a status x: target? If so, please use that to set
```
targetpdgs[iVertex]
targetenergies[iVertex]
targetdirs[iVertex]
```

include/WCSimNuHepMC3Reader.hh
tdealtry, line 1: You have comments on each function, which is great!
Can you change them to `///` instead of `//`, which should allow them to get into the doxygen docs?

include/WCSimNuHepMC3Reader.hh
```diff
@@ -0,0 +1,64 @@
...
+    // Vector of particles in event
+    std::vector<HepMC3::GenParticlePtr> particles;
+
+    // Populate particles using event.particles()
+    bool GetParticles();
+
+    // Return Lorentz vector of a specific particle
+    G4LorentzVector GetLorentzVector(int particle_index);
+
+    // Generate a random position in the detector construiction
```
tdealtry:
```suggestion
    // Generate a random position in the detector construction
```
typo

include/WCSimNuHepMC3Reader.hh
```diff
@@ -0,0 +1,64 @@
...
+    // Rotate the event
+    void RotateEvent();
+
+    // Return a vector of momenta and PDGs
+    std::vector<std::pair<G4LorentzVector, int>> GetMomentaAndPDGs();
+
+    // void GenDirection(double energy);
+
+    // Reader object
+    std::shared_ptr<HepMC3::Reader> my_hepmc_reader;
```
tdealtry, line 44:
```suggestion
    ReaderPtr my_hepmc_reader;
```
You `typedef` it, might as well use it

Alternatively, get rid of the `typedef`, as it doesn't appear to be used

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -1552,6 +1558,63 @@ void WCSimPrimaryGeneratorAction::GeneratePrimaries(G4Event* anEvent)
...
+    } else if (useHepMC3Evt) {
+
+      G4cout << "Using HepMC3 event" << G4endl;
+      // Check if the WCSimNuHepMC3Reader object has been initiaited yet
+      if (!hepmc3_reader) {
+        hepmc3_reader = new WCSimNuHepMC3Reader(hepmc3_filename, myDetector);
+      }
+
+      // Get the next event from the reader
+      hepmc3_reader->ReadEvent(hepmc3_positionGen);
```
tdealtry: `ReadEvent()` returns a `bool` to say whether it read the event successfully or not (technically whether there are any particles or not, but I presume it's the same thing)

I wonder if you can add a check here, and if there is no event, to call
```
 G4RunManager::GetRunManager()-> AbortRun();
```
(as is done for muline when the file is exhausted)

I'm thinking of the situation where the input file has 10 events, but the WCSim user has requested 1000 events via `beamOn`. This will be slow...

src/WCSimNuHepMC3Reader.cc
```diff
@@ -0,0 +1,128 @@
...
+    if (!GetParticles()) {
+        return false;
+    };
+
+    events_parsed++;
+    return true;
+}
+
+// Get particles
+bool WCSimNuHepMC3Reader::GetParticles() {
```
tdealtry: I don't like the name `GetParticles()`, since the return value isn't `vector<Particle>` or similar

A bad idea is this. Better suggestions are welcomed!
```suggestion
bool WCSimNuHepMC3Reader::ExtractParticlesFromEvent() {
```

include/WCSimNuHepMC3Reader.hh
```diff
@@ -0,0 +1,64 @@
...
+    // Generate a random direction
+    HepMC3::FourVector GenRandomDirection();
+
+    // Rotate the event
+    void RotateEvent();
+
+    // Return a vector of momenta and PDGs
+    std::vector<std::pair<G4LorentzVector, int>> GetMomentaAndPDGs();
+
+    // void GenDirection(double energy);
```
tdealtry:
```suggestion
```

include/WCSimNuHepMC3Reader.hh
```diff
@@ -0,0 +1,64 @@
...
+
+    // Generate a random position in the detector construiction
+    HepMC3::FourVector GenRandomPosition();
+
+    // Generate a random direction
+    HepMC3::FourVector GenRandomDirection();
+
+    // Rotate the event
+    void RotateEvent();
+
```
tdealtry:
```suggestion
```
Not defined (it suprised me that it doesn't complain about a missing symbol when running...)

include/WCSimNuHepMC3Reader.hh
```diff
@@ -0,0 +1,64 @@
...
+    // Return Lorentz vector of a specific particle
+    G4LorentzVector GetLorentzVector(int particle_index);
+
+    // Generate a random position in the detector construiction
+    HepMC3::FourVector GenRandomPosition();
+
+    // Generate a random direction
+    HepMC3::FourVector GenRandomDirection();
+
```
tdealtry:
```suggestion
```
Not defined (it suprised me that it doesn't complain about a missing symbol when running...)

tdealtry (changes requested):
This looks nice. Thanks @JackFannon !

There's quite a few comments inline (mostly to do with typos/variable naming) Please also
- As is tradition, document the new commands via example in `WCSim.mac` (commented out)
- Add instructions to the `README.md` saying that
  - `NuHepMC3` is a required dependency
  - Installation instructions are available at https://gitlab.cern.ch/hepmc/HepMC3#introduction
  - Suggest using the package manager instructions

I'm going to look at getting the CI running

tdealtry:
Got hepmc installing in the CI, so will commit a few of the typo suggestions to force it to run on this

tdealtry:
CI runs! 🎆 

In terms of testing this, I think it'd be good to do an equivalency test between the same NEUT events in rootracker & hepmc formats run through WCSim. If things are working correctly, things should be identical (so should pass the `complete_comparison.C` WCSim script).

@pabloferm mentioned in a [talk at the software meeting last week](https://hkdbweb.in2p3.fr/doc/meetingfiles/A00039940/20240418_NEUT.pdf) that there is a [conversion tool](https://github.com/neut-devel/neutvect-converter/tree/main) to switch between NEUT file types, so I think this is doable.

Pablo, I wonder if you have a pair of test files available such that we can do this test?

CMakeLists.txt
JackFannon: Sure, this works for me I'll commit it to the PR. I'm surprised the lowercase version doesn't work. The `find_package` command is meant to be case-insensitive.

src/WCSimPrimaryGeneratorMessenger.cc
JackFannon: I like `hepmc3positionGenMode`. It leaves it open for more options to be added in the future! Possibly other positions generation options, constant, generated from a flux etc.

pabloferm:
@JackFannon and @tdealtry , sorry for the delay. So nice to have this already implemented, thanks a lot! I do have a few files with both formats produced with https://github.com/neut-devel/neutvect-converter/

src/WCSimPrimaryGeneratorAction.cc
JackFannon: nuhepmc can have multiple vertices, but must only have one that has the primary interaction vertex status code (1). The other vertex codes are FSI Summary Vertex (2) and "Generator-dependent" (21-999). 3-20 are reserved for future NuHepMC standards and 0 is undefined and labelled as "Do not use"

src/WCSimPrimaryGeneratorAction.cc
JackFannon: Ah, thanks for that!

Is it not needed to set the energy and direction for the individual particles? There needs to be some kind of counter right? 

_edit_ - Just seen your comment below! Ignore this

include/WCSimNuHepMC3Reader.hh
JackFannon: Thanks for catching this. The `typedef` was left in after debugging an issue I was having with getting HepMC3 to play nicely with WCSim but, you're right, it's no longer needed

src/WCSimPrimaryGeneratorAction.cc
JackFannon: I've had a look through `muline`. Is it enough to just remove these calls to `SetBeam...`?

I can't see how `muline` is writing out the outgoing particles past the calls to `particleGun->...`

src/WCSimPrimaryGeneratorAction.cc
JackFannon: Added this using `if (part->status() == 20)`

include/WCSimPrimaryGeneratorAction.hh
```diff
@@ -251,6 +260,14 @@ private:
   inline void SetIBDModel(G4String choice) { ibd_model = choice; }
   inline G4String GetIBDModel()  { return ibd_model; }
 
+  // HEPMC3 reader
+  inline void SetHepMC3EvtGenerator(G4bool choice) { useHepMC3Evt = choice; }
+  inline G4bool IsUsingHepMC3EvtGenerator()  { return useHepMC3Evt; }
+  inline void SetHepMC3Filename(G4String choice) { hepmc3_filename = choice; }
+  inline G4String GetHepMC3Filename()  { return hepmc3_filename; }
+  inline void SetHepMC3PositionGen(G4bool choice) { hepmc3_positionGen = choice; }
+  inline G4bool SetHepMC3DirectionGen(G4bool choice) { return hepmc3_positionGen;}
```
tdealtry:
```suggestion
```
I *think* this isn't used

Alternatively, keep it but fix it! It currently sets `hepmc3_positionGen` (i.e. position, not direction)

CMakeLists.txt
```diff
@@ -103,9 +114,11 @@ endif()
 #########
 
 if (CMAKE_COMPILER_IS_GNUCC AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 4.7)
-    set(CMAKE_CXX_FLAGS " ${CMAKE_CXX_FLAGS} -std=c++11 -Wno-overloaded-virtual")
+    # set(CMAKE_CXX_FLAGS " ${CMAKE_CXX_FLAGS} -std=c++11 -Wno-overloaded-virtual")
+    set(CMAKE_CXX_FLAGS " ${CMAKE_CXX_FLAGS} -std=c++11 -w")
 else()
-    set(CMAKE_CXX_FLAGS " ${CMAKE_CXX_FLAGS} -std=c++0x")
+    # set(CMAKE_CXX_FLAGS " ${CMAKE_CXX_FLAGS} -std=c++0x")
+    set(CMAKE_CXX_FLAGS " ${CMAKE_CXX_FLAGS} -std=c++0x -w")
```
tdealtry: You're setting `-w` which [seems to be](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html) "Inhibit all warning messages."
Please don't!

tdealtry (changes requested):
Couple of easy (but important) fixes I missed the first time round

Otherwise looks good, other than
- please add something to the README about the new dependency
- we're discussing the beam/target/vertex stuff in chat

include/WCSimPrimaryGeneratorAction.hh
JackFannon: I must have been in a rush on the day I wrote this. It's meant to be `GetHepMC3PositionGen` and return the bool. Solved this now!

CMakeLists.txt
JackFannon: This was never meant to be committed 🤦

tdealtry:
Looks great. Thanks @JackFannon Think all that's left is confirming you get the same primary particles when using a rootracker file vs a nuhepmc files

JackFannon:
> @JackFannon and @tdealtry , sorry for the delay. So nice to have this already implemented, thanks a lot! I do have a few files with both formats produced with https://github.com/neut-devel/neutvect-converter/

Hi @pabloferm ! Would you please be able to share a couple of these files with me? I have a neut output file but I'm unsure in how to get this in the rootracker format. I only need a quick comparison so any pair of rootracker and hepmc3 files would be great.

pabloferm:
> > @JackFannon and @tdealtry , sorry for the delay. So nice to have this already implemented, thanks a lot! I do have a few files with both formats produced with https://github.com/neut-devel/neutvect-converter/
> 
> Hi @pabloferm ! Would you please be able to share a couple of these files with me? I have a neut output file but I'm unsure in how to get this in the rootracker format. I only need a quick comparison so any pair of rootracker and hepmc3 files would be great. 

Sure, how you should I share them with you? They are around 100 MB. I can send you a link if you give me your email.

JackFannon:
Yeah, email would be great. My email is ***@***.*** Thanks
for this!

JackFannon:
Hi, I've done some validations that @tdealtry and I discussed over Mattermost: 

I've compared the 3-momenta and initial positions between the HepMC3 input file and the WCSim output tree and they match up 1 to 1. The units for the position get converted from mm to cm inside WCSim at some point, though I think this is unavoidable without modifying the WCSimEventAction class. Let me know what you think and if you want to see any more checks! 
image

tdealtry:
Great thanks @JackFannon! Will merge this after we get the CI back up & running

tdealtry:
The CI is only failing for WCTE. I'm certain that this is because it's running on your branch (rather than the result of the merge), and you've not pulled in the recent WCTE changes. So I'm going to merge now
Thanks @JackFannon !

tdealtry (approved):

Issue #434, opened 2024-04-24
DarkRate .mac options not changing number nor distribution of the DarkNoise hits


DiegoCostas97:
Hi, I've already talked a bit about this with Ka Ming, and I don't know if this is a bug, but anyway I think this need to be discussed.

I'm working on Neutron Tagging, and developing an "algorithm" that finds neutron candidates within background (DarkNoise only). For the moment, the code works fine, and I wanted to know if it would still work fine with a bigger amount of DarkNoise hits, since it's my only background source.

I showed my macro to Ka Ming, and he said I surely need to fine tune the trigger window, and suggested to start using the NoTrigger option.

This is what I'm doing now, and the problem is that I'm seeing no change at all in the DarkRate.

I'm trying 12 different configurations, and see no change in none of them. I'm playing around with 3 main parameters:
- `/DarkRate/SetDarkRate 1 kHz` (default for the PMTs I'm using), 10 kHz and 1000 kHz
- `/DarkRate/SetDetectorElement tank` (on and off)
- `/DarkRate/SetDarkMode` (when 0, -400 to 950 window; when 1, 4000 window)

The thing is, with `SetDarkMode 1`, I see the same amount of DarkNoise hits wether I'm using 1, 10 or 1000 kHz (225800 hits when `SetDetectorElement` **is not set** to "tank" and 18153 when `SetDetectorElement` **is set** to tank).

With `SetDarkMode 0`, I see the same amount of DarkNoise hits wether I'm using 1, 10 or 1000 kHz (225800 again when `SetDetectorElement` **is not set** to "tank" and 3008 when `SetDetectorElement` **is set** to tank)

So you see, the only difference is made when `/DarkRate/SetDetectorElement tank` is or is not set, and `/DarkRate/SetDarkRate` has no effect. 

I've checked every output in the terminal and in every run the changes were correctly applied. I'm using WatChMaL DataTools pipeline to extract the .root data and filter using a Pandas DataFrame so I can count how many DarkNoise hits we have in every run.

What do you think?

Thanks in advance! Diego

I leave here the WCTE.mac and daq.mac that I'm using.

WCTE.mac:
```
# nuPRISMBeamTest WCTE macro with no visualization

/run/verbose 1
/tracking/verbose 0
/hits/verbose 0
#/control/execute vis.mac
#/vis/open OGLSQt

## select the geometry
# Default config if you do nothing is currently SuperK
[...]

PR #435, opened 2024-04-25
Update macros/WCTE.mac for correct DarkRate functioning

DiegoCostas97:
DarkNoise hits were not changing when `SetDarkRate` was changed in the macro file. Now it is.

### DIFF
--- a/macros/WCTE.mac
+++ b/macros/WCTE.mac
@@ -97,7 +97,9 @@
 #/random/setSeeds 2 1
 #/WCSim/random/seed 4
 
-
+# Note: SetDetectorElement **must** come before SetDarkRate, if not, changes in the 
+# DarkRate will not be applied
+#/DarkRate/SetDetectorElement tank
 # default dark noise frequency (and conversion factor) is PMT property (NEW), set in the code.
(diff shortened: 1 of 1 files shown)


macros/WCTE.mac
```diff
@@ -97,7 +97,9 @@
 #/random/setSeeds 2 1
 #/WCSim/random/seed 4
 
-
+# Note: SetDetectorElement **must** come before SetDarkRate, if not, changes in the 
+# DarkRate will not be applied
+#/DarkRate/SetDetectorElement tank
```
tdealtry:
```suggestion
# Note: SetDetectorElement **must** come before all /DarkRate/ options (e.g. /DarkRate/SetDarkRate), if not, changes specified
# (e.g. in the DarkRate) will not be applied
# (Technically they will be applied to the OD PMT dark parameters, which don't exist in WCTE = no effect is seen)
/DarkRate/SetDetectorElement tank
```

Just generalising the comment a bit more, and also turning it on by default, so people don't miss it in the future

tdealtry:
Thanks @DiegoCostas97 

Closes #434

PR #438, opened 2024-05-02
Add option to select WCSimRoot only installation

guiguem:
This allows to only install WCSimRoot libraries using the `-D WCSim_WCSimRoot_only=ON` cmake option.
When doing so, all the G4 dependencies won't be looked for, only looking for ROOT

This allows for packages to be able to decode the file format, but not run the simulation itself, allowing a cleaner split between the simulation and the reconstruction.
For HK, this should take the form of a new package `hyperk/WCSimRoot` used by fiTQun (and others...) instead of the full WCSim package.

### DIFF
--- a/.github/workflows/build_wcsimroot.yml
+++ b/.github/workflows/build_wcsimroot.yml
@@ -1,0 +1,82 @@
+name: Build WCSimRoot
+
+on:
+  push:
+    branches: [ develop ]
+  pull_request_target: #runs on the PR branch, rather than the merge branch. But is required for access to secrets
+    branches: [ develop ]
+  release:
+  
+concurrency:
+  #When a PR is pushed too multiple times in a short period, this would usually trigger multiple CI runs
+  #This setting cancels any in-progress CI when a new push is performed
+  #But we can have multiple jobs testing different things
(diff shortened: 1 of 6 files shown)


guiguem:
We could add a workflow for testing the compilation of this package with this option. I can put something together (in parrallel of the physics build one).

tdealtry:
This looks good. Thanks @guiguem 

The CI check would be great, if you can set that up

Could you also add to the README
- [In the list of cmake build options](https://github.com/WCSim/WCSim/blob/develop/README.md?plain=1#L142) describe the new option
- [Somewhere around here](https://github.com/WCSim/WCSim/blob/develop/README.md?plain=1#L64) add something like "If you just need to read WCSim output files, you can do this without the G4 dependency"

guiguem:
It seems that this PR won't trigger the new CI I made. Not sure why, it could be because this is a fork outside of WCSim org that tries to do stuff...

tdealtry:
Exactly how CI works on PRs that change CI is something I've not figured out yet...

The rest looks good, so I'll merge & see what runs. Thanks @guiguem

PR #439, opened 2024-05-02
Adapt CXX Standard based on ROOT

guiguem:
ROOT is being rather picky about the CXX standard and doesn't allow we use different ones for different packages. So here we check the standard from its flags and use this as base standard; not very gracious but very robust!


tdealtry:
Thanks @guiguem

PR #442, opened 2024-05-08
Implement ex-situ mPMT in WCTE

kmtsui:
This PR adds the ex-situ mPMT in WCTE. Major changes are

- Define new material and surface properties for mPMT construction
- New `WCSimDetectorConstruction::ConstructExSituPMT()` and `WCSimDetectorConstruction::ConstructExSituMultiPMT()` methods to construct the accurate 3-inch PMT and mPMT geometries
- Update placement code and WCTE mPMT position table to allow the use of ex-situ mPMTs
- Add new `WCSimWCSD::ProcessHits_boundary()` method to handle photons from InteriorWCPMT hitting photocathode


kmtsui:
Illustration of 3-inch PMT model
image
image
image

kmtsui:
Illustration of mPMT model
image
image

kmtsui:
Figures below show the cos angle between the photon ending direction and photocathode normal. First one is with the normal mPMTs where all values are positive. Second one is with the ex-situ mPMTs where photons can be absorbed on the back of the photocathode surface.
image
image

tdealtry (approved):
From my side, this looks good
- no geometry overlap warnings picked up from the CI
- no killed tracks or similar (suggesting other geometry overlaps)
- the code looks good

The one thing I will flag is that there is a [10% drop in true hits e.g. here](https://wcsim.github.io/Validation/b091160af203bc8b2a97ac935374d61c31a244b2/21/1/index.html). This is a significant change, so you'll want to discuss this with your WCTE colleagues (if you haven't already)

kmtsui:
Yes I noticed the drop in hit rates. Since this is the first time we use the new model in simulation, I cannot tell whether this level of reduction is correct or not. We will do some discussions this week, so can wait for a while before merging the PR.

kmtsui:
Found an air gap in the original model between PMT gel and acrylic dome. After correction, the hit rates are increased by 30%. Now the new model has around 20% more hits than the WCSim's default model.

image

kmtsui:
The extra photon detection at the inner photocathode surface can account for a bit less than half of the increased efficiency.

image

kmtsui:
Another reason for the increased efficiency is that the air volume inside mPMT now allows photon propagation. Upon killing photons in that air volume, the hit rates of the old mPMT model and the ex-situ model become similar.

Simulation dataset: Isotropic photon emission at WCTE tank center

image

kmtsui:
@tdealtry I think we can merge this PR first. The hit rate difference is investigated and shown in the threads above.
There will be another PR on the in-situ mPMT model, further tuning will be done there.

kmtsui:
The CI failed again... Problem with `Install HepMC3`

kmtsui:
@tdealtry let's merge this first, then I will put another PR for another (in-situ) mPMT model and further tuning

kmtsui:
@tdealtry Should we merge this? There will be another PR for another (in-situ) mPMT model and further tuning

tdealtry:
Oops sorry I missed a couple of "this is ready now" messages. Yes, let's merge

Just one question first. Is the WCTE geom setup macro for the CI [here](https://github.com/WCSim/Validation/blob/master/Generate/macReference/geom/wcte.mac) setup to be "the officially recommended best model"? My recommendation would be to setup the WCTE CI to mirror what people run by default. This might be the case, I'm just not plugged in to know

Similarly, since #453 I'd suggest making a WCTE version of `tuning_parameters.mac` in the validation that uses the default WCTE model (which I presume isn't the WCSim default). Steps would be
- add new tuning file [here](https://github.com/WCSim/Validation/tree/master/Generate/macReference/tune)
- add to **every** test in [this json](https://github.com/WCSim/Validation/blob/master/tests.json) a new `TuningMacName` argument linking to either `tuning_parameters.mac` or the new e.g. `tuning_parameters_WCTE.mac`
- Add a third argument to [this script](https://github.com/WCSim/Validation/blob/7aeeab988e6937ea9543b104c15ee6d247134a85/Generate/Generate.sh#L3) to replace `$ValidationPath/Generate/macReference/tune/tuning_parameters.mac`
- Add the third argument to [this python script](https://github.com/WCSim/Validation/blob/7aeeab988e6937ea9543b104c15ee6d247134a85/RunTests.py#L93) that calls `Generate.sh` for each physics comparison job, where the value of the third argument is taken from the new entry in the json
- Put a temporary hack in [this script that makes the reference files on a tag](https://github.com/WCSim/Validation/blob/7aeeab988e6937ea9543b104c15ee6d247134a85/MakeReference.py#L35) to use the right `tuning_parameters.mac` file. Something like `tuning_mac_name = 'tuning_parameters_WCTE.mac' if 'wcte' in mac.lower() else 'tuning_parameters.mac'`
  - Yes, this is a hack. But there's a plan to rewrite `MakeReference.py` soon to remove repetition with `RunTests.py`, and importantly, to use the same `tests.json` file to define the tests

Please submit a PR in WCSim/Validation if you'd like to make such changes.

Feel free to postpone this until after the next tag. Or do the easy geometry bit now, the harder tuning bit later. Or just ignore the suggestion altogther

Let me know either way!

tdealtry:
The reason I'm asking this now is that I'll issue a tag after this merge, such that the WCTE reference files are updated & tests aren't all red!

kmtsui:
> Just one question first. Is the WCTE geom setup macro for the CI [here](https://github.com/WCSim/Validation/blob/master/Generate/macReference/geom/wcte.mac) setup to be "the officially recommended best model"? 

Yes we will still use this `wcte.mac` as default.

> Similarly, since #453 I'd suggest making a WCTE version of `tuning_parameters.mac` in the validation that uses the default WCTE model (which I presume isn't the WCSim default). 

We still haven't decided which one to use as default, so I will keep using the WCSim one for now. Once we have the decision, I will add the WCTE specific tuning file in CI.

tdealtry:
Ok great. Thanks @kmtsui

PR #446, opened 2024-05-14
Hard exit when using a PMT QE method other than SensitiveDetector_Only

### LINKED ISSUE #445: Bugs with /WCSim/PMTQEMethod Stacking_Only
Maybe this is already known, but when we use `/WCSim/PMTQEMethod Stacking_Only` in detector with OD, the below 400 nm photon QE is boosted by the WLS absorption length, which causes too much ID photons to survive. 

Not sure if any physics study is using such configuration. We just found this bug in IWCD fiTQun tuning.

We may just show a warning in the macro option explaining this issue, or
we can fix it by applying a physical/logical volume check in `WCSimStackingAction::ClassifyNewTrack()` to separate the ID and OD photons.

tdealtry:
See discussion in #445

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -9,6 +9,7 @@ New feature
 * Pull request #438 @guiguem: Adds `-DWCSim_WCSimRoot_only=ON` cmake option to compile lighter-weight WCSimRoot library without Geant4 dependency, for just reading WCSim files
 
 Update
+* Pull request #446 @tdealtry: Hard exit when using a PMT QE method other than SensitiveDetector_Only
 * Pull request #439 @guiguem: Use the same C++ standard as was used to compile ROOT
 
 Bug fix
--- a/include/WCSimDetectorConstruction.hh
+++ b/include/WCSimDetectorConstruction.hh
@@ -207,14 +207,20 @@ public:
   void   SaveCaptureInfo(G4bool choice) {captureInfo_isSaved=choice;}
 
   void   SetPMT_QE_Method(G4int choice){
-    if(choice != 3) {
+    switch(choice) {
+    case 3:
+    case 4:
+      PMT_QE_Method = choice;
+      break;
+    default:
       G4cerr << G4endl << "******************************" << G4endl
-	     << "PMT QE method must be set to 3 (SensitiveDetector_Only)" << G4endl
+	     << "PMT QE method must be set to one of:" << G4endl
+	     << " 3 (SensitiveDetector_Only)" << G4endl
+	     << " 4 (DoNotApplyQE)" << G4endl
 	     << "Other methods are not fully tested" << G4endl
 	     << "Exiting..." << G4endl;
       exit(-1);
     }
-    PMT_QE_Method = choice;
   }
   void   SetPMT_Coll_Eff(G4int choice){PMT_Coll_Eff = choice;}
   void   SetVis_Choice(G4String choice){Vis_Choice = choice;}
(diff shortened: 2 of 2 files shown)


spradlin:
Would it make sense to also allow DoNotApplyQE?

tdealtry:
Good point. Thanks

nickwp:
Sorry I should have picked up on this before it got merged, but this change breaks the default WCTE config that [does use](https://github.com/WCSim/WCSim/blob/develop/macros/WCTE.mac#L78) Stacking_Only.
For WCTE, with only one PMT type and no WLS, I don't think that mode had any bugs and gives a slight speed up in simulation due to fewer photons needing to be simulated (which is the original reason for this mode existing).

But given the bugs that mode has in other situations, do we just change WCTE to use SensitiveDetector_Only for consistency and accept the performance hit? I think it's not very significant difference but I can try to test it.
@kmtsui do you have an opinion?

tdealtry:
No worries. I suppose could modify the exit to only exit if the detector name is not one of the WCTE ones. Feel free to put a PR in

But do of course ensure that `Stacking_Only` gives consistent results to `SensitiveDetector_Only` first!

kmtsui:
I am actually changing the macro default to `SensitiveDetector_Only` in PR https://github.com/WCSim/WCSim/pull/442. 
Personally I think simulating all the photons would be better as we care about reflections on different components, having more photons at the beginning should give us more realistic response, especially for low event events.
CPU time is probably not very important for WCTE since the detector is small.

nickwp:
I'm testing it now, and it seems it actually makes a fairly big difference to computation time.
100 muons in WCTE took:
144 seconds with `Stacking_Only`
473 seconds with `SensitiveDetector_Only`
This actually makes sense that the computation time is reduced to about 30%: the maximum QE is less than this and so the number of photons simulated will be reduced to less than 30%.

> Personally I think simulating all the photons would be better as we care about reflections on different components, having more photons at the beginning should give us more realistic response, especially for low event events.

It actually shouldn't make any difference to how realistic the response is, unless there's some process that changes the photon's wavelength (and therefore changes it's QE), or if we have multiple PMT types (and therefore don't know the QE of the PMT it will hit). Applying the QE to kill photons when they are created has exactly the same effect as applying the QE when they hit the sensitive detector. In either case you just randomly ignore the percentage of photons in the end. I beleive in either case the QE is applied before storing the CherenkovHits and CherenkovDigiHits, so the photons that don't pass QE have no effect on output regardless of when the QE is applied (during stacking or during sensitive detector). But if the `SensitiveDetector_only` QE is applied after storing the CherenkovHits, or if we're storing true photon tracks, then we would have more true hits or more photons tracks output in that case, and then maybe it's worth simulating the extra photons. But I don't think this is actually the case by default, even for `SensitiveDetector_Only`.

kmtsui:
Fair point. I can imagine Raman scattering is one of the processes that would change the wavelength, but it's not in WCSim or geant4. For normal usage in single PMT geometry, `Stacking_Only` should mostly work fine.

nickwp:
The suggestion of @tdealtry to only allow it for WCTE geometries is probably safe.
It could potentially also work for other geometries that use a single PMT type and don't have any WLS, but checking these things might be a pain. And like you say, other future developments like Raman scattering could introduce problems.
So I guess the question is whether a ~3x increse in CPU time is a problem for WCTE? It seems significant but maybe it's fast enough anyway that we don't really care?

kmtsui:
The production so far is fine with `SensitiveDetector_Only`, CPU time shouldn't be a problem unless we want a very large dataset. And maybe bottleneck is from fiTQun?

nickwp:
Yes I'm sure fiTQun will dominate the CPU time for a produciton...
So I guess we keep things as they are after this PR and #442 that will change the default WCTE macro to `SensitiveDetector_Only`

We can try to bear in mind for future that a significant performance boost could be gained by a correct implementation of `Stacking_And_SensitiveDetector`: Find the maximum QE value across all wavelengths and PMT types, apply this maximum QE in the StackingAction, then apply the remaining PMT and wavelength-dependent part of the QE in the SensitiveDetector.

PR #451, opened 2024-06-11
Fix comsic muon position and direction for NuPrism type detector

kmtsui:
For NuPrsim type geometry, the detector is rotated such that y-axis is the vertical axis of the cylinder. This makes the cosmic muon generator not working properly, i.e. muons are coming from the side of the detector instead of the top. 

This PR makes a quick fix for the muon direction and position. However, we still need the properly flux files for non-HK detector sites (IWCD, WCTE).


tdealtry:
Thanks for this @kmtsui 

Please do a pull from develop to pickup the changes to the workflow files I had to make (welcome to CentOS Stream 8 EOF)

kmtsui:
@tdealtry CI looks good, should be okay to merge

PR #452, opened 2024-06-11
Update where yum gets its info from

tdealtry:
CI is currently failing at the `yum install` stage

This seems to because CentOS Stream 8 (the CI OS) [became EOL on 31 May 2024](https://blog.centos.org/2023/04/end-dates-are-coming-for-centos-stream-8-and-centos-linux-7/)

Trying [this stackoverflow fix](https://stackoverflow.com/a/71309215)


tdealtry:
Let's merge this so we can try the fix...

PR #456, opened 2024-07-02
Prevent absolute path to header files being included in the WCSimRoot library

tdealtry:
As found in the HK WCSim container (see discussion [here](https://git.hyperk.org/hyperk/externals/WCSim/-/issues/2)) there is an issue where there are lots of warnings/errors when you attempt to use `libWCSimRoot.so` to read an output file if you have deleted the code source directory, as the absolute path to the headers are included in the library

This PR uses
- a relative path for the argument to `ROOT_GENERATE_DICTIONARY`
- the absolute path for the header installs

@guiguem I wonder if you have any CMake magic to improve this? i.e. to prevent the repition in the definition of `dict_headers` & `dict_headers_to_install`

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -32,7 +32,18 @@ endif()
 # WCSimRoot library
 
 set (dict_headers
-		${PROJECT_SOURCE_DIR}/include/jhfNtuple.h
+        jhfNtuple.h
+        WCSimRootEvent.hh
+        WCSimRootGeom.hh
+        WCSimPmtInfo.hh
+        WCSimEnumerations.hh
+        WCSimRootOptions.hh
+        WCSimRootTools.hh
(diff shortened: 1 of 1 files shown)


guiguem:
Hi @tdealtry I thought about this and look over some of my past example. it seems indeed one has to have absolute paths for this (leading to a duplicated list of headers which is rather annoying imo). if this solves the issue, I am not sure there is another solution.
The main issue is related to the install target we use for the installation and the basic root examples don't have this in their cmakeLists. maybe we can open an issue in the ROOT forum...

tdealtry:
Thanks for looking into this @guiguem I agree that it'd be better if there was a more elegant solution, but as you say, it works, so let's just use it for WCSim & move on (we can dig more if we require the functionality for other HK packages)

PR #459, opened 2024-07-16
in-situ mPMT model

kmtsui:
This is the promised geant4 model for the in-situ mPMT. One important feature is that the small PMT cathode glass positions are exactly the same in the ex-situ and in-situ models. 

Another minor change is to define the mPMT dome as the detector boundary point, since the blacksheet is removed at the mPMT placement positions.

Figures below show the structural difference between the ex-situ and in-situ mPMTs. 
image
image
image
image
image

The light acceptance is the highest in the ex-situ model, followed by the in-situ model, and the WCSim one comes in last.
image

While we have both ex-situ and in-situ models implemented, we have not decided the distribution of the two models. For now, we will just use the ex-situ alone in the default macro, so there should be no change in the validation output.


kmtsui:
@tdealtry CI is dead again?
```
  /usr/bin/docker --config /home/runner/work/_temp/.docker_4d8735fa-4988-4aca-aeca-51e92d70bc7e login ghcr.io -u kmtsui --password-stdin
  Error: Docker login for 'ghcr.io' failed with exit code 1
```

macros/WCTE.mac
tdealtry, line 1: Thanks for spotting & updating these. The G4 docs seem to change location rather regularly...

tdealtry (approved):
Looks good. Thanks @kmtsui 

(Yes the CI was due to a token that expired a few days ago that I wrongly thought was non-critcal... easy fix though)

PR #460, opened 2024-07-16
Include unistd header

gondiaz:
If not added, I get the following compillation error on MacOS (Sonorama 14.5)

```
/Users/gdiazlop/HK_Software/WCSim/src/WCSim.cc:32:36: error: use of undeclared identifier 'F_OK'
    bool exists = access(filename, F_OK) != -1;
```

### DIFF
--- a/WCSim.cc
+++ b/WCSim.cc
@@ -24,6 +24,7 @@
 
 #include <string>
 #include <sstream>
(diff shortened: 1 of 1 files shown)


tdealtry:
Looks good. Thanks!

PR #399, opened 2023-12-19
Addition of working assembly based hybrid mPMT geometry.

patrickstowell:
Merge of new realistic PMT geometry based on integration diagrams.
Issues with mPMT hits have been traced back to a rotation and offset issue in the WCPMT2 objects. This has been resolved and now weighted PMT hits are agreeing within 1-2%.

**Comparisons of the 500MeVmum_xdir validation set**
ORIGINAL PMT1 COUNTS [AVGNHIT,NPMT]: 2108.33 19826 
REALISTIC PMT1 COUNTS [AVGNHIT,NPMT]: 2059.1 19746
COMPARISON WEIGHTED HITS OLD VS NEW : 0.106341 0.104279 -> 0.98061

ORIGINAL PMT2 COUNTS [AVGNHIT,NPMT]: 599.71 188613
REALISTIC PMT2 COUNTS [AVGNHIT,NPMT]: 49.13 15504
COMPARISON WEIGHTED HITS OLD VS NEW : 0.003179 0.003168 -> 0.99654

PR #468, opened 2024-08-02
Feature/od coverages

federiconova:
- extending the range of propagated photons (the OD has wavelength-shifting plates, so we want to keep photons down to 200 nm even though the PMTs cannot directly see them)
- storing photon wavelength information
- allowing different Tyvek reflectivity values in different OD sections (top / bottom / barrel and inwall / outwall)
- allowing different 3 inch PMT properties (Hamamatsu vs NNVT, as these are the two candidates we still have for HK)


src/WCSimDetectorMessenger.cc
```diff
@@ -399,7 +419,6 @@ WCSimDetectorMessenger::WCSimDetectorMessenger(WCSimDetectorConstruction* WCSimD
 				    "BoxandLine12inchHQE "
 				    "PMT3inchR12199_02 "
 				    "PMT3inchR14374 "
-            "PMT3inchR14374_WCTE "
```
tdealtry, line 402: Please add these `PMT3inchR14374_WCTE` lines back in - WCTE geometries can't be run without them, as [seen in the CI](https://github.com/WCSim/WCSim/actions/runs/11232171878/job/31223532823?pr=468#step:12:206)

WCSim__hyperk__noID__OD__10k_PMTs__dark_rate_400Hz__cosmic_muons__notrigger__2microsec__10events.mac
tdealtry, line 1: Please `git mv` this into the `macros/` directory

src/WCSimWCSD.cc
```diff
@@ -202,12 +202,12 @@ G4bool WCSimWCSD::ProcessHits(G4Step* aStep, G4TouchableHistory*)
...
     // maxQE = fdet->GetPMTQE(WCIDCollectionName,wavelength,0,200,700,ratio);
-    maxQE = fdet->GetPMTQE(WCCollectionName,wavelength,0,240,660,ratio);
-    photonQE = fdet->GetPMTQE(volumeName, wavelength,1,240,660,ratio);
+    maxQE = fdet->GetPMTQE(WCCollectionName,wavelength,0,200,800,ratio);
+    photonQE = fdet->GetPMTQE(WCCollectionName, wavelength,1,200,800,ratio);
     photonQE = photonQE/maxQE;
   }else if (fdet->GetPMT_QE_Method() == 3){
     ratio = 1./(1.-0.25);
-    photonQE = fdet->GetPMTQE(WCCollectionName, wavelength,1,240,660,ratio);
+    photonQE = fdet->GetPMTQE(WCCollectionName, wavelength,1,200,800,ratio);
```
tdealtry: When speaking to you previously, you mentioned that lowering the wavelengths 240->200 is due to the WLS being sensitive in that region

Is increasing the upper cut 660->800 due to WLS sensitivity too?

src/WCSimDetectorMessenger.cc
tdealtry: bump

src/WCSimDetectorMessenger.cc
```diff
@@ -419,7 +438,6 @@ WCSimDetectorMessenger::WCSimDetectorMessenger(WCSimDetectorConstruction* WCSimD
...
```
tdealtry, line 422: please add back in

src/WCSimDetectorMessenger.cc
```diff
@@ -524,7 +542,6 @@ WCSimDetectorMessenger::WCSimDetectorMessenger(WCSimDetectorConstruction* WCSimD
           "PMT3inchGT\n"
           "PMT3inchR12199_02\n"
           "PMT3inchR14374\n"
-          "PMT3inchR14374_WCTE\n"
```
tdealtry, line 527: please add back in

include/WCSimDetectorMessenger.hh
```diff
@@ -105,6 +105,9 @@ class WCSimDetectorMessenger: public G4UImessenger
   G4UIcmdWithAnInteger* PMTODperCellHorizontal;
   G4UIcmdWithAnInteger* PMTODperCellVertical;
   G4UIcmdWithADouble* PMTODPercentCoverage;
+  G4UIcmdWithADouble* PMTODPercentCoverageTop;
+  G4UIcmdWithADouble* PMTODPercentCoverageBottom;
+  G4UIcmdWithADouble* PMTODPercentCoverageBarrel;
```
tdealtry, line 110: FYI these options won't work with the HyperK_HybridmPMT_WithOD_Realistic, and shouldn't be made to

include/WCSimWCDigi.hh
```diff
@@ -186,6 +189,7 @@ public:
     double index_time,index_timepresmear,index_pe;
     int index_trackID;
     int index_parentSavedTrackID;
+    G4double index_wavelength;
```
tdealtry, line 192: You forgot to use this variable to swap the order of things in the `wavelength` map

As it is now, sorting hits means that the wavelengths are all incorrect

include/WCSimPMTObject.hh
```diff
@@ -461,33 +463,6 @@ protected:
   G4double* GetCollectionEfficiencyArray();
 };
 
-class PMT3inchR14374_WCTE : public WCSimPMTObject
```
tdealtry, line 464: Please bring back the WCTE PMT

src/WCSimPMTObject.cc
tdealtry, line 1: General comment for this file. If you want to implement a new set of constants for a PMT, either
- Add the new implementation as a new PMT class
- Or get the goahead by all users of that PMT type 

e.g. the `PMT3inchR14374` is the default for the FD mPMT too

src/WCSimPrimaryGeneratorMessenger.cc
tdealtry, line 1: Please revert this change

You can keep this change locally so you can test building & running. Just don't `git commit` it

src/WCSimPMTQE.cc
```diff
@@ -40,7 +40,7 @@ G4double WCSimDetectorConstruction::GetPMTQE(G4String CollectionName, G4double P
   // return 0 for wavelength outside the range
   if (flag==1){
     /*ODCOMMENT -- We use photons < 280nm so we need ->*/ //if (PhotonWavelength <= low_wl || PhotonWavelength >= high_wl ){
-    if (PhotonWavelength <= low_wl || PhotonWavelength >= high_wl || PhotonWavelength <=280 || PhotonWavelength >=660){
+    if (PhotonWavelength <= low_wl || PhotonWavelength >= high_wl || PhotonWavelength <=200 || PhotonWavelength >=800){
```
tdealtry: Same comment as elsewhere. Is the change 660->800 required?

src/WCSimEventAction.cc
```diff
@@ -2119,7 +2131,9 @@ void WCSimEventAction::FillRootEventHybrid(G4int event_id,
 				      photonEndPos,
 				      photonStartDir,
 				      photonEndDir,
-              photonCreatorProcess);         // INCLUDE THE VECTOR OF CREATOR PROCESSES
+				      photonCreatorProcess,
+				      wavelength);         // INCLUDE THE VECTOR OF CREATOR PROCESSES
```
tdealtry:
```suggestion
				      wavelength);
```

src/WCSimDetectorConfigs.cc
```diff
@@ -873,16 +887,21 @@ void WCSimDetectorConstruction::SetHyperK_HybridmPMT_WithOD_Geometry()
   WCODCapPMTSpacing  = (pi*WCIDDiameter/(round(WCIDDiameter*sqrt(pi*WCPMTODPercentCoverage)/(10.0*WCPMTODRadius))));
   WCODCapEdgeLimit = WCIDDiameter/2.0 - WCODWLSPlatesLength/2;
 
+  // F. Nova 2022/05/13 remove combine pmt qe
```
tdealtry, line 890: I've always been unclear about the purpose of this combined PMT QE stuff. Can you please explain a little about the purpose of it? And why it's no longer required?

This is still active for the FD realistic geometries, and IWCD+OD geometries. It may be that we want to deactivate it in those places too

src/WCSimConstructRealisticPlacement.cc
tdealtry, line 1: This looks like it's using a single tyvek for all surfaces

I'm 99% sure that barrel & wall tyvek is handled as one in the current setup. But, I'm also 99% sure that it is possible to distinguish between black/white (ID-side) & white/white (concrete side) tyvek of the OD. You may want to implement it (if you don't, please just click resolve)

macros/tuning_parameters.mac
tdealtry, line 1: Please add a comment to say which option(s) are used in the FD realistic geometry case

src/WCSimConstructMaterials.cc
```diff
@@ -1572,8 +1714,10 @@ void WCSimDetectorConstruction::ConstructMaterials()
 
 
   WlsPlasticMPT->AddProperty("RINDEX",WLSProps->GetPhotonEnergy(),WLSProps->GetRIndex(),WLSProps->GetNumEntries());
+#if 1 // turn to 0 to make plate non-reemitting, purely passive
```
tdealtry: Do you really want a purely passive plate as the default?

tdealtry: I'd recommend following the method of the overlap checking, such that you can turn on/off things like this via CMake commands. Check with
```
grep -R WCSIM_CHECK_GEOMETRY_OVERLAPS src/ include/ CMakeLists.txt
```



You can also do the same thing for the default OD PMT type (although why you need to given the ability to set via config file, I don't know...)

Feel free to ignore.

src/WCSimDetectorConstruction.cc
```diff
@@ -524,9 +524,6 @@ WCSimPMTObject *WCSimDetectorConstruction::CreatePMTObject(G4String PMTType, G4S
   else if (PMTType == "PMT5inch"){
     PMT = new PMT5inch;
   }
-  else if (PMTType == "PMT3inchR14374_WCTE"){
```
tdealtry: Please add back in the WCTE PMT

src/WCSimStackingAction.cc
tdealtry, line 1: Sorry, there was a rewrite to make it more easily readable. Can you rewrite your changes to match, so that the diff is clearer?

The obvious change is the range [240,660] -> [200,800]. I can't tell if there's anything else

src/WCSimConstructCylinder.cc
```diff
@@ -276,8 +275,7 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructCylinder()
 					  false,
 					  checkOverlaps);
 
-	//G4LogicalSkinSurface *TyvekCaveBarrelSurface =
-	new G4LogicalSkinSurface("TyvekCaveBarrelSurface", logicCaveTyvek, OpWaterTySurface);
+    G4LogicalSkinSurface *TyvekCaveBarrelSurface = new G4LogicalSkinSurface("TyvekCaveBarrelSurface", logicCaveTyvek, OpWaterTySurfaceOutWallBarrel);
```
tdealtry: General comment for this file: don't declare variables that you don't need to use. We currently have a build that is clear of compilation warnings - I'd like to keep it that way
```suggestion
    new G4LogicalSkinSurface("TyvekCaveBarrelSurface", logicCaveTyvek, OpWaterTySurfaceOutWallBarrel);
```

src/WCSimConstructCylinder.cc
```diff
@@ -288,49 +286,66 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructCylinder()
...
-					  false,
-					  checkOverlaps);
+    G4VPhysicalVolume *physiTopCaveTyvek =
+        new G4PVPlacement(0,
+                          CaveTyvekPosition,
+						  logicCaveCapsTyvekTop,
+                          "CaveTopTyvek",
+						  logicWCBarrel,
+                          false,
+                          0);
```
tdealtry: Please revert. As it was, we can easily turn overlap checks on via compilation options. Without, we can't
```suggestion
                          checkOverlaps);
```

src/WCSimConstructCylinder.cc
```diff
@@ -288,49 +286,66 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructCylinder()
...
-					  false,
-					  checkOverlaps);
+    G4VPhysicalVolume *physiBottomCaveTyvek =
+        new G4PVPlacement(0,
+                          CaveTyvekPosition,
+						  logicCaveCapsTyvekBottom,
+                          "CaveBottomTyvek",
+						  logicWCBarrel,
+                          false,
+                          0);
```
tdealtry:
```suggestion
                          checkOverlaps);
```

src/WCSimConstructCylinder.cc
```diff
@@ -825,19 +840,16 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructCylinder()
...
+          G4LogicalSkinSurface *WaterTyTVSurfaceBot = new G4LogicalSkinSurface("WaterTyTVSurfaceBot", logicWCTVTyvek, OpWaterTySurfaceInWallBottom);
+	  //Top
+	  G4VPhysicalVolume* physiWCTVTyvekTop =
+			new G4PVPlacement(	0,
+		                  		G4ThreeVector(0.,0.,0.5*m
+													+WCTyvekThickness/2),
+								logicWCTVTyvek,
+		               			"WCTVTyvekTop",
+		          				logicWCTopVeto,
+				 				false,0,true);
```
tdealtry:
```suggestion
				 				false,0,checkOverlaps);
```
We also don't want to check for overlaps every time we run - only when we want to

src/WCSimConstructCylinder.cc
```diff
@@ -1488,17 +1507,67 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructCylinder()
...
+		}
+
+        G4ThreeVector topWLSpos = G4ThreeVector(xoffset,
+                                                yoffset,
+                                                ((WCIDHeight + 2*WCODDeadSpace)/2)+WCODTyvekSheetThickness);
+
+        if (((sqrt(xoffset*xoffset + yoffset*yoffset) + WCPMTODRadius) < WCODCapEdgeLimit) ) {
+
+
+		  //		  std::cout << " qqqq cap i " << i << " of " << CapNCellTop << " j " << j << " of " << CapNCellTop << " Container (" << topWLSpos.x() << ", " << topWLSpos.y()
```
tdealtry: I know that it's commented out, but a reminder. Please don't use `std::cout`, `std::cerr`, `std::endl` in WCSim. Use the G4 versions. Otherwise, printout can mangle

(The only exception is in the classes defined in WCSimRoot)

src/WCSimConstructCylinder.cc
```diff
@@ -1488,17 +1507,67 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructCylinder()
...
+		  //					<< ", " << topWLSpos.z() << ") r " << sqrt(pow(topWLSpos.x(),2) + pow(topWLSpos.y(),2)) << " WCODCapPMTSpacing " << WCODCapPMTSpacing << std::endl;
+
+			G4VPhysicalVolume* physiTopCapWLSPlate =
+					new G4PVPlacement(0,                   // its rotation
+									  topWLSpos,
+									  logicWCODWLSAndPMT,   // its logical volume
+									  "WCTopCapContainerOD",// its name
+									  logicWCBarrel,       // its mother volume
+									  false,               // no boolean operations
+									  icopy);
```
tdealtry:
```suggestion
									  icopy, checkOverlapsPMT);
```

tdealtry (changes requested):
Thanks for this @federiconova 

There's been a lot of divergence, so that means a lot of comments. Sorry!

Can I please please please request that you answer these comments from the website, rather than replying to email? It gets **very** mangled when not using the website.
Another benefit of doing it on the website is that you'll be able to implement the suggestions I made with a click of a button (just remember to `git pull` from your branch afterwards!)

The first is a major one. This PR changes a lot of the output. See [here](https://wcsim.github.io/Validation/8599e990df738ed068bbcc0c3c1e31f67c614132/index.html) for the 
- WCTE needs fixing. It currently won't run at all
- There are changes for IWCD, even in the ID-only geometry. Is this expected?
- There are changes for the FD, even in the ID-only geometry. Is this expected? 
- This PR adds some [overlap or material issue](https://wcsim.github.io/Validation/8599e990df738ed068bbcc0c3c1e31f67c614132/2/wcsim_hkfdcomplete_SensitiveDetector_Only_NoTrigger_gun500MeVemODbarrelIn_startX_dark0_n200_bad.diff.txt) to the FD realistic geometry output. This has been clean since it's been added. This is very concerning

I suggest that you implement the comments, push, and we can check the CI output again


Another comment is that the official FD geometry we use is `HyperK_HybridmPMT_WithOD_Realistic`. This is an improvement on the old HK FD geometries in that
- There are significantly fewer (I think zero) overlaps / gaps in the geometry
- The number of PMTs, and their positions, are laid out according to integration diagrams
  - mPMTs are laid out correctly, rather than in stripes
  - This is not perfect, as the number of PMTs can differ by a few. But it is signicant improvement on the previous way

I would suggest that you confirm that your new options affect this geometry correctly. 

Sorry for the long todo list! But thank you very much for the contributions

src/WCSimDetectorMessenger.cc
federiconova: done

federiconova:
> Can I please please please request that you answer these comments from the website, rather than replying to email? It gets **very** mangled when not using the website. Another benefit of doing it on the website is that you'll be able to implement the suggestions I made with a click of a button (just remember to `git pull` from your branch afterwards!)

I guess it's a matter of taste. For me the website is more cumbersome, and also I do not want to change the code online, I want to change it in the terminal and then check that it works! Anyway, I'll reply here.

federiconova:
>     * WCTE needs fixing. It currently won't run at all

I added back the PMT option for WCTE. Is WCTE really using PMT3inchR14374 ? I find it strange because those are the ones we plan to use in the FD OD and I never knew it was going to be the same model (don't the WCTE use flat face 3 inch? the OD ones are hemispherical)
 
>     * There are changes for IWCD, even in the ID-only geometry. Is this expected?

No. What changes?

>     * There are changes for the FD, even in the ID-only geometry. Is this expected?

No. What changes?

federiconova:
> Another comment is that the official FD geometry we use is `HyperK_HybridmPMT_WithOD_Realistic`. This is an improvement on the old HK FD geometries in that
> 
>     * There are significantly fewer (I think zero) overlaps / gaps in the geometry
> 
>     * The number of PMTs, and their positions, are laid out according to integration diagrams
>       
>       * mPMTs are laid out correctly, rather than in stripes
>       * This is not perfect, as the number of PMTs can differ by a few. But it is signicant improvement on the previous way
> 

Sorry, I did not know about this realistic geometry thing. But how do we know it works better? Has it been checked? I have never seen any report about this in the OD group.

WCSim__hyperk__noID__OD__10k_PMTs__dark_rate_400Hz__cosmic_muons__notrigger__2microsec__10events.mac
federiconova: done

src/WCSimWCSD.cc
federiconova: You are right, I put the value back to 660. By the way this would much better be defined as a constant rather than hardcoded like this.

src/WCSimDetectorMessenger.cc
federiconova: done

src/WCSimDetectorMessenger.cc
federiconova: done

include/WCSimDetectorMessenger.hh
federiconova: why not? what is changed in that version? is there no more TYVEK?

include/WCSimWCDigi.hh
federiconova: sorry, fixed now (I think)

include/WCSimPMTObject.hh
federiconova: sorry, done

src/WCSimPMTObject.cc
federiconova: The FD OF will have Hamamatsu R14374-31; is this really the same as the mPMTs are using?

src/WCSimPMTQE.cc
federiconova: done

src/WCSimEventAction.cc
federiconova: done

src/WCSimDetectorConfigs.cc
federiconova: this was coded in by Mahdi Taani (not me). I think his idea was to form a combined QE curve that takes into account both the PMT QE and the WLS plate QE. In my experience it does not work at all.

src/WCSimConstructMaterials.cc
federiconova: the default is 1, so not passive. If you turn it to 0, it becomes passive. This is used to simulate the aged detector (after 20 years of running), under the pessimistic assumption that the plates age badly and turn completely passive.
I had to do it via #if and #endif because the config file does not work - the value you put in the config file is then overwritten again at run time with the default value. It's always bothered me because of course it would be better to control it via mac file and not recompile. But I never could make it work for some of the options (some work, some don't)

src/WCSimDetectorConstruction.cc
federiconova: sorry, done

src/WCSimStackingAction.cc
federiconova: done

src/WCSimConstructCylinder.cc
federiconova: you are right, thanks, I fixed them all

src/WCSimConstructMaterials.cc
kmtsui: wanna do something similar to `WCSIM_SAVE_PHOTON_HISTORY_FLAG`?
https://github.com/WCSim/WCSim/blob/8cc7f87ae1f9f8c96775cd0c2204c80e762fc9dc/CMakeLists.txt#L145

src/WCSimWCSD.cc
tdealtry: Agreed. Opened #485. Feel free to implement if you'd like

src/WCSimDetectorMessenger.cc
```diff
@@ -534,7 +554,7 @@ WCSimDetectorMessenger::WCSimDetectorMessenger(WCSimDetectorConstruction* WCSimD
           "HPD20inchHQE\n"
           "PMT20inch\n");
   SetPMTType->SetParameterName("PMTType", false);
-  SetPMTType->SetCandidates("PMT3inch PMT3inchGT PMT3inchR12199_02 PMT3inchR14374 PMT3inchR14374_WCTE PMT3inch_ETEL9302B PMT5inch PMT8inch PMT10inchHQE PMT10inch PMT12inchHQE HPD20inchHQE PMT20inch");
+  SetPMTType->SetCandidates("PMT3inch PMT3inchGT PMT3inchR12199_02 PMT3inchR14374 PMT3inch_ETEL9302B PMT5inch PMT8inch PMT10inchHQE PMT10inch PMT12inchHQE HPD20inchHQE PMT20inch");
```
tdealtry: Please add WCTE PMT back in

include/WCSimDetectorMessenger.hh
tdealtry: Erm what does OD % coverage have to do with tyvek...?

But what I mean is that the PMTs are not put into cells with blacksheet, tyvek, etc., and cells tessalated together - with extra features (e.g. "extra tower") that is an easy source of bad geometries (overlaps, gaps, etc.)

Instead, the "sheets" are placed as concentric _cylinders_ (no overlaps/gaps possible) with PMTs inserted as "cut outs" of the sheet, based on diagrams from integration
- [Cap PMT placement](https://github.com/WCSim/WCSim/blob/develop/src/WCSimConstructRealisticPlacement.cc#L973)
- [Barrel PMT placement](https://github.com/WCSim/WCSim/blob/develop/src/WCSimConstructRealisticPlacement.cc#L746)

include/WCSimWCDigi.hh
tdealtry: Not quite... a couple more lines required
1. in the loop
2. at the end of the loop

See, for example, the usage of the variables containing the phrase `trackID`, and mirror that

src/WCSimConstructMaterials.cc
tdealtry: Sorry. Let me blame the jetlag...

Yeah there are certainly some things that need to be specified not in the "main" .mac file, but elsewhere e.g. jobOptions or tuning_parameters (these are mainly Geant4 limitations, but it might be that WCSim has more limitations than vanilla G4)

If it is useful for you to more easily recompile (via command line only, rather than having to edit the file to swap `#1` for `#0`) then do take a look at either of the CMake options `WCSIM_SAVE_PHOTON_HISTORY_FLAG` or `WCSIM_CHECK_GEOMETRY_OVERLAPS` (both basically do the same thing, 

Otherwise, we can resolve this conversation now

src/WCSimConstructCylinder.cc
```diff
@@ -1349,21 +1356,30 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructCylinder()
...
+	if( WCPMTODPercentCoverageTop )
+	  WCODCapPMTSpacingTop  = (pi*WCIDDiameter/(round(WCIDDiameter*sqrt(pi*WCPMTODPercentCoverageTop)/(10.0*WCPMTODRadius_for_building))));
+	if( WCPMTODPercentCoverageBottom )
+	  WCODCapPMTSpacingBottom  = (pi*WCIDDiameter/(round(WCIDDiameter*sqrt(pi*WCPMTODPercentCoverageBottom)/(10.0*WCPMTODRadius_for_building))));
 
     if(WCPMTODperCellHorizontal == 0 && WCPMTODperCellVertical == 0){
       ComputeWCODPMT((G4int)NPMTODByCell,WCPMTODperCellHorizontal,WCPMTODperCellVertical);
     }
 
+	std::cout << " NPMTODByCell " << NPMTODByCell << " WCPMTODperCellHorizontal " << WCPMTODperCellHorizontal << " WCPMTODperCellVertical " << WCPMTODperCellVertical << std::endl;
```
tdealtry:
```suggestion
	G4cout << " NPMTODByCell " << NPMTODByCell << " WCPMTODperCellHorizontal " << WCPMTODperCellHorizontal << " WCPMTODperCellVertical " << WCPMTODperCellVertical << G4endl;
```

tdealtry (changes requested):
Thanks for the changes so far @federiconova !!

> I added back the PMT option for WCTE. Is WCTE really using PMT3inchR14374 ? I find it strange because those are the ones we plan to use in the FD OD and I never knew it was going to be the same model (don't the WCTE use flat face 3 inch? the OD ones are hemispherical)

Think there's one last place to fix it I noticed today (the candidate list in the messenger) - there's an inline comment about it

@kmtsui can comment. But all I know is that the vanilla WCSim WCTE geometry uses this PMT type. (I know there's a separate geometry that's closer to "as built" which uses a mix of mPMTs (IWCD-like & FD-like), but I don't know the details of which 3" PMT type is in which module)


> There are changes for IWCD, even in the ID-only geometry

Actually the ID-only stuff has been fixed with your recent batch of commits. But, there are still differences in the [IWCD+OD](https://wcsim.github.io/Validation/0eb7ccccf4af08a9b979caa96e7c4ef011aa1d79/index.html) output, specifically the geometry: [OD barrel PMT positions in IWCD+OD, only at a specific `y` value](https://wcsim.github.io/Validation/0eb7ccccf4af08a9b979caa96e7c4ef011aa1d79/17/geofile_IWCD_mPMT_WithOD.txt.diff.txt). I don't understand why this happens... @kmtsui any thoughts on this?
- I'm not suprised given a geometry change that there is a knock on effect in terms of number of OD PMT hits. But there is [a 50% decrease in true hits](https://wcsim.github.io/Validation/0eb7ccccf4af08a9b979caa96e7c4ef011aa1d79/15/3/index.html), so needs to be understood

> There are changes for the FD, even in the ID-only geometry

- See a load of plots for a load of different particle gun configurations [here](https://wcsim.github.io/Validation/0eb7ccccf4af08a9b979caa96e7c4ef011aa1d79/index.html). But looking e.g. at the ID only with a 500 MeV electron
  - [Number of 20" hits is close to "normal" (within 1%)](https://wcsim.github.io/Validation/0eb7ccccf4af08a9b979caa96e7c4ef011aa1d79/9/1/index.html) so probably fine
  - [Number of mPMT hits is reduced by 20%](https://wcsim.github.io/Validation/0eb7ccccf4af08a9b979caa96e7c4ef011aa1d79/9/2/index.html). This is bad
  - What actually concerns me most is [this](https://wcsim.github.io/Validation/0eb7ccccf4af08a9b979caa96e7c4ef011aa1d79/9/wcsim_hybrid_SensitiveDetector_Only_NoTrigger_gun500MeVem_xdir_dark0_n1000_bad.diff.txt). What we do is grep the output log for common exceptions & this `GeomNav1002` is something that indicates a track has got stuck in the geometry somewhere. The FD realistic geometry has been "clean" since it's been produced. Again, I don't understand what's going on here...

> Sorry, I did not know about this realistic geometry thing. But how do we know it works better? Has it been checked? I have never seen any report about this in the OD group.

Sorry this is an oversight on my part. I did announce it via the software mailing list in January, and discussed it at many software meetings. But we should of course given a talk at a C&P meeting (there was just a one-line in the PSC plenary talk at the Feburary collaboration meeting...)
In terms of what the geometry is/looks like, see for example [@patrickstowell talks on the subject](https://hkdbweb.in2p3.fr/meetingplanner.htm?preloadvw=ds&Owner=pstowell) from last year (ignore the QA one!)  and dissucssion on the PR #399 (@patrickstowell I'm sure there was more discussion on github somewhere, but can't find it)

kmtsui:
> I added back the PMT option for WCTE. Is WCTE really using PMT3inchR14374 ? I find it strange because those are the ones we plan to use in the FD OD and I never knew it was going to be the same model (don't the WCTE use flat face 3 inch? the OD ones are hemispherical)

`PMT3inchR12199_02` is the default option for WCTE in older version. Now we have switched to `PMT3inchR14374_WCTE`, which is based on `PMT3inchR14374` with updates in time and charge response. But IWCD OD is still using `PMT3inchR14374`.

include/WCSimPMTObject.hh
kmtsui, line 1: Maybe just revert this file (except the `protected: G4double* GetCollectionEfficiencyArray();` line)?
Other changes are not real but just switching the names which make the diff confusing

macros/tuning_parameters.mac
kmtsui, line 1: This also change the IWCD OD behavior. Are those numbers well studied, or just randomly assigned?

kmtsui:
> Actually the ID-only stuff has been fixed with your recent batch of commits. But, there are still differences in the [IWCD+OD](https://wcsim.github.io/Validation/0eb7ccccf4af08a9b979caa96e7c4ef011aa1d79/index.html) output, specifically the geometry: [OD barrel PMT positions in IWCD+OD, only at a specific y value](https://wcsim.github.io/Validation/0eb7ccccf4af08a9b979caa96e7c4ef011aa1d79/17/geofile_IWCD_mPMT_WithOD.txt.diff.txt). I don't understand why this happens... @kmtsui any thoughts on this?

The old OD PMT placement code (line 1492 of [src/WCSimConstructCylinder.cc](https://github.com/WCSim/WCSim/pull/468/files#diff-ce91adfc22a9d1b71e87ba74b0b122c1e0dfd211ffae84dc88435b2c82046b06)) places the PMTs in top and bottom endcaps alternately, while the new one finishes top cap first then to bottom cap.
The ordering is changed, but overall layout seems to be the same.

src/WCSimPMTObject.cc
```diff
@@ -3300,47 +3301,95 @@ G4double* PMT3inchR14374::Getqpe()
...
 }
 
-// TF: Used WebPlotDigitizer on Fig.2a from VLVnT13 proceedings, red curve (average)
-G4double* PMT3inchR14374::GetQEWavelength(){
-
-  //TEST: make QE same!!
-  //static G4double wavelength_value[20] = { 280., 300., 320., 340., 360., 380., 400., 420., 440., 460., 480., 500., 520., 540., 560., 580., 600., 620., 640., 660.};
-  static G4double wavelength_value[21] = { 300., 320., 340., 360., 380., 400., 420., 440., 460., 480., 500., 520., 540., 560., 580., 600., 620., 640., 660., 680., 700.};
-  return wavelength_value;
+#if 0 // from ETEL
```
kmtsui: I prefer defining new PMT objects instead of using `#if`, so that we can easily change the PMT type through macro instead of recompiling the program

src/WCSimPrimaryGeneratorMessenger.cc
kmtsui, line 1: why commenting all the hepcm3 lines?

src/WCSimStackingAction.cc
```diff
@@ -40,35 +40,31 @@ G4ClassificationOfNewTrack WCSimStackingAction::ClassifyNewTrack
...
-	    wavelengthQE  = DetConstruct->GetPMTQE(WCIDCollectionName,photonWavelength,0,240,660,ratio);
-	    //wavelengthQE  = DetConstruct->GetPMTQE(volumeName,photonWavelength,0,240,660,ratio);
-	  }else if (DetConstruct->GetPMT_QE_Method()==3 || DetConstruct->GetPMT_QE_Method() == 4){
-	    wavelengthQE = 1.1;
-	  }
-	}//Geometry without OD
-	else {
-	  //Geometry with OD
+	G4double ratio = 1./(1.0-0.25);
+      if(aTrack->GetCreatorProcess()==NULL) {
```
kmtsui: what is this `if` for?

src/WCSimWCSD.cc
kmtsui, line 1: You also need to implement the same changes to `WCSimWCSD::ProcessHits_boundary()`

macros/tuning_parameters.mac
tdealtry: FYI the `tuning_parameters.mac` used in the CI is [this reference file](https://github.com/WCSim/Validation/blob/master/Generate/macReference/tune/tuning_parameters.mac), not the one in the WCSim repo/this PR. So the change of this file does not explain the large changes in the number of hits in the IWCD OD

Of course IWCD can decide to use different values. Since it can be easily changed via `WCSim` executable arguments, we can have `tuning_parameters_IWCD.mac`, `tuning_parameters_HKFD.mac`, ...

include/WCSimDetectorMessenger.hh
federiconova: sorry I was confused. Yes nothing to do with Tyvek.
Well the point of those options is that in the HK FD OD we are still trying to figure out how to optimally put the PMTs. You are right, in the "Realistic" version (by the way, a bad name in my opinion) the PMTs are assigned one by one, so one cannot separately define the coverages.

include/WCSimWCDigi.hh
federiconova: right, fixed now

include/WCSimDetectorMessenger.hh
tdealtry: Ahh sorry. I was under the impression that the PMT positions were finalised now. Since it effects e.g. how detector construction shifts are planned

I suppose what this means then is we need to keep two versions of the FD geometry
- `HyperK_HybridmPMT_WithOD_Realistic` which is used for most studies, and official MC productions
- `HyperK_HybridmPMT_WithOD` which is used only by the OD group for some dedicated studies
  - Do note that there are typically geometry overlaps/gaps with this geometry that can effect results

In both instances, the OD group should provide default options for their parameters


Thinking about it, yeah I agree "realistic" is premature. It is a massive improvement on the old geometry (recall [Menai's mPMT plot fig 1 here](https://wiki.hyperk.org/pub/HyperK/TechNote0075/Tech_Note_HK_Data_Rate.pdf)), but with a long-term view it "based_on_Autumn_2023_integration_diagrams" is more accurate.

Anyway... it's the last _planned_ HKFD geometry in WCSim. So I think we can live with this name

federiconova: Yes I agree with all you say. You may be right that we won't change the PMT positions anymore, although they still ask me to check other options every now and then.
Yes I am aware of the geometry overlap and I am very happy to have a version without overlaps. I think the main point of this PR is to reach an OD simulation that has both the right PMT positions (Patrick work) and the right parameters (my work)

patrickstowell:
Sorry for the late reply, been a busy week.

I agree realistic isn't the greatest name. The idea behind this was to
provide one last geometry update to WCSim that had an up to date number of
mPMTs included as the default values in the old one had not been updated in
a while and had almost an order of magnitude higher number than the latest
estimates.

The code was designed to be as encapsulated as possible so that it could be
ported over to GHOST relatively easily when required, and could easily be
extended in the future to allow individual PMT placement tweaks. This is
the reason we first place arrays of dummy pmts before looping through at
the end and placing the actual PMT logical volumes.

The positions of the OD pmts are based on the pattern given by the
integration group that appeared to give roughly the correct total number of
planned OD pmts suggested at the time.

Since the idea was the geometry was being completely restructured to
prepare for ghost, we did simple tests at the time to verify that the
number of photons in each portion of the geometry looked sensible given the
reduction in photo coverage suggested by the integration group. It wasn't
really possible to cover all possible physics use cases for all groups at
the time so I've been waiting for feedback by users on whether there are
any problematic areas.

Purely coincidentally Liz Kneale noticed a bug just this week where the OD
pmt+plate objects were flipped in space due to the way the logical volumes
are setup for that PMT type, so clearly those initial tests didn't provide
full coverage of use cases. This has been fixed in the recent pull request.

With regards to Tom's suggestion, the OD group could use the old geometry
for studies, but the new "realistic"/"integration" geo uses the materials
and pmts directly from the other WCSim loaders. So it should be able to
automatically pick up new wavelength shifter definitions or place new PMT
geometries by simply replacing the PMT selection strings or upgrading these
to macro commands.

In addition if you need a way to vary the OD coverage, we could do this by
just modifying the nested replica "stampers" in the definition itself. So
you reduce the total coverage by a percentage in each pattern, and then
this new pattern gets duplicated around the entire barrel.
There is already an example doing this using a periodic throw (every 100th
PMT is thrown away for example) as we needed something like this to test
how the drop in number of photons made sense with reduced coverage, so all
you'd need is a more sophisticated way to decide which pmts to keep and
throw away.

Both of these options I'd be happy to add to the new geo if the OD group
could provide some feedback on specific requirements and validation tests
they need between the old and new geometry.

include/WCSimDetectorMessenger.hh
tdealtry: That sounds great! 👍 
Closing this thread now

src/WCSimDetectorConfigs.cc
tdealtry: Thanks. Will test removing it for other geometries. Setup issue to remind me #496

src/WCSimConstructRealisticPlacement.cc
tdealtry: This will be implemented in a future PR

macros/tuning_parameters.mac
tdealtry: This will be implemented in a future PR

src/WCSimConstructMaterials.cc
tdealtry: Closing this, as it is quite a niche thing to do (and the default is sensible). If HK FD or IWCD OD groups think implementing this via a messenger or cmake command, then please implement it

PR #470, opened 2024-08-21
New WCTE geometry

kmtsui:
New feature: 

- introduce `WCBarrelPMTTopOffset` and `WCBarrelPMTBotOffset` to allow asymmetry construction of top and bottom CapAssembly, only for NoReplica detector construction.

WCTE specific geometry change:

- update tank height and radius, and define properly PMT offset values for top and bottom endcaps
- reverse phi orderings of 3-inch PMTs inside ex-situ and in-situ mPMT models
- update mPMT positions and orientations based on WCTE construction database and geometry package

Top view: beam comes from left to right. Side view: beam goes into the paper.
image


tdealtry (approved):
Looks good. Thanks @kmtsui 

Please do check the validation & let me know when you're happy that this PR is good to go (e.g. see 6% fewer digits from 500 MeV electrons [here](https://wcsim.github.io/Validation/3dd1c4d0baf67dff4693925a62d74fe075dd01d6/21/1/index.html))

When you're happy I can merge & issue a new release (but FYI I'm on leave for a few days from tomorrow)

kmtsui:
The drop in hit rate is kind of expected because of the reduction in ID volume and the asymmetric PMT placements. We may do some more checks and come back to merge next week

kmtsui:
@tdealtry We have done some extra checks on PMT coordinates and event generations, and are ready to merge now

PR #476, opened 2024-09-05
Fix WCSIMDIR bug

DiegoCostas97:
When wanting to run a simulation with ambeevt generator, this "WCSIMDIR_BUILD_DIR" was breaking the run.


tdealtry:
Thanks @DiegoCostas97 !

PR #477, opened 2024-09-11
Start event and n_events to simulate for datatable reader

JackFannon:
# Additions
- Start event functionality
  - Set in the macro file using `/mygen/dataTableStartEvent`
    - Set as an integer.
    - Has a default value of 0.
    - If set to a value larger than the number of events in the file then WCSim will exit with an error.
  - Specify the event from which WCSim will start simulating.
  - Indexed beginning at 0.
  - Events are defined using the particle index from the data table.
    - Each event begins with a particle with an index of 0.
    - Particles in the same event have an index counted upwards from 0.
- Specify the number of events for WCSim to simulate.
  - Set in the macro file using `/mygen/dataTableNEvents`
    - Has a default value of -1.
      -  If set to -1 all events in the file will be simulated.
    - If set to larger than the number of events in the file then all of the events will be simulated
    - Counts from the event set by `dataTableStartEvent`
      - e.g. `/mygen/dataTableStartEvent 10` and `/mygen/dataTableNEvents` 5 will simulate events 10, 11, 12, 13, and 14 from the datatable. 
  - Begins at the start event.
  - Simulation stops when reached
# Bugfix
- The check for whether or not the end of the file had been reached in the datatable reader did not work.
  - Fixed this by moving it outside the `while` loop in the reader.
  - Previously WCSim would duplicate the last event simulated until the event number defined by `/rub/beamOn` was reached.
  - Now the run is aborted if the end of the datatable is reached.


tdealtry:
Looks good @JackFannon Thanks!

My only comment is that `/mygen/dataTableNEvents` & `/run/beamOn` are basically the same thing, and I can envisage some user confusion with it. So options
1. See if there's a way to get to get access to the Geant4 event number inside `WCSimPrimaryGeneratorAction`. If there is, use that instead of `dataTableNEvents`
2. Add a comment to the mac file in the data table bit, suggesting to use a massive `beamOn` number, and control number of events via `dataTableNEvents`

1 is the preference, but of course GHOST is coming. So 2 is also fine

JackFannon:
Thanks for the feedback @tdealtry ! You make a very good point. I'd not considered trying to access the `beamOn` value. I'll look into seeing whether I can grab it in the generator.

PR #482, opened 2024-10-14
Debug trigger algo

kmtsui:
Debug modulus operation for `window_start_time==0`, which causes `window_start_time` to move back by one `window_step_size`.


tdealtry:
Thanks @kmtsui

Issue #484, opened 2024-10-22
Endcap asymmetry


ekneale:
There is something causing asymmetry in the endcaps. When generating photons with WCSimLIGen, I found asymmetry when plotting charge maps (diffuser open angle 40 degrees):

![Screenshot from 2024-10-18 17-02-24](https://github.com/user-attachments/assets/a00f3795-2096-42c2-9c11-9feec0bd453a)
Top endcap, central diffuser at (0,0,+3305.82)

![Screenshot from 2024-10-18 17-08-26](https://github.com/user-attachments/assets/5c7856f0-d725-4706-984d-66acb75443d7)
Bottom endcap, central diffuser at (0,0,-3305.82)

![Screenshot from 2024-10-17 10-12-20](https://github.com/user-attachments/assets/d5293741-f486-4a74-aa71-a62490a29e4e)
Bottom endcap, centre edge diffuser

I have checked amongst other things:

1. Using hard-coded diffuser angles in WCSimLIGen in case there was an error reading in the angles.
2. Generated photon angles - these are in line with the diffuser angles e.g. injector angles of theta = 0, phi = 0, with a direction (0,0,-1) sets all photon directions to (0,0,-1) as expected. 
3. Deactivating WLS plates

Photon visualisation looks fine:

![Screenshot from 2024-10-17 12-59-29](https://github.com/user-attachments/assets/51cb9229-9cbf-44e2-89fe-d198c87fa9c2)

If I set the injector open angle to zero and point it straight down, I get the same asymmetry. However if I use the old geometry it looks pretty symmetrical:

![Screenshot from 2024-10-22 17-15-43](https://github.com/user-attachments/assets/8f5417c7-e8b1-45e7-a6ce-0cb37789bbc8)

I can't find any obvious overlaps in the new geometry but will continue to investigate.

tdealtry:
Thanks for opening the issue @ekneale. I know we've spoken a bit offline, but a few more possibilities to check

- is the same G4-defined material used for top & bottom? In terms of tyvek, the way the code is setup (applying it to the surface of a cylinder) this is unlikely, but @patrickstowell should confirm that my understanding is correct
- are all rotations applied correctly? e.g. are the top PMTs/WLS plates "upside down"? I think the easiest way to check is to visual it (@patrickstowell @kmtsui know how to visualise geometry in WCSim. I've never got it working personally)
[...]

PR #486, opened 2024-10-24
Fixes bug in OD PMT orientation

ekneale:
Fixes bug in OD PMT orientation so that they point outwards rather than inwards. Resolves issue #484 .

ID PMTs (locations 0-2) still point inwards, while OD PMTs (locations 3-5) now point outwards as per output file:

[geofile_HyperK_HybridmPMT_WithOD_Realistic.txt](https://github.com/user-attachments/files/17510359/geofile_HyperK_HybridmPMT_WithOD_Realistic.txt)


tdealtry:
Thanks @ekneale !

Unfortunately this might fix the OD, but has broken the ID
- Number of 20" hits for a 5 MeV particle gun @ detector centre [has dropped dramatically](https://wcsim.github.io/Validation/b98c602b6da501793eb2573783cce5414b007d2f/11/1/index.html)
- This is probably related to the [geometry difference](https://wcsim.github.io/Validation/b98c602b6da501793eb2573783cce5414b007d2f/11/geofile_HyperK_HybridmPMT_IDonly_Realistic.txt.diff.txt)
  - Interestingly the only PMTs with a difference are 20" barrel PMTs. No changes in OD PMTs...

For ease, here's the first few of the diff
```
<   13248 13248   0  3241.222790189   106.106647744 -3250.036000000 -0.999464587 -0.032719083  0.000000000   1
<   13247 13247   0  3242.766171259    35.374494696 -3179.352000000 -0.999940505 -0.010908091  0.000000000   1
<   13246 13246   0  3242.766171259   -35.374494696 -3250.036000000 -0.999940505  0.010908091  0.000000000   1

>   13248 13248   0  3223.390120916   105.522866597 -3250.036000000  0.999464587  0.032719083  0.000000000   1
>   13247 13247   0  3224.925010560    35.179870103 -3179.352000000  0.999940505  0.010908091  0.000000000   1
>   13246 13246   0  3224.925010560   -35.179870103 -3250.036000000  0.999940505 -0.010908091  0.000000000   1
```
So the barrel PMTs have reversed (x & y) orientation. Otherwise, things seem consistent

I think this leads to this, which comes from the [20" PMT position (for a single PMT)](https://github.com/WCSim/WCSim/blob/develop/src/WCSimConstructGeometryTables.cc#L232)
```
< Detector radius & height  3242.96 6592.94
---
> Detector radius & height  3225.12 6592.94
```

src/WCSimConstructRealisticPlacement.cc
```diff
@@ -660,9 +660,11 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructRealisticPlacement()
     multipmt_central_rotation->rotateZ(270*deg);
 
     // For the OD PMT we instead need to make it face outwards from the white tyvek in the OD.
+    // Rotating with respect to ID PMT orientation
     G4ThreeVector odpmt_central_position = G4ThreeVector(config.WhiteTyvekOuterRadius+2*cm,0.0,0.0);
-    G4RotationMatrix* pmtod_central_rotation = new G4RotationMatrix;
-    pmtod_central_rotation->rotateY(270*deg);
+    G4RotationMatrix* pmtod_central_rotation = pmt_central_rotation;
```
tdealtry: Oh wait... it's probably this
This is basically assigning an alias to the ID barrel rotation matrix, then rotating it - which will effect both ID & OD
```suggestion
    G4RotationMatrix* pmtod_central_rotation = new G4RotationMatrix(*pmt_central_rotation);
```

tdealtry:
Hmmm better. But still not perfect
See [here for a muon starting within the OD barrel region, pointing straight up](https://wcsim.github.io/Validation/190a115f917b9b2f08e4a418f47298644bbfadf5/5/index.html)
- There is 1 event with 20" hits
- There is 1 event with mPMT hits (presumably the same event)
- The tank height is changed according to the geofile. I don't understand this one...

So this is bad. It suggests light leaking from OD to ID. I checked the CI job log output (which uses `-DWCSim_DEBUG_COMPILE_FLAG=ON`) and can't see anything obvious in the PR. Do you have any ideas what it could be?

A couple of random suggestions
- Outside chance of the muon scattering into the ID. You could take a look at the ROOT output & check the track/hit information in detail (we use one random number generator throughout WCSim/Geant4, so some change anywhere can cause a change in the physics that's simulated)
- Could be the size of the OD PMT being used. Since the default OD tube is 8", it may be that it's too big & causing some issue, if the "dummy" PMT is 3". (It's maybe not caused an issue before, given it was rotated so the "big bit" wasn't on the detector side & causing an issue)

ekneale (commented):
Agh yes that would do it.

ekneale:
The dummy OD PMT is 8". I wonder if an overlap has been introduced with the change in detector height. I'll look into it.

I get this for the 3" PMT: Detector radius & height  3242.96 6696.96
and this for the 8" PMT: Detector radius & height  3242.96 6697.19

according to the geofile.

ekneale:
@tdealtry I'm still looking into this. I can confirm that I get hits in the inner detector if I generate muons pointing straight up but I don't get any hits in the inner detector if I simulate optical photons from a diffuser pointing either straight up or towards the ID.

ekneale:
Tried a few things (using 200 x 500 MeV muons pointing straight up in the OD with fixed seed to test):

**1.** Increased the size of the dummy PMT to exactly correspond to the 8" PMT radius (currently the default PMT for the OD). No change. For the 3" PMT, I get 4 events with hits in the ID. For the 8" PMTs, I get 2 events with hits in the ID.
**2.** Changed WCBlackSheetThickness to 1 mm, which is currently set to 2 cm by default. Simulated with 3" PMTs, which changed number of events with ID hits but still getting them. Simulated with 8" PMTs, and didn't get any ID hits.
**3.** Ran with checkOverlaps and checkOverlapsPMTs true. Found overlaps - looks like an issue with mPMTs. Some examples as follows:

**When placing mPMTS:**

-------- WWWW ------- G4Exception-START -------- WWWW -------
*** G4Exception : GeomVol1002
      issued by : G4PVPlacement::CheckOverlaps()
Overlap with volume already placed !
          Overlap is detected for volume WCMultiPMT
          with av_17_impr_2_pmt20_pv_3028 volume's
          local point (137.256,32.9686,137.306), overlapping by at least: 5.70757 cm
NOTE: Reached maximum fixed number -1- of overlaps reports for this volume !
*** This is just a warning message. ***
-------- WWWW -------- G4Exception-END --------- WWWW -------

-------- WWWW ------- G4Exception-START -------- WWWW -------
*** G4Exception : GeomVol1002
      issued by : G4PVPlacement::CheckOverlaps()
Overlap with volume already placed !
          Overlap is detected for volume WCMultiPMT
          with av_17_impr_2_pmt20_pv_386 volume's
          local point (15.9709,133.875,126.835), overlapping by at least: 6.88931 cm
NOTE: Reached maximum fixed number -1- of overlaps reports for this volume !
*** This is just a warning message. ***
-------- WWWW -------- G4Exception-END --------- WWWW -------

-------- WWWW ------- G4Exception-START -------- WWWW -------
*** G4Exception : GeomVol1002
      issued by : G4PVPlacement::CheckOverlaps()
Overlap with volume already placed !
          Overlap is detected for volume WCMultiPMT
          with WCMultiPMT volume's
          local point (148.762,-200.3,170.996), overlapping by at least: 4.49969 mm
NOTE: Reached maximum fixed number -1- of overlaps reports for this volume !
*** This is just a warning message. ***
-------- WWWW -------- G4Exception-END --------- WWWW -------

**When placing 20" PMTs:**

-------- WWWW ------- G4Exception-START -------- WWWW -------
*** G4Exception : GeomVol1002
      issued by : G4PVPlacement::CheckOverlaps()
Overlap with volume already placed !
          Overlap is detected for volume WCPMT
          with WCMultiPMT volume's
          local point (148.762,202.262,166.839), overlapping by at least: 2.92288 mm
NOTE: Reached maximum fixed number -1- of overlaps reports for this volume !
*** This is just a warning message. ***
-------- WWWW -------- G4Exception-END --------- WWWW -------


-------- WWWW ------- G4Exception-START -------- WWWW -------
*** G4Exception : GeomVol1002
      issued by : G4PVPlacement::CheckOverlaps()
Overlap with volume already placed !
          Overlap is detected for volume WCPMT
          with WCMultiPMT volume's
          local point (49.4151,-210.893,122.029), overlapping by at least: 3.73951 cm
NOTE: Reached maximum fixed number -1- of overlaps reports for this volume !
*** This is just a warning message. ***
-------- WWWW -------- G4Exception-END --------- WWWW -------

tdealtry:
Hi @ekneale Just following up with some progress I made on this today.
I used a muon particle gun file (500 MeV, aiming down the barrel) with 1 out of 100 events having hits in the ID
Doing some digging in truth, I see that a Bremsstrahlung gamma is created & scatters through the tyvek & blacksheet into the ID
I'm currently trying to setup some CI to see if something like this is going on in this PR. I'm hopeful that we can get this merged before the holidays

tdealtry:
Hi @ekneale So the [CI has run](https://wcsim.github.io/Validation/90fdb63a333b41a90c8845320ca096fa800b2f6d/index.html). This is what I see
- Number of OD hits go up dramatically. This is expected, as the sensitive detector is now exposed, rather than hidden in the ID/OD deadspace
- Number of overlaps in the geometry has dropped from 3577 to 45, consisteny with the numbers I found (the remaining are from 20"/mPMT overlaps) & with the number of OD PMTs
- For the "OD barrel up" particle guns, we see 1 or 2 events where there is some true particle track that isn't an electron or Cherenkov photon in the ID. And it's only these events where we're seeing ID hits. This is the same as for my small test file where I see Bremstrahlung happening, causing a gamma to scatter into the ID & produce a bunch of hits. So I'm happy with this
- For the "ID start 500 MeV" particle guns, we see a difference in the number of hits in the ID (both mPMT & 20"). Other than the fact that OD PMT overlaps have been fixed, I don't understand why things should change in the ID... But the change is <1%, so is livable
- The thing that concerns me are the "OD barrel in" particle guns, where we see a large change in the number of 20" & mPMT hits by ~20%. Weirdly, this is a decrease for "startX" and an increase for "startY". Things are now more symmetric, which seems more correct too
- There's also a more "Optical photon is killed because of missing refractive index" warnings in the new code for the "OD barrel" particle guns (both up & down), which could be related to some of the hit differences...

I think I want to do a bit more work understanding the events before I merge. Particularly the "OD barrel in" stuff. A plausible explanation is that the muons are getting through the OD & dead region with a "more correct" energy distribution now (going through fewer overlap regions that Geant4 struggles with).

This'll be next year now. Enjoy the holidays!

ekneale:
Hi @tdealtry thanks for looking into this! I'm concerned about the ~13% difference between the startX and startY 'OD barrel in' particle guns. Granted this is a smaller difference than previously but I don't understand the root cause of such a significant asymmetry. Perhaps it would be helpful to look at equivalent OD start positions in terms of the PMT and mPMT positions to reduce the effect of PMT/mPMT asymmetry around the start points.

tdealtry:
Hi @ekneale yes I agree it looks odd, much larger energy loss than naively I would expect for particles travelling through a PMT/not. We do still have a few (45) overlaps - it may also be that these are causing problems...
I think this is a good suggestion. I was planning on producing a load of muons & electron particle guns, with different starting positions around the OD to try & understand this. But I don't have any time to set this running until after next week (teaching...). If you happen to have any time to look into this before then, I'd really appreciate it. But no worries if not

patmanmx1991:
After much pain and suffering trying to get develop built again on OSX I have the vis working, and it is now clear why this still doesn't work. The rotation applied only brings it 50% in line, which means a huge overlap between the OD PMTs and the DeadSpace (DS) volume between OD and ID as they are left perp. to the barrel.

In the final "Corrected" fix in the slides below the orientation is sorted. I'll submit a new pull request tomorrow with this fixed (and the vis tools added to avoid this sort of problem in the future...). Sorry that this bug snuck in!

[HK_ODPMT_Fix.pdf](https://github.com/user-attachments/files/18633666/HK_ODPMT_Fix.pdf)

tdealtry:
Awesome thanks @patmanmx1991!

tdealtry:
Thanks @ekneale 

The CI looks good... mostly.
- Number of overlaps has reduced
- See a lot more OD hits. e.g. mean digits goes up from 95->230 in OD barrel up Y
- We still see 1/200 events in barrel up Y with hits in the ID. But this appears to be something like a compton scatter (but in barrel up X we go from 1 to 0 of these - it's just a MC fluctuation)

So that's all great. What I want to understand a bit more (tomorrow) before merging is
- Number of ID hits for the barrel in CI is reduced. I think we discussed this before - need to check whether the particles are being produced directly facing (or inside) a PMT, or not
- The PMT position file is different, even for ID PMTs. I want to check whether this is some kind of reordering, or precision issue, or if it's real.

src/WCSimConstructRealisticPlacement.cc
```diff
@@ -654,8 +654,8 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructRealisticPlacement()
     // Since assembly volumes just expand relative to their center point the means
     // all nested volumes will also have the PMT againt the tyvek.
 
-    G4ThreeVector pmt_central_position = G4ThreeVector(config.InnerDetectorOuterRadius-59.62*mm,0.0,0.0);
-    G4ThreeVector mpmt_central_position = G4ThreeVector(config.InnerDetectorOuterRadius-59.62*mm,0.0,0.0);
+    G4ThreeVector pmt_central_position = G4ThreeVector(config.InnerDetectorOuterRadius-2*mm,0.0,0.0);
+    G4ThreeVector mpmt_central_position = G4ThreeVector(config.InnerDetectorOuterRadius-2*mm,0.0,0.0);
```
tdealtry:
```suggestion
    G4ThreeVector pmt_central_position = G4ThreeVector(config.InnerDetectorOuterRadius-59.62*mm,0.0,0.0);
    G4ThreeVector mpmt_central_position = G4ThreeVector(config.InnerDetectorOuterRadius-59.62*mm,0.0,0.0);
```
I'd just like to run the CI with this change reverted. There hasn't been any discussion as to why it was required. And I'd like to, if possible, not alter the ID as this will require fitqun tuning

ekneale: Hi Tom, the change to the ID PMT offset was to bring them flush to the wall.

patmanmx1991:
Sorry this debugging snuck in.

The expose height before was based on the original geometry which shifts
them out slightly based on pmt module height.
It is very difficult to follow exactly why it is needed in the original
because the position calcs pass through a few stages.

This also needs to be looked in to now that we actually have vis working
properly again.

You are right that we should change one at a time, so we should remove
this, and then raise a seperate pull request for the other shift after a
chat in a software meeting.

tdealtry:
Thanks both! Yes, I agree separating the ID & OD differences is sensible (it is worrying that the 20" PMTs are poking out into the ID a bit too much. It'd be good to see some displays & have a chat in a software meeting - maybe the premeeting? I know short notice now)

In terms of the "ID PMTs back to _normal_" CI run, it seems much of a muchness
- ID PMT positions are back to normal. Good!
- Fewer (but not 0) geometry overlaps still
- minor ID hit changes for ID-starting particles iff they're energetic enough to get into the OD (makes sense as this will cause the random number stream to alter for subsequent events)
- large increase in OD hits ✨ 
- significant change in ID hits for "OD barrel in" CI. Still need to understand this. Tomorrow I'll prioritise taking a closer look

I also checked the OD PMT positions - they seem to be correct (it seems mostly to be the cyloc being swapped - I also need to check which is "correct" according to the expectation in the docs - another fix might be required in the making of the geometry tables)

tdealtry:
I'm going to merge this now. Here's some checks I've done, to try to convince myself that the (inconsistent) drop in ID hits for the 500 MeV electron starting in the OD barrel, aiming @ detector centre (i.e. should go through deadspace into the ID) are reasonable

## Tests

### Code
Comparing this code, with the develop branch as of February (i.e. the same reference as the last CI run)

### Setup
There are 8 configurations (1-4 X A-B)

1. `StartX` @ `34, 0 0 m`, direction `-1 0 0`
2. `StartY` @ `0 34 0 m`, direction `0 -1 0`
3. `StartXOUTSIDEPMTS` @ `33.42 6.2532871356 0 m`, direction `-0.98294118 -0.18392021 -0.`
4. `StartYOUTSIDEPMTS` @ `6.96 33.28 0 m`, direction `-0.20470588 -0.97882353 -0.`

1 & 2 are the default CI parameters.
3 & 4 was calculated by studying the positions of all PMT types, and looking for a starting position (close to 1&2) that is the furthest away from the PMT position. The idea being, this is most likely to produce a primary track that will _just_ pass through tyvek & blacksheet on the way to the ID (i.e. more likely to avoid PMT & WLS plate materials)

A. 500 MeV electon
B. 500 MeV muon

A is the default CI particle.
B is a particle that is less likely to deviate from the initial trajectory before it reaches the ID

Per 4 starting positions, there are
- 6000 electron events
- 1000 muon events

### Checks
1. Plot a few "Number of 20" PMT digitised hits per event" histograms on the same axes
2. Get mean & stddev (+ their errors) with ROOT `TH1` methods. Put these on the plot in the legend
3. Calculate unbinned K-S test, to calculate (use scipy `ks_2samp`). Put these in a table

The idea is that, for like distributions (e.g. using the same code, `StartXOUTSIDEPMTS` should be compatible with `StartYOUTSIDEPMTS`, but not necessarily `StartX` or `StartY`)
- means & stddeviations are consistent within 1 sigma ~68% of the time, and most others are consistent within 2 sigma
- K-S test returns a compatible probability of >5% 

This is done comparing
- different starting positions, for each code & particle type
- different code, for each particle type

## Results

### 500 MeV electrons w/ old code
compare_ref_e

| | 500 MeV $e^-$ startXOUTSIDEPMTS | 500 MeV $e^-$ startX | 500 MeV $e^-$ startYOUTSIDEPMTS | 500 MeV $e^-$ startY |
|--- | --- | --- | --- | --- |
| 500 MeV $e^-$ startXOUTSIDEPMTS | 100% | | | |
| 500 MeV $e^-$ startX | 89% | 100% | | |
| 500 MeV $e^-$ startYOUTSIDEPMTS | 55% | 12% | 100% | |
| 500 MeV $e^-$ startY | 8% | 9% | 2% | 100% |

- `OUTSIDEPMTS` X & Y have consistent means (within 1 sigma)
- `OUTSIDEPMTS` X & Y standard deviations are consistent only at >2 sigma. Not so worried about this, due to the non-Gaussian distribution
- Most distributions consistent with pvalue > 5%
- `startY` inconsistent (or borderline consistent) with other distributions. Suggests that this is a different setup to the other 3 (lower mean suggests particle tracks are likely to pass through PMT/WLS plate material)

### 500 MeV muons w/ old code
compare_ref_mu

| | 500 MeV $\mu^-$ startXOUTSIDEPMTS | 500 MeV $\mu^-$ startX | 500 MeV $\mu^-$ startYOUTSIDEPMTS | 500 MeV $\mu^-$ startY |
|--- | --- | --- | --- | --- |
| 500 MeV $\mu^-$ startXOUTSIDEPMTS | 100% | | | |
| 500 MeV $\mu^-$ startX | 6% | 100% | | |
| 500 MeV $\mu^-$ startYOUTSIDEPMTS | 57% | 57% | 100% | |
| 500 MeV $\mu^-$ startY | 50% | 31% | 98% | 100% |

- `OUTSIDEPMTS` X & Y have almost consistent means (just over 1 sigma)
- `OUTSIDEPMTS` X & Y standard deviations are consistent only at almost 2 sigma
- All distributions consistent with pvalue > 5%. `StartX` vs `StartY` is borderline

### 500 MeV electrons w/ bug fix code
compare_fix_e

| | 500 MeV $e^-$ startXOUTSIDEPMTS | 500 MeV $e^-$ startX | 500 MeV $e^-$ startYOUTSIDEPMTS | 500 MeV $e^-$ startY |
|--- | --- | --- | --- | --- |
| 500 MeV $e^-$ startXOUTSIDEPMTS | 100% | | | |
| 500 MeV $e^-$ startX | 17% | 100% | | |
| 500 MeV $e^-$ startYOUTSIDEPMTS | 85% | 4% | 100% | |
| 500 MeV $e^-$ startY | 73% | 9% | 21% | 100% |

- `OUTSIDEPMTS` X & Y have consistent means (within 1 sigma)
- `OUTSIDEPMTS` X & Y standard deviations are consistent (just over 1 sigma)
- Most distributions consistent with pvalue > 5%. 
- `startYOUTSIDEPMTS` vs `startX` is inconsistent. `startX` is the one that is consistently less compatible with others, also having a lower mean, again suggesting that there is a systematic difference. It should be noted that in the old code `startY` was the discrepency

### 500 MeV muons w/ bug fix code
compare_fix_mu

| | 500 MeV $\mu^-$ startXOUTSIDEPMTS | 500 MeV $\mu^-$ startX | 500 MeV $\mu^-$ startYOUTSIDEPMTS | 500 MeV $\mu^-$ startY |
|--- | --- | --- | --- | --- |
| 500 MeV $\mu^-$ startXOUTSIDEPMTS | 100% | | | |
| 500 MeV $\mu^-$ startX | 36% | 100% | | |
| 500 MeV $\mu^-$ startYOUTSIDEPMTS | 9% | 50% | 100% | |
| 500 MeV $\mu^-$ startY | 87% | 37% | 18% | 100% |

- `OUTSIDEPMTS` X & Y have consistent means (within 1 sigma)
- `OUTSIDEPMTS` X & Y standard deviations are consistent (just over 1 sigma)
- All distributions consistent with pvalue > 5%

### 500 MeV electrons `OUTSIDEPMTS` comparing new/bug-fix code
compare_code_e

| | ref 500 MeV $e^-$ startXOUTSIDEPMTS | ref 500 MeV $e^-$ startYOUTSIDEPMTS | bugfix 500 MeV $e^-$ startXOUTSIDEPMTS | bugfix 500 MeV $e^-$ startYOUTSIDEPMTS |
|--- | --- | --- | --- | --- |
| ref 500 MeV $e^-$ startXOUTSIDEPMTS | 100% | | | |
| ref 500 MeV $e^-$ startYOUTSIDEPMTS | 55% | 100% | | |
| bugfix 500 MeV $e^-$ startXOUTSIDEPMTS | 85% | 35% | 100% | |
| bugfix 500 MeV $e^-$ startYOUTSIDEPMTS | 77% | 54% | 85% | 100% |

- Bugfix code is more consistent with itself (means/std deviations within 1 sigma, better p-value)  than the old code

### 500 MeV muons `OUTSIDEPMTS` comparing new/bug-fix code
compare_code_mu

| | ref 500 MeV $\mu^-$ startXOUTSIDEPMTS | ref 500 MeV $\mu^-$ startYOUTSIDEPMTS | bugfix 500 MeV $\mu^-$ startXOUTSIDEPMTS | bugfix 500 MeV $\mu^-$ startYOUTSIDEPMTS |
|--- | --- | --- | --- | --- |
| ref 500 MeV $\mu^-$ startXOUTSIDEPMTS | 100% | | | |
| ref 500 MeV $\mu^-$ startYOUTSIDEPMTS | 57% | 100% | | |
| bugfix 500 MeV $\mu^-$ startXOUTSIDEPMTS | 85% | 46% | 100% | |
| bugfix 500 MeV $\mu^-$ startYOUTSIDEPMTS | 9% | 34% | 9% | 100% |

- Bugfix code is less consistent with itself (means consistent only @ ~3 sigma)  than the old code

## Summary
- Using the new `OUTSIDEPMTS` setup, there is, in general, more consistency between starting at `+x` or `+y` (both old & new code)
- 500 MeV electron distributions look much better using bugfix code
- 500 MeV muon distributions are an anomaly

tdealtry:
Another test was performed by @kmtsui, in response to worries from Miura-san that the bug fix code changes the relative position between the PMT volume & the tyvek, referring to the diagrams from @patrickstowell [here](https://github.com/user-attachments/files/18633666/HK_ODPMT_Fix.pdf)

> Page 1 is before fixing OD bug and the OD surface runs middle of OD PMT edge
in the barrel (left figure). Page 3 is after fix and OD surface runs on
OD PMT edge. I feel OD PMT position relative to OD surface is slightly different
before/after bug fix. I guess the white thick line in OD PMT edge is showing
WLS plate, but I'm wondering white-black tyveK sheet locates behind the WLS plate.
If there are no tyvek sheet behind the plate, light leak to ID region may happen.
>
> The difference between +x and +y vertex case may indicate OD modules locates
away from +x but close to +y if above assumption is correct.

Looking at the event displays on p.5 and p.6 of your slides, I do think the bug fixed version is more correct, as the red box (OD PMT logical volume) is not overlapping with the OD Tyvek visually. 
Since the old version has the OD flipped, the OD volume overlaps with Tyvek volume, given the placement position is exactly on the Tyvek surface. 
After the bug fix, it should be just touching the Tyvek surface. With two surfaces supposingly touching, it's very hard to see any possible overlapping with just visualization. 
But I guess overlap check is run before the merge, so if no overlapping error appears I would say we are safe. 

As an extra check, one can do a geantino particle gun, shooting towards the OD PMT from outside to the inner Tyvek, then look at tracking verbose to confirm tracking volume.

For example, I shoot a geantino from bottom upwards to pass the OD PMT:
*********************************************************************************************************
* G4Track Information:   Particle = geantino,   Track ID = 1,   Parent ID = 0
*********************************************************************************************************

Step#    X(mm)    Y(mm)    Z(mm) KinE(MeV)  dE(MeV) StepLeng TrackLeng  NextVolume ProcName
    0 3.11e+04 -6.35e+03 -3.37e+04   2.5e-06        0        0         0 OuterDetector initStep
    1 3.11e+04 -6.35e+03 -3.36e+04   2.5e-06        0      112       112 HyperK_HybridmPMT_WithOD_Realistic-glassFaceWCPMT_OD Transportation
    2 3.11e+04 -6.35e+03 -3.36e+04   2.5e-06        0      5.5       117 InteriorWCPMT Transportation
    3 3.11e+04 -6.35e+03 -3.35e+04   2.5e-06        0     86.1       203  WhiteTyvek Transportation
    4 3.11e+04 -6.35e+03 -3.35e+04   2.5e-06        0        1       204   DeadSpace Transportation
    5 3.11e+04 -6.35e+03 -3.29e+04   2.5e-06        0      600       804  BlackTyvek Transportation
    6 3.11e+04 -6.35e+03 -3.29e+04   2.5e-06        0       20       824 InnerDetector Transportation
    7 3.11e+04 -6.35e+03 3.29e+04   2.5e-06        0 6.58e+04  6.66e+04  BlackTyvek Transportation
    8 3.11e+04 -6.35e+03 3.29e+04   2.5e-06        0       20  6.66e+04   DeadSpace Transportation
    9 3.11e+04 -6.35e+03 3.35e+04   2.5e-06        0      600  6.72e+04  WhiteTyvek Transportation
   10 3.11e+04 -6.35e+03 3.35e+04   2.5e-06        0        1  6.72e+04 InteriorWCPMT Transportation
   11 3.11e+04 -6.35e+03 3.36e+04   2.5e-06        0     86.1  6.73e+04 HyperK_HybridmPMT_WithOD_Realistic-glassFaceWCPMT_OD Transportation
   12 3.11e+04 -6.35e+03 3.36e+04   2.5e-06        0      5.5  6.73e+04 OuterDetector Transportation
   13 3.11e+04 -6.35e+03 3.55e+04   2.5e-06        0 1.91e+03  6.92e+04   WallTyvek Transportation
   14 3.11e+04 -6.35e+03 3.55e+04   2.5e-06        0        1  6.92e+04    WCBarrel Transportation
   15 3.11e+04 -6.35e+03 3.55e+04   2.5e-06        0      2.5  6.92e+04       WCBox Transportation
   16 3.11e+04 -6.35e+03 3.56e+04   2.5e-06        0      150  6.94e+04     expHall Transportation
   17 3.11e+04 -6.35e+03 1.05e+05   2.5e-06        0 6.94e+04  1.39e+05  OutOfWorld Transportation

This geantino passes through the OD PMT into ID. OD PMT volume is touching the Tyvek surface. The track length in WhiteTyevk is 1 mm, which is consistent with the code definition [here](https://eur02.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FWCSim%2FWCSim%2Fblob%2Fd1a19a76f2e66bb7915dfff2ee767689f4688e3a%2Fsrc%2FWCSimConstructRealisticPlacement.cc%23L342&data=05%7C02%7Cdealtry%40live.lancs.ac.uk%7C4ad6e8a3446e451b70a108dd939e3559%7C9c9bcd11977a4e9ca9a0bc734090164a%7C0%7C0%7C638829030183740898%7CUnknown%7CTWFpbGZsb3d8eyJFbXB0eU1hcGkiOnRydWUsIlYiOiIwLjAuMDAwMCIsIlAiOiJXaW4zMiIsIkFOIjoiTWFpbCIsIldUIjoyfQ%3D%3D%7C0%7C%7C%7C&sdata=9Vh9aheeV5JP9IgNffV95EoDcFgXd59JzVl8WzxLmyg%3D&reserved=0).

Then there is another geantino slightly off which goes through the WLS plate instead.
*********************************************************************************************************
* G4Track Information:   Particle = geantino,   Track ID = 1,   Parent ID = 0
*********************************************************************************************************

Step#    X(mm)    Y(mm)    Z(mm) KinE(MeV)  dE(MeV) StepLeng TrackLeng  NextVolume ProcName
    0 3.11e+04 -6.5e+03 -3.37e+04   2.5e-06        0        0         0 OuterDetector initStep
    1 3.11e+04 -6.5e+03 -3.36e+04   2.5e-06        0      112       112 WCBorderCellODContainer Transportation
    2 3.11e+04 -6.5e+03 -3.35e+04   2.5e-06        0     80.6       192 WCCellWLSPlateOD Transportation
    3 3.11e+04 -6.5e+03 -3.35e+04   2.5e-06        0       10       202 WCBorderCellODContainer Transportation
    4 3.11e+04 -6.5e+03 -3.35e+04   2.5e-06        0        1       203  WhiteTyvek Transportation
    5 3.11e+04 -6.5e+03 -3.35e+04   2.5e-06        0        1       204   DeadSpace Transportation
    6 3.11e+04 -6.5e+03 -3.29e+04   2.5e-06        0      600       804  BlackTyvek Transportation
    7 3.11e+04 -6.5e+03 -3.29e+04   2.5e-06        0       20       824 InnerDetector Transportation
    8 3.11e+04 -6.5e+03 3.29e+04   2.5e-06        0 6.58e+04  6.66e+04  BlackTyvek Transportation
    9 3.11e+04 -6.5e+03 3.29e+04   2.5e-06        0       20  6.66e+04   DeadSpace Transportation
   10 3.11e+04 -6.5e+03 3.35e+04   2.5e-06        0      600  6.72e+04  WhiteTyvek Transportation
   11 3.11e+04 -6.5e+03 3.35e+04   2.5e-06        0        1  6.72e+04 WCBorderCellODContainer Transportation
   12 3.11e+04 -6.5e+03 3.35e+04   2.5e-06        0        1  6.72e+04 WCCellWLSPlateOD Transportation
   13 3.11e+04 -6.5e+03 3.35e+04   2.5e-06        0       10  6.72e+04 WCBorderCellODContainer Transportation
   14 3.11e+04 -6.5e+03 3.36e+04   2.5e-06        0     80.6  6.73e+04 OuterDetector Transportation
   15 3.11e+04 -6.5e+03 3.55e+04   2.5e-06        0 1.91e+03  6.92e+04   WallTyvek Transportation
   16 3.11e+04 -6.5e+03 3.55e+04   2.5e-06        0        1  6.92e+04    WCBarrel Transportation
   17 3.11e+04 -6.5e+03 3.55e+04   2.5e-06        0      2.5  6.92e+04       WCBox Transportation
   18 3.11e+04 -6.5e+03 3.56e+04   2.5e-06        0      150  6.94e+04     expHall Transportation
   19 3.11e+04 -6.5e+03 1.05e+05   2.5e-06        0 6.94e+04  1.39e+05  OutOfWorld Transportation

And we see there is actually a 1 mm gap between the WLS plate and the Tyvek surface.

With these printouts, I personally think the overlapping is solved. But it wouldn't hurt to have more tests.


TomD comment: this looks reassuring that there are no hidden overlaps

tdealtry:
And follow up from @patrickstowell who is also happy that things are working as expected

> Yes I've checked in detail zooming in on the visualiser and there is a 1mm gap as Ka Ming says which is to be expected. 
Before the bug fix there was a rotation and spatial problem which put them too far out (a result of us trying to fix things blind when the visualiser broke).
>
> The geometry roughly goes : plate -> 1mm gap -> white tyvek -> OD Dead space -> black tyvek -> ID.
>
> We put some tolerance of 1mm given it would be impossible to get this to zero in construction.

tdealtry:
Given the visulisations/tests that we've performed, this is obviously a massive improvement to before the bug fix. So
- I'm going to merge this in now
- A release will be released shortly after. This will be the basis for the HK MC production
- I'll present the plots/tables I made at the next software meeting, to gather peoples' opinions on whether there is anything that needs investigating further

Thank you @patrickstowell @kmtsui @ekneale for all your hard work with this!

patrickstowell:
Here is a screenshot

image

PR #489, opened 2024-10-29
Avoid warning for first event (ie everytime WCSim is run)

gondiaz:


tdealtry:
Thanks @gondiaz

Erwanlbv:
Dear Tom, Gonzalo,

I updated my wcsim to 12.18 and now I have the following display when I simulate 999 particules

> ***********************************************************
> Created WCSimRootEvent 1000. There is a memory leak when doing TTree::GetEntry() when reading files. It is also presumed that filling is leaky. It is strongly recommended not to generate or analyse O(1000) events in a single job
> ***********************************************************
> 

Then just after  (which I explain even less)

> ***********************************************************
> Created WCSimRootEvent 2000. There is a memory leak when doing TTree::GetEntry() when reading files. It is also presumed that filling is leaky. It is strongly recommended not to generate or analyse O(1000) events in a single job
> ***********************************************************

From
/run/beamOn 999

nickwp:
Sorry to hijack this discussion, but I'm adding an extra comment here, because it's related to this "memory leak":
Are we sure that there is genuinely an unavoidable memory leak when processing WCSimRoot files using `TTree::GetEntry()` in the current WCSim version?
I haven't tested this out in recent versions yet (but plan to do so soon) but in previous versions, while these memory leaks have plagued this kind of processing, I always eventually managed to avoid them by finding the exact specific conditions that allowed ROOT to avoid any leak. Anything ther than the right condtions would either have a memory leak, or segfaults, but in the end I was able to process files with tens/hundreds of thousands of events without memory increasing significantly during the process. The techniques that helped were to:
- Using `SetAutoDelete(kTRUE)` on the `wcsimT` `TTree` and/or the `TBranch`(es) containing any `WCSimRootEvent` objects
- Ensuring that `WCSimRootEvent*` and/or `WCSimRootTrigger*` objects are _initialised_ (to 0 or `nullptr`) _before_ any calls to `SetBranchAddress()` (for `WCSimRootEvent*`s) and _before_ starting the loop over events (for `WCSimRootTrigger*`s). Specifically, I found that `WCSimRootTrigger*` pointers declared within the event loop could be problematic. I honestly don't understand why this mattered, but it seemed to make a difference
- Being careful about whether/where to explicitly `delete` or call `->Delete()` on these pointers anywhere in the event loop, since doing so caused segfaults or other errors, I guess due to interfering with ROOT's own processes enabled by `SetAutoDelete(kTRUE)`
- Maybe other things I'm forgetting, but hopefully can figure out from various old WCSim processing/analysis code

I'm not sure exactly the right conditions to get everything to work without memory building up, as it changed for different WCSim versions (and maybe even ROOT versions), but I really hope I can manage to find a way as I'm currently developing a tool for processing WCSimRoot files for ML. In some cases it really makes sense for us to have much larger numbers of events per file for huge ML datasets where we want to avoid having hundreds of thousands of smaller files. Hopefully I can figure it out and share a working recipe.

But if you have any hints to what has already been tried and failed to work, or if you know for sure there is a memory leak that cannot be worked around with tricks like above (since the warning of this PR was presumably added because it seemed to be unavoidable) please let me know so I can avoid wasting too much time on any futile attempts!

tdealtry:
Hi @nickwp. @tjones79 & I spent quite a lot of time a few months back trying the some of these (and Tom delved into valgrind). We tried a lot of combinations, without success. If you have some time to look into this, then good luck!! And thank you!!! If you manage to find the right conditions, that's be really great. But currently `sample_readfile.C` doesn't even work without leaking...

I'm unsure whether it is a ROOT6 thing, or a WCSim bug that's been introduced. I'd recommend starting with a version of WCSim you happen to remember worked without leaking, or go back until you hopefully find a WCSim version that doesn't leak (I did this & as far back as I went it still leaked. I just can't remember how far back I delved... but circa post-"the nuprism merge")

On the warning appearing more often than you suspect @Erwanlbv. I suspect that it's because you have multiple PMT types in your simulation. Each one saves results in its own WCSimRootEvent class instance. So 999 events = 2997 WCSimRootEvents (I think even if you have just one PMT type in your geometry, they're still created as empty - as your output file should show)

nickwp:
A quick thought about the memory leak: I'm fairly certain there must be a way to avoid it, unless it's a bug in ROOT itself, because when generating events with WCSim it's constructing and destructing exactly the same WCSimRootXyz objects that are constructed and destructed when reading and processing a WCSim output file. But WCSim itself does so without any memory leak, so we should be able to when reading the files too. I'll keep you updated on whether or not I figure it out.

Erwanlbv:
Hi @tdealtry, @nickwp 

Thank you for your clarifications. It is of great help. 
To be sure I understood correctly, the fact that I have this 2000 WCSimRoo0Event call comes from

- I have PMT and mPMTs in my geometry ( 1 + 1 ) 
- I have OD PMT (even though it is called one of the ID Only geometry) (1)

Hence there should be no problem of memory leak, it's just that, as you said 999 * 3 = 2997.

Would it be possible to add some identifier on the pmt type and count only the WCSimRootEvent calls for the PMTs (by default ?)

tdealtry:
Hi @Erwanlbv I don't think it's particularly easy to do actually, at least not the way it's currently setup (it's a combined message to warn for both reading & writing). I'm also skeptical as to whether it's urgent. The message is just a warning for the user to have a think about what they're doing - it's not saying that the user has to only run a limited number of events (or even killing the program if the user exceeds a limit). It is of course added printout, but this happening every 333 events is small relative to the amount of printout that is seen in each event.

The message is accurate (it counts WCSimRootEvent instances, instead of events). But as this is a subtlety that most users don't understand (and is rather confusing given the use of "event" inside WCSimRootEvent) we could modify the message to be something like
```
Created WCSimRootEvent 1000 (when creating WCSim files, 3 are created each event). There is a memory leak when doing TTree::GetEntry() when reading files. It is also presumed that filling is leaky. It is strongly recommended not to generate or analyse O(1000) events in a single job
```
What do you think?

gondiaz:
Hi,
I have never tried to figure out the reasons of the memory leak, but let me share my particular experience because I actually needed to proccess a large amount of events to create the time pdfs for fitqun.

To process the simulation files to produce the pdfs, I use the following script, that I made: 
https://github.com/gondiaz/WCSimFQTuningTools/blob/main/Time/fiTQun/create_2D_histogram.cc

As you can see, WCSim data is loaded through fitqun's WCSim wrapper, and if I am not wrong (just a fast check) this fullfils @nickwp conditions to avoid the leakage (except WCSimRootTrigger initialization to null maybe). Running this script as it is sufered from leakage even though the number of events per file was reasonable, for sure less than 1000 but I dont remember the exact number, but not more than 100.

Since I use Python to look into this data and develop Pyfitqun, I overcame the leakage by developping [WCSimDataTools](https://github.com/gondiaz/WCSimDataTools/) which converts the root files into hdf using the same data structure as in the original root files. This is not only solving the leakage problem but also removes the need of PyROOT and WCSimRoot library in the data reading and processing.

Erwanlbv:
Hello all,

@tdealtry Thank you for your answer. I agree on the fact that it is not urgent, as long as the warning message is easily understood.
I like your update of the warningmessage. Maybe we can add even more details and say, as you told me, 

`[...]1 pmt type = 1 WCSimRootEvent. Hence if you have ID PMTs + mPMTs + OD PMTs then one run (from /run/beamOn n) = 3 WCSimRootEvent calls. The total number of call to WCSimRootEvent being 3 * n after the simulation`.

What do you think ? 
As you said, it is displayed only every 333 run, so having a "big" warning message does not seem problematic to me.

@gondiaz This is a great input. I am wondering, is your 
https://github.com/gondiaz/WCSimFQTuningTools/blob/main/Time/fiTQun/create_2D_histogram.cc
based on **AnalyzeWCSimHierachy.cc** we can find in WCSim ? If not, what are the differences between the extracted data of both programs ?

Also now **to process WCSim's root output files** we have 

- [DataTools](https://github.com/WatChMaL/DataTools) from the [Watchmal](https://github.com/WatChMaL)  collaboration, which does (**wcsim root-> .npy -> .hdf5** and **rely on pyroot**)
- [WCSimDataTools](https://github.com/gondiaz/WCSimDataTools/) you developed (**wcsim root -> hdf5** directly no numpy ?) 
I saw some "import ROOT" from your .py sources, do you confirm you don't rely on PyRoot ? i. e. we don't need to install any pyroot ro run the entire code ?
- [CAVERNS/WCSimRootToRoot](https://github.com/CAVERNS-Org/WCSimRootToRoot) from CAVERNS, (**wcsim root -> flat root**) that can be combined with [CAVERNS/RootToGraph](https://github.com/CAVERNS-Org/RootToGraph) which only uses uproot for **flat root -> torch tensors**.

It is getting out of the context of this PR, but maybe we could switch to a slack channel or smthg to discuss about what does what ? I know people from Watchmal want to update **DataTools**, maybe this is a good time to look at what @gondiaz just introduced us.

Last update for the memory leakage : see #503 (Thx @nickwp  !)

gondiaz:
> @gondiaz This is a great input. I am wondering, is your https://github.com/gondiaz/WCSimFQTuningTools/blob/main/Time/fiTQun/create_2D_histogram.cc based on **AnalyzeWCSimHierachy.cc** we can find in WCSim ? If not, what are the differences between the extracted data of both programs ?

**create_2D_histogram.cc was made from scratch** and is not based on any other script.

>     * [WCSimDataTools](https://github.com/gondiaz/WCSimDataTools/) you developed (**wcsim root -> hdf5** directly no numpy ?)
>       I saw some "import ROOT" from your .py sources, do you confirm you don't rely on PyRoot ? i. e. we don't need to install any pyroot ro run the entire code ?

**It is a direct wcsim root -> hdf conversion**. As I (implicitly) said it needs PyROOT dependency since you need to read WCSim ROOT files, thus loading the corresponding library. The point is that once this conversion is done, all the data is hdf saved and you no longer need ROOT to work with the data.

PR #491, opened 2024-11-20
WCTE beam pipe and camera housing construction

kmtsui:
This PR constructs the beam pipe and photogrammetry camera housing structures used in WCTE. Plus a minor bug fix in `src/G4MultiUnion_v1072.cc`

---
## Beam pipe
A stainless steel structure with blacksheet reflectivity, except the beam window which has reflector (90%) reflectivity. Inside is a hollow volume filled with air.
image

## Camera housing
Since the structure is wrapped with black Tyvek, it is simplified as an upper spherical acrylic shell and lower stainless steel cylinder.
image

## Placement
Integrated as part of the PMT placement loop. Use the PMT table input to control positions and directions. The structures are intruding the blacksheet volume in the same way as PMT
image


tdealtry:
Thanks @kmtsui Happy to merge & release when you're happy

Just to mention that the number of PMTs has changed too - is this expected? See e.g. https://wcsim.github.io/Validation/8778442fdc72eebe4f7feca76d5e0a321c728447/23/geofile_NuPRISMBeamTest_16cShort_mPMT.txt.diff.txt

kmtsui:
> Just to mention that the number of PMTs has changed too - is this expected? See e.g. https://wcsim.github.io/Validation/8778442fdc72eebe4f7feca76d5e0a321c728447/23/geofile_NuPRISMBeamTest_16cShort_mPMT.txt.diff.txt

Yes one of the mPMT is replaced by the beam pipe, which is also shown in the event display figure.

PR #499, opened 2025-01-13
Updates to LI code to integrate ID

menailj:
Update structure so more adaptable for ID and OD profiles and LI positioning. 
Changed access of theta value from water profile to correspond to HK geometry.
Updated where profile is deleted so that the code doesn't crash upon completion.


macros/LIGen.mac
```diff
@@ -107,23 +107,26 @@
...
 # Set /mygen/photonMode 1 to read pre-generated photon positions and directions 
 # from the 'photonsfile' specified in the injector file.
-#/mygen/photonMode 1
+# /mygen/photonMode 1
+
 
 ## change the name of the output root file, default = wcsim.root
 
-/WCSimIO/RootFile wcsim_LIGen.root
+/WCSimIO/RootFile wcsim_LIGen_new.root
```
tdealtry, line 125:
```suggestion
/WCSimIO/RootFile wcsim_LIGen.root
```
It's up to you as calibration group representive (so feel free to ignore this), but for future users the word "new" means nothing here

src/WCSimLIGen.cc
```diff
@@ -46,49 +46,76 @@ void WCSimLIGen::SetPhotonMode(G4bool photonmode){
...
+    
     if (db.empty()) {
         db = wcsimdir + "LightInjectors.json";
     }
     else {
         db = wcsimdir + injectorFilename;
     }
+
+    if (db_pos.empty()) {
+        db_pos = wcsimdir + "LightInjectorsLocation.json";
```
tdealtry:
```suggestion
        db_pos = wcsimdir + "LightInjectorsDetails.json";
```
I think this is the correct default?

include/WCSimLIGen.hh
```diff
@@ -33,18 +33,19 @@ class WCSimLIGen
         G4double GetPhotonEnergy();
         // Set whether to read in profile or photon list
         void SetPhotonMode(G4bool photonMode);
-
-    private:
-
-        G4ParticleGun*                  myLIGun;
-
+        TH2D *hProfile = nullptr;
```
tdealtry: I can't see why this was made `public`, as I don't think it's been used in that way. Can you change it back to `private`?

src/WCSimLIGen.cc
```diff
@@ -22,8 +22,8 @@ WCSimLIGen::WCSimLIGen(){
 WCSimLIGen::~WCSimLIGen(){
 
     // things to delete
-    delete myLIGun;
-    if(hProfile) delete hProfile;
+  //delete myLIGun;
+  //if(hProfile) delete hProfile;
```
tdealtry: The answer is probably "doing this because of ROOT/Geant4 memory management complications", but just to check that you've tried running with just one delete statement and still see a seg fault at the end?

tdealtry (changes requested):
Looks good thanks @menailj Just a few minor suggestions (sorry for the delay on them!)

One other thing to think about is whether to add documentation of the file format in the `doc/` folder. e.g. this is [a nice description of a different input file format](https://github.com/WCSim/WCSim/blob/develop/doc/CathodeParameterFormat.md). But the benefit of this is that this format is JSON, and so this makes it (mostly) very clear via the examples already present

tdealtry (approved):
Thanks Menai

PR #503, opened 2025-02-12
Change TObjArray of triggers to TClonesArray to fix memory leak

nickwp:
After many hours trying to track down the issue, easiest fix for the memory leak was to replace `WCSimRootEvent`'s `TObjArray` of `WCSimRootTrigger`s into a `TClonesArray`, and update the related management of the objects to use more standard ROOT way.

I don't fully understand why the `TObjArray` caused the memory leak. It's due to the `TObjArray` not "owning" the objects it points to (while `TClonesArray` does) and so it doesn't automatically delete them when the `TObjArray` is deleted (`TClonesArray` clears those objects automatically and recursively). But there was code to manually delete them in the `TObjArray` destructor. I think maybe this failed because it was deleting the entries during the loop over entries, which may not behave as expected, but I didn't figure out a way to fix that. Using `SetAutoDelete` seemed to cause all but about 50kb per event of memory to be cleaned up, but I wasn't able to track down exactly where that remaining memory came from.

In my tests of this version, the memory leak in `GetEntry` is completely gone, but further verification of this would be nice if someone has time. Using `SetAutoDelete` is still required (or manually deleting and resetting the `WCSimRootEvent*` to `nullptr`, or I think just calling its `Reinitialize` method, in each iteration of the event loop).

In any case, `TObjArray` is supposed to hold objects of different class types, whereas `TClonesArray` holds objects of the same class type, so is more appropriate anyway and should perform better. It actually does not delete the members when emptying the array, but just clears them, saving some overhead of calling contructors and destructors.

I don't think the underlying container is exposed by any `WCSimRootEvent` methods, so hopefully this change does not break anything downstream.

### DIFF
--- a/include/WCSimRootEvent.hh
+++ b/include/WCSimRootEvent.hh
@@ -656,7 +656,6 @@ public:
   void Initialize();
 
   void ReInitialize();
-  static int fNumberOfWCSimRootEventCreated; //!< Count of the number of calls to the WCSimRootEvent constructor & ReInitalise()
 
 private:
   //std::vector<WCSimRootTrigger*> fEventList;
--- a/include/WCSimRootEvent.hh
+++ b/include/WCSimRootEvent.hh
@@ -641,12 +641,12 @@ public:
     ++Current; 
 
     //Initialize() creates array with 10 entries.
-    // Using AddAtAndExpand() doubles the array size whenever
+    // TClonesArray expands the array size whenever
     // we hit the current array size.
     // This ensures we can keep adding entries,
     // even in the case of huge events
     // e.g. if we have a close SN
-    fEventList->AddAtAndExpand(new WCSimRootTrigger(num,Current),Current);
+    new((*fEventList)[Current]) WCSimRootTrigger(num,Current);
   }
   
   /*  void ReInitialize() { // need to remove all subevents at the end, or they just get added anyway...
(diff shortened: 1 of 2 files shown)


nickwp:
By the way, there is a simpler way to resolve the memory leak, keeping it as a `TObjArray`: simply adding `fEventList->SetOwner()` after creating the `TObjArray* fEventList`, and removing the loop that deletes the contents manually in the `WCSimRootEvent` destructor. But I think it's better to use this PR's approach of changing it to a `TClonesArray` (which is more appropriate and should perform better).

include/WCSimRootEvent.hh
```diff
@@ -660,7 +660,7 @@ public:
 
 private:
   //std::vector<WCSimRootTrigger*> fEventList;
-  TObjArray* fEventList; //!< Array of WCSimRootTrigger
+  TClonesArray* fEventList; //!< Array of WCSimRootTrigger
   Int_t Current;                      //!               means transient, not writable to file
 
   ClassDef(WCSimRootEvent,4)
```
tdealtry:
```suggestion
  ClassDef(WCSimRootEvent,5)
```
We best do this, in order to help ROOT out when reading files created with the different version of WCSim (new/old & old/new combinations)

tdealtry (approved):
Amazing. Thanks @nickwp ✨ 

I'm just going to push some minor things

tdealtry:
Being bad and merging despite all checks not finished running yet...
But the first few checks are all fine. And the commits should break everything or nothing

PR #509, opened 2025-02-13
Allow the multiDigitsPerTrigger=false option to work for the NoTrigger trigger

### LINKED ISSUE #505: MultiDigitsPerTrigger option not respected for NoTrigger mode.
Haven't had time to properly investigate this, but it appears that the daq option
```
/DAQ/MultiDigitsPerTrigger false
```
is not respected when using the NoTrigger mode. There can still be multiple hits on a given PMT.
At least in a test I ran with this option enabled, and dark rate window wide enough to get multiple per PMT, there appears to be multiple hits per PMT in the single trigger of each event.
It does print "Will restrict number of digits per PMT per trigger to <= 1" at the start of the run, but I guess without any trigger this doesn't actually happen.

It would be useful to be able to have a NoTrigger mode without multiple hits per trigger (meaning per event in this mode) on a given PMT.

tdealtry:
Closes #505 @nickwp please confirm

### DIFF
--- a/macros/daq.mac
+++ b/macros/daq.mac
@@ -14,7 +14,9 @@
 
 #generic trigger options (defaults are class-specfic. Can be overrideen here)
 # allow the number of digits per PMT per trigger to be > 1?
-#/DAQ/MultiDigitsPerTrigger false
+#/DAQ/MultiDigitsPerTrigger false #default for NDigits trigger
+#/DAQ/MultiDigitsPerTrigger true #default for NoTrigger trigger
+
 
 #trigger algorithm development options (defaults are set in WCSimWCDAQMessenger. Can be overidden here)
 # do we want to save only triggered events (mode 0), both triggered events & failed events (mode 1), or only failed events (mode 2)?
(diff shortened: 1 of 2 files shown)


nickwp:
Quick question: is multiDigitsPerTrigger enabled by default and only disabled if explicitly set to false?
I ask because you mentioned that it's currently expected behaviour that with NoTrigger it allows multiple hits to be stored as is, to be processed downstream. With this change, it allows disabling that as I requested, but if the default is for multiDigitsPerTrigger to be set to false then this might unexpectedly break other people's current procedures.
So I just wanted to check that the new behaviour only gets enabled if someone explicitly sets multiDigitsPerTrigger to false.

tdealtry:
Good point! Looking in the code, it has a different default for each trigger type (NoTrigger vs NDigits) so the default behaviour if `MultiDigitsPerTrigger` isn't set is unchanged (it's `true`)

This made me notice that there was a bug in my original fix. It should work now (tested via looking at the WCSimRootOptions output for all combinations of trigger (NoTrigger/NDigits) & `/DAQ/MultiDigitsPerTrigger = true/false/commented out` 

If you have some MC you can easily test on to check it's working as expected, that'd be great

nickwp (approved):
Checked some MC that had multiple hits per PMT per trigger before and now they don't. Thanks for this!

PR #512, opened 2025-02-23
Working build with optional hepmc3

patrickstowell:
HEPMC3 is only needed in limited cases. Added a "-DWCSIM_HEPMC3_ENABLED=ON" option for the build files, so a system install of hepmc3 is not always needed. 

An explanation of usage can be seen using 'make' GUI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,9 @@ set(CMAKE_CXX_STANDARD 14) # Enable c++14 standard
 ########
 # HepMC3
 ########
+option(WCSIM_HEPMC3_ENABLED "HEPMC3 INTERFACE" OFF)
+if (WCSIM_HEPMC3_ENABLED)
+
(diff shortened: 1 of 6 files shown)


tdealtry:
Thanks @patrickstowell Looks good! And will certainly make my life easier. Thanks

2 minor requests
- Add precompiler guard to WCSimPrimaryGeneratorMessenger.cc to prevent users trying to run the hepmc generator & getting empty events. I'd suggest expanding [this line](https://github.com/WCSim/WCSim/blob/develop/src/WCSimPrimaryGeneratorMessenger.cc#L21) across multiple lines, in a similar way to [this](https://github.com/WCSim/WCSim/blob/develop/src/WCSimDetectorMessenger.cc#L59) (Note that the trailing spaces in the quotes are important)
- Add documentation of this to README.md [here](https://github.com/WCSim/WCSim/blob/develop/README.md?plain=1#L131)

patrickstowell:
Done

tdealtry:
Thanks Patrick

PR #513, opened 2025-02-23
Feature/geonavigator scan

patrickstowell:
Adds geometry scanning tools to check for overlaps/misalignment based on G4Navigator itself. Avoids vis issues with checking geo as this is as close as possible to what logical volumes optical photons see when traversing the geometry.

Must be called after the geometry is closed in a macro (run/initialize). 

GeometryScan functions use the G4Navigator to determine the logical volume at given xyz points, scanning a point cloud to a map for debugging.

**Example Macro**
```
/WCSim/ScanGeometryBoundaryCenter -3185 0 0 cm
/WCSim/ScanGeometryBoundaryWidth 100 100 100 cm
/WCSim/ScanGeometryStepSize 1 1 1 cm
/WCSim/ScanGeometryToFile test_pointscan_barrel_edge.txt
```

Example Plot
image


src/WCSimDetectorMessenger.cc
```diff
@@ -1132,13 +1168,28 @@ void WCSimDetectorMessenger::SetNewValue(G4UIcommand* command,G4String newValue)
 	  WCSimDetector->SetPMTPositionInput(newValue);
 	}
 
-  if(command == SetODPMTPositionInput){
-	  WCSimDetector->SetODPMTPositionInput(newValue);
+        if(command == SetCDSFile){
+	  WCSimDetector->SetCDSFile(newValue);
 	}
```
tdealtry, line 1177: Not sure why `SetODPMTPositionInput` has been removed

src/WCSimConstructRealisticPlacement.cc
```diff
@@ -317,6 +317,7 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructRealisticPlacement()
   // and light tight volumes. PMTs are then placed in the corresponding
   // volumes using G4Assembly classes.
 
+    bool checkOverlapsPMTRealistic = false;
```
tdealtry: Please add this as a data member [here](https://github.com/WCSim/WCSim/blob/develop/include/WCSimDetectorConstruction.hh#L503), and set based on the cmake option. We're using this overlap checking in the CI, and so it is useful to have a way to turn it on/off

If it really is useful to have the realistic PMT overlap checking set differently to the other overlap checking, then please add as a new cmake option (just a minor bit of copy paste)

Or actually, it may make most sense to add a new CMake option e.g. `WCSIM_CHECK_PMT_GEOMETRY_OVERLAPS`, and use `checkOverlapsPMT` everywhere (no new `checkOverlapsPMTRealistic` variable)

WCSim.mac
```diff
@@ -265,6 +265,14 @@
 #/control/execute macros/visRayTracer.mac
 #/control/execute macros/visOGLQt.mac
 
+##################
+# SCANNING TOOLSS
+##################
```
tdealtry:
```suggestion
##################
# GEOMETRY SCANNING TOOLS
# Use to visualise the geometry by requesting the geometry material at regular test points in the detector, in order to check for misalignment, overlaps, gaps, etc.
#
# See macro sample-root-scripts/ for a way to visualise the result
# (Macro not yet put in the WCSim codebase. In the meantime, please ask Patrick Stowell)
#
# These MUST be given after the geometry is fully setup
##################
```

tdealtry (changes requested):
This looks really nice. Thanks @patrickstowell 

Can you also add to e.g. `sample-root-scripts/` the macro you used to make the plot from the text file?

src/WCSimConstructRealisticPlacement.cc
tdealtry: Just to note that this change is the reason of the [CI failure](https://wcsim.github.io/Validation/e6621975a5ff17d5da42805226d4d5646a47fb5f/24/index.html).
Always good when there's a clear explanation for a change!

patrickstowell:
Ah does the CI turn on overlap checking and grep the stdout?

tdealtry:
Exactly. Not perfectly informative, but something easy & (fairly) robust that will flash up any obvious bad changes to geometries

tdealtry (approved):
Thanks @patrickstowell

PR #515, opened 2025-02-23
Easier vis macro loading

patrickstowell:
Current WCSim setup means passing any extra arguments drops it into Batch mode. For Vis we need to load tuning_parameters and a macro to construct the geometry from the command line, before then opening the vis window with the G4 UI manager. Otherwise the default geometry gets drawn (slow).

Future apps need to have a "-v" flag to make sure the UI gets created, but for now this just checks if three arguments have been given. If so, it asssumes that the last one is vis. Not the nicest solution, but getting vis working again is quite important...

`WCSim WCsim.mac macros/tuning_parameters.mac macros/vis_OGLQT.mac`


tdealtry (approved):
Looks good. Thanks for bringing back this functionality

PR #517, opened 2025-03-12
Fix boundary types for HK FD realistic geometry

### LINKED ISSUE #516: Wrong boundary types for HK FD realistic geometry
Shuoyu Chen found that when using the realistic geometry, the [boundary types](https://github.com/WCSim/WCSim/blob/develop/include/WCSimRootEvent.hh#L62) are always returns 3 [`kTyvek`](https://github.com/WCSim/WCSim/blob/develop/include/WCSimEnumerations.hh#L38)

This variable is filled based on the name of volumes. e.g. for the ID blacksheet, it expects [BlackSheet](https://github.com/WCSim/WCSim/blob/926d2a5183bf09ddd5831d4e49b5ce82d427b210/src/WCSimTrajectory.cc#L262)

This seems to be because the realistic geometry is using different names for some volumes e.g. for the ID blacksheet it is called [BlackTyvek](https://github.com/WCSim/WCSim/blob/develop/src/WCSimConstructRealisticPlacement.cc#L532)

I see 2 options
- change the realistic geometry names in WCSimConstructRealisticPlacement.cc
  - BlackTyvek -> BlackSheet
  - WallTyvek -> CaveWallTyvek
- expand WCSimTrajecory.cc to except "BlackTyvek" as returning `kBlackSheet`

I think the first is best, as it'll prevent any other similar volume-related bugs (now and in future). Any objections? @patrickstowell @nickwp 
If not, I'll put in a PR in a couple of days, and ask Shuoyu Chen to test

As an aside, I don't like the fact that `kTyvek` is meant to mean "inner tyvek wall of OD", since the "outer wall of the OD" is also covered in tyvek... We could be more agressive and rename the enumerations too (e.g. `kInnerTyvek` and `kOuterTyvek`). Any opinions on this? Or better naming suggestions?

tdealtry:
Closes #516 

@ShuoyuChen-physics please test that this fixes your issue!

### DIFF
--- a/include/WCSimEnumerations.hh
+++ b/include/WCSimEnumerations.hh
@@ -35,8 +35,8 @@ typedef enum EBoundaryType {
   kNull=0,
   kBlackSheet,
   kMPMT,
-  kTyvek,
-  kCave
+  kInnerTyvek,
+  kOuterTyvek
 } BoundaryType_t;
 
 typedef enum EReflectionSurface {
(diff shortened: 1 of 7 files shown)


cshuoyu:
Now it works well, thank you for solving my issue!

PR #501, opened 2025-01-28
Update IWCD OD PMT positions

kmtsui:
Major changes

- Decouple OD volume from ID cylinder and caps in NoReplica construction
- Implement PMT position table input for OD PMTs
- Add default ID and OD PMT positions for IWCD nominal geometry

Figure below shows the updated IWCD OD constructions. Yellow lines show the Tyvek surfaces, red dots+cyan rectangles are the OD PMTs and WLS plate. The barrel OD PMT arrangement is referenced from p.13 [here](https://docs.google.com/presentation/d/1yb16r-RWFwRn3eDf1RM1gjgNIUk5cOH--uyQL28VAhc/edit?usp=sharing), with the horizontal spacing between left and right columns be 1.16m. Cap placement is probably overfilled, masking studies will be done later.
 
image

PR #518, opened 2025-03-14
Update IWCD ID mPMTs and OD layout

kmtsui:
Major changes

1. Use in-situ mPMT instead of default one
2. Increase barrel mPMT vertical separation from 67.4 to 71 cm
3. Change cap volume from tube to Polyhedra, to avoid overlap with OD Tyvek
4. Increase vertical OD water thickness from 70 to 78 cm, place the truncated bottom frame structure into OD
5. Remove OD PMT along the bottom frame xy beams.

For now, the bottom frame file (53MB) is not committed, @pdeperio do you want it be in the WCSim repo or somewhere else? Any suggestion from @tdealtry ?

Geometry displays shown below. Right is the bottom view showing the bottom frame.
@pdeperio do you want larger vertical separation of barrel mPMT? And should I remove more OD cap PMTs? In PR https://github.com/WCSim/WCSim/pull/501, you showed a design with much less OD cap PMTs (with 90 deg rotational symmetry?).
image


kmtsui:
Since the OD cap volume is changed from a tube to Polyhedra in the NoReplica method, WCTE geometry is also affected. At the hit level there is a O(1%) difference, could be due to MC statistical uncertainties (800 particle gun events in CI)

tdealtry (approved):
Thanks @kmtsui 

I was going to suggest using LFS, however it looks like there's a [1 GB / month limit for free github for downloads from LFS](https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-storage-and-bandwidth-usage)... which for a 50 MB file is 20 downloads. This doesn't sound like a lot, especially if you ever want it running in the CI...

So I think the thing to do is just put it in the git repo as normal. [It'll complain](https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-large-files-on-github#file-size-limits) but will be fine

I will caveat that if you're needing to change this file every week, this will quickly increase the WCSim repo size (50 x 50 = 2.5 GB). But I expect that you won't need to do this that regularly! (Even if you do, we could look at [pruning the history for that file](https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-large-files-on-github#removing-files-from-a-repositorys-history))

Another way to do it is to host it somewhere else on the web that can be e.g. `wget`ed using a `get_data.sh` script. But this should be somewhere "official", rather than someone's university webspace that isn't future-proof

Anyway, the code looks good to me. So let me know when you've finalised it, and are happy with the CI results (for both IWCD & WCTE)

PR #519, opened 2025-04-07
Add sea level cosmic muon file for WCTE 

hrubyfr:
Addition of a WCTE_cosmics.dat data file for sea level cosmics simulation for WCTE, added file copy to CMakeLists and added an option for the WCTE macro to use cosmics


kmtsui:
Thanks for the hard work. As discussed in the WCTE meeting, the current implementation of cosmic generator assumes a fixed muon energy per incoming angle, which may not be ideal if we want to simulate low energy muons that stop inside the WCTE volume. Could you run some checks to study whether the low energy components are important in certain incoming angles?
Also, the cosmic muon generator puts the vertex a bit outside the WCTE volume which would affect the muon entrance energy. So if is necessary to simulate the low energy stopping muons in WCTE, we would need to change the vertex generation logic.

macros/WCTE.mac
```diff
@@ -192,6 +192,13 @@
 # In order to run this correctly, the AmBe source must be in the geometry & neutron capture must be activated - for details see the "AmBe CALIBRATION SOURCE" section above
 #/mygen/generator ambeevt
 
+## Or you can use a cosmics generator
+## Based on Reyna model for sea level cosmic muons
```
kmtsui: could you make a reference (arxiv, etc.) of this model?

hrubyfr: Yes, of course, I pushed the link (https://arxiv.org/abs/hep-ph/0604145) to my fork, will that register the change or do I need to do any other step for it to get merged with the main branch properly?

kmtsui: no problem the PR will automatically pick up the change in your branch

hrubyfr: I checked the flux for different angles again, and the bigger angles indeed still have a considerable flux in lower energies, I added a comparison of two angles in attachment. The higher angles are shifted towards higher energies in the mean energy setting by a smaller peak, and they are also more likely to have muons of higher energies. They have maximum flux at energy around ~3 GeV, compared to vertical flux which has maximum at ~1 GeV 
Flux_0_vs_75
Flux_0_vs_75_zoom
Flux_0_vs_75_zoom_nolog

kmtsui: Interesting. Looking at the data from the Reyna paper, those low energy muon data points are mostly from the large theta angles, I wonder how valid to extrapolate it to small theta regions.
And most muons are still above 1 GeV which are going through the WCTE detector, so maybe it's fine to use the mean energy per angle; in the future if we want to study stopping muons, we can make another implementation of energy sampling

kmtsui:
Can you make some validation plots and post it here? e.g. Muon entrance points and energies in the WCTE detector volume, flux shape as a function of theta (separately from the input spectrum and generated muons)

hrubyfr:
OK, I looked at the data, got these histograms for entry and exit points
ee_can
I got energies of the muons going through the boundaries through GetBoundaryKE and got these plots, which are actually missing a few events (might not be going through any boundaries?)
c_en_ex
The energy difference of the entering and exiting particle (i.e. deposited energy) is here: 
c_energy
Finally I looked at the angle distribution which looks a bit weird when looking at purely theta here
c_angles
c_theta
, but when converting to cos(theta) it looks to create them according to the theorized flux cos^2(theta) here
c_cos_theta

kmtsui:
The statistics is a bit low, but the overall shape is reasonable. Maybe zoom in a bit the energy spectra since most muons are below 10 GeV.

hrubyfr:
Oh right, I left that there by mistake, here are the zoomed in plots
c_en_ex
c_energy

kmtsui:
Cool this looks good. @tdealtry are you happy to merge it?

tdealtry:
Thanks @hrubyfr !

PR #520, opened 2025-04-15
Update LI code, better sampling

menailj:
Sample from costheta rather than theta for LI profiles


tdealtry (approved):
Looks good thanks @menailj 
If you could please share some plots (or link to a talk) showing that this is working as expected, that'd be great

And just to check one thing - all users of the LI generator are happy with sampling from `cos(theta)` now? If both are useful, it'd be nice to have a mac file option like `sample_from` with options `theta` and `cos-theta`, so that the user can use what they need. Of course, if `cos-theta`-sampling is better, then this is required. Just don't really want another PR in a few months adding `theta`-sampling support back in...

ekneale:
Hi @menailj this is running quickly now, and the sampled distributions look sensible. Just a few comments: 
1. LightInjectorsDetails.json needs adding to WCSim/CMakelists.txt. 
2. I think that it would be better to define our profiles with theta=0 in the forward direction in the json, rather than having the 90 degree offset in WCSimLIGen.cc, to avoid issues where diffuser profiles are defined over ranges other than (0,180). We can add a comment to LightsInjectors.json to that effect.
3. Just to note, eventually we will have a measured profile for each diffuser, so we will either need to revert to the single json, or we will need to do something like e.g. ID28 for the indexing in LightInjectors.json, where ID/OD refers to the inner/outer detector and the index refers to the corresponding index in LightInjectorsDetails.json. This might be the best way forward, as it gives us the flexibility to continue with current studies using a single profile for now. However, if we are going to do that, we should add an example (plus comments) in LightInjectors.json at this stage so that there is a clear method for anyone else who uses this code.

@tdealtry I have additional changes e.g. changes to the pulsewidth/timing. Would you like me to open a separate pull request, or add them to this one?

menailj:
Hi @ekneale thanks for looking at this. I've added the file to the CMakeList.txt. The other two I can certainly fix as I believe the diffuser water scans are finished (or very near finished). I already have to convert the data format so can add the offset in during the conversion. We will need the second option you suggested as the collimator scans will take a lot longer so having the option to switch will be good. 

I suggest that's covered in a separate PR in future given there will be a lot added including all the scan data. It will also take me a little longer to figure out.

tdealtry:
Thanks @menailj & @ekneale !

Yes, I think let's add the other changes as a separate PR. Thanks

@menailj can you please pull from the latest develop, before you push the angle offset? The CI is almost certainly failing because of the OD fix that went in a couple of days ago

PR #521, opened 2025-04-18
Minimal version of cmake

guiguem:
With the arrival of cmake 4+, cmake 3.4- are no longer supported and the minimal requirement is 3.5


tdealtry:
Thanks for this @guiguem!
I can confirm that all machines & containers that I use satisfy this requirement. This includes both the hk-software & hk-hyperk containers, so anything CentOS8+ should have an appropriate version bundled already (if not, cmake can be installed via e.g. yum (if you have permissions), or by hand)

Will keep this open for a few more days for more people to comment - if this really breaks things for you, please shout!

tdealtry (approved):
Thanks!

PR #522, opened 2025-05-24
save sheet dimensions

tdealtry:
The current radius & length stored in WCSimRootGeom are not so useful
- they are to do with the maximum extent of the PMT position (depends on PMT type)
- with geometries with an OD, them being 

This adds an extra variable such that we can store radii/full lengths of important _boundaries_ (the variable itself is extentable, such that in the future we can store info on other shapes, if required)

This is done for the RealisticPlacement & NoReplica methods for now (I believe that this encompases all official geometries for HK FD, IWCD & WCTE - please do let me know if this is wrong), and in each the following are saved
- ID blacksheet position (effectively ID wall)
- OD inner (next to deadspace) tyvek position (effectively OD inner wall)
- OD outer (next to rock/sand/...) tyvek position (effectively OD outer wall)

When using HK FD,
```
/WCSim/WCgeom HyperK_HybridmPMT_WithOD_Realistic
/WCSim/Construct
```
 this produces
```
-----------------------------------------------
Name         || Radius (cm) || Full length (cm)
-----------------------------------------------
IDBlacksheet || 32400       || 65751
ODInnerTyvek || 33021       || 66993
ODOuterTyvek || 34021       || 70993
-----------------------------------------------
```
which have the same values as the numbers printed out by the RealisticPlacement function

For IWCD, 
```
/WCSim/WCgeom IWCD_mPMT_WithOD
/WCSim/PMT/ReplicaPlacement false
/WCSim/Construct
```
this produces
```
-----------------------------------------------
Name         || Radius (cm) || Full length (cm)
-----------------------------------------------
IDBlacksheet || 3500       || 8000
ODInnerTyvek || 3721       || 8400
ODOuterTyvek || 4628.72       || 9844
-----------------------------------------------
```
@kmtsui I would appreciate it if you could check these IWCD numbers. They look wrong (e.g. ~2 m dead space) I'm obviously missing something in the tyvek geometry positioning definitions... Thanks

### DIFF
--- a/src/WCSimConstructCylinder.cc
+++ b/src/WCSimConstructCylinder.cc
@@ -3718,14 +3718,14 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructCylinderNoReplica()
 
     G4double odTopCapZ[4] = {
       (-WCODDeadSpace+1*mm+WCBlackSheetThickness+pmt_blacksheet_offset),
-      -.5*(WCODTyvekSheetThickness),
-      -.5*(WCODTyvekSheetThickness),
-      .5*(WCODTyvekSheetThickness)};
+      WCBlackSheetThickness,
+      WCBlackSheetThickness,
+      1*(WCODTyvekSheetThickness)+WCBlackSheetThickness};
     G4double odBotCapZ[4] = {
       -(-WCODDeadSpace+1*mm+WCBlackSheetThickness+pmt_blacksheet_offset),
-      .5*(WCODTyvekSheetThickness),
-      .5*(WCODTyvekSheetThickness),
-      -.5*(WCODTyvekSheetThickness)};
+      -WCBlackSheetThickness,
+      -WCBlackSheetThickness,
+      -1*(WCODTyvekSheetThickness)-WCBlackSheetThickness};
     G4double odCapRmin[4] = {WCODRadius-WCODTyvekSheetThickness,
                 WCODRadius-WCODTyvekSheetThickness,
                 0,
--- a/src/WCSimConstructCylinder.cc
+++ b/src/WCSimConstructCylinder.cc
@@ -2710,7 +2710,7 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructCylinderNoReplica()
   // Tell DetectorConstruction about boundary walls
   AddBoundaryWallCylinderDimensions(kBoundaryWallODInnerTyvek,
 									WCODRadius,
-									(WCIDHeight + 2*WCODDeadSpace)); //from CapTyvekPosition creation, without the /2 as we store the full length in the boundary wall variables
+									(WCIDHeight + 2*WCODDeadSpace + 2*WCODTyvekSheetThickness)); //from CapTyvekPosition creation, without the /2 as we store the full length in the boundary wall variables, including the OD tyvek thickness
 
 
   // the radii are measured to the center of the surfaces
--- a/src/WCSimConstructCylinder.cc
+++ b/src/WCSimConstructCylinder.cc
@@ -3718,14 +3718,14 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructCylinderNoReplica()
 
     G4double odTopCapZ[4] = {
       (-WCODDeadSpace+1*mm+WCBlackSheetThickness+pmt_blacksheet_offset),
-      WCBlackSheetThickness,
-      WCBlackSheetThickness,
-      1*(WCODTyvekSheetThickness)+WCBlackSheetThickness};
+      -.5*(WCODTyvekSheetThickness),
+      -.5*(WCODTyvekSheetThickness),
+      .5*(WCODTyvekSheetThickness)};
     G4double odBotCapZ[4] = {
       -(-WCODDeadSpace+1*mm+WCBlackSheetThickness+pmt_blacksheet_offset),
-      -WCBlackSheetThickness,
-      -WCBlackSheetThickness,
-      -1*(WCODTyvekSheetThickness)-WCBlackSheetThickness};
+      .5*(WCODTyvekSheetThickness),
+      .5*(WCODTyvekSheetThickness),
+      -.5*(WCODTyvekSheetThickness)};
     G4double odCapRmin[4] = {WCODRadius-WCODTyvekSheetThickness,
                 WCODRadius-WCODTyvekSheetThickness,
                 0,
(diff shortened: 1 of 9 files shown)


src/WCSimRootGeom.cc
```diff
@@ -182,3 +182,36 @@ bool WCSimRootPMT::CompareAllVariables(const WCSimRootPMT * c) const
...
+    std::cerr << "Boundary type: " << WCSimEnumerations::EnumAsString(s) << " not found" << std::endl;
+    return -999;
+  }
+  return fBoundaryWallDimensions.at(s).at(1); //1 = full length
+}
+//______________________________________________________________________________
+void WCSimRootGeom::PrintBoundaryWallInfo() const {
+  std::cout << "Boundary wall information" << std::endl
+	    << "-----------------------------------------------" << std::endl
+	    << "Name         || Radius (cm) || Full length (cm)" << std::endl
```
kmtsui: Unit should be mm

src/WCSimConstructCylinder.cc
```diff
@@ -2707,6 +2707,11 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructCylinderNoReplica()
 	<< "GEOMCHECK3 WCPMTODExposeHeight \t" << WCPMTODExposeHeight << G4endl
 	;
 #endif
+  // Tell DetectorConstruction about boundary walls
+  AddBoundaryWallCylinderDimensions(kBoundaryWallODInnerTyvek,
+									WCODRadius,
+									(WCIDHeight + 2*WCODDeadSpace)); //from CapTyvekPosition creation, without the /2 as we store the full length in the boundary wall variables
```
kmtsui: should be `WCIDHeight + 2*WCODDeadSpace+2*WCODTyvekSheetThickness`, to be consistent with `WCODRadius` which is the surface touching OD water

kmtsui: Need to also debug the placement position
https://github.com/WCSim/WCSim/blob/cacd2c5a08f2503192f74f4ce861338b9047163e/src/WCSimConstructCylinder.cc#L3719

```
    G4double odTopCapZ[4] = {
      (-WCODDeadSpace+1*mm+WCBlackSheetThickness+pmt_blacksheet_offset),
      WCBlackSheetThickness,
      WCBlackSheetThickness,
      1*(WCODTyvekSheetThickness)+WCBlackSheetThickness};
    G4double odBotCapZ[4] = {
      -(-WCODDeadSpace+1*mm+WCBlackSheetThickness+pmt_blacksheet_offset),
      -WCBlackSheetThickness,
      -WCBlackSheetThickness,
      -1*(WCODTyvekSheetThickness)-WCBlackSheetThickness};
```

src/WCSimConstructCylinder.cc
```diff
@@ -2886,6 +2891,9 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructCylinderNoReplica()
 					  checkOverlaps);
 
 
+	// Tell DetectorConstruction about boundary walls
+	AddBoundaryWallCylinderDimensions(kBoundaryWallODOuterTyvek,
+									  WCRadius, WCLength);
```
kmtsui, line 2896: Need to debug `WCLength`
https://github.com/WCSim/WCSim/blob/cacd2c5a08f2503192f74f4ce861338b9047163e/src/WCSimConstructCylinder.cc#L2719

```
  if (!isODConstructed) WCLength = WCIDHeight + 2*(WCBlackSheetThickness + 1*mm + pmt_blacksheet_offset);
  else WCLength = WCIDHeight + 2*(WCODHeightWaterDepth + WCBlackSheetThickness + WCODDeadSpace + WCODTyvekSheetThickness);
```

tdealtry:
Thanks for this quick review @kmtsui 

For reference, here's the updated IWCD ID+OD geometry numbers (modified by only a couple of mm, and which make a lot more sense when I look at them with the correct units...)
```
-----------------------------------------------
Name         || Radius (mm) || Full length (mm)
-----------------------------------------------
IDBlacksheet || 3500       || 8000
ODInnerTyvek || 3721       || 8402
ODOuterTyvek || 4628.72       || 9842
-----------------------------------------------
```

I've applied the two fixes to `WCLength` & tyvek placement. Thank you for those!
CI will of course now inevitably fail, but I expect the changes to be small.

tdealtry:
I've reverted the two bug fixes, as it has introduced [240 overlap warnings](https://wcsim.github.io/Validation/918fe31bdd97fe973750ffc2a89fd742e760ee33/26/wcsim_iwcdod_SensitiveDetector_Only_NoTrigger_gun5MeVem_xdir_dark0_n1_bad.diff.txt) See for example [here](https://github.com/WCSim/WCSim/actions/runs/15236335062/job/42850779344#step:12:4922). 

The CI will hopefully run ok now, and after I'll just merge the original feature of this PR

kmtsui:
> I've reverted the two bug fixes, as it has introduced [240 overlap warnings](https://wcsim.github.io/Validation/918fe31bdd97fe973750ffc2a89fd742e760ee33/26/wcsim_iwcdod_SensitiveDetector_Only_NoTrigger_gun5MeVem_xdir_dark0_n1_bad.diff.txt) See for example [here](https://github.com/WCSim/WCSim/actions/runs/15236335062/job/42850779344#step:12:4922).

oh right the OD cap PMTs have to be shifted too. in case you want to fix it:
https://github.com/WCSim/WCSim/blob/fc6b3a8cb8104157aad52f437e4536d216c57510/src/WCSimConstructCylinder.cc#L4162
```
        G4double zpos = ((WCIDHeight + 2*WCODDeadSpace)/2)+WCODTyvekSheetThickness+WCBlackSheetThickness;
```
https://github.com/WCSim/WCSim/blob/fc6b3a8cb8104157aad52f437e4536d216c57510/src/WCSimConstructCylinder.cc#L4207
```
          G4ThreeVector topWLSpos = G4ThreeVector(xoffset,
                                                  yoffset,
                                                  ((WCIDHeight + 2*WCODDeadSpace)/2)+WCODTyvekSheetThickness+WCBlackSheetThickness);
```

But maybe I can include in the next IWCD geometry update or another PR. 
@tdealtry which do yo prefer?

tdealtry:
Oops sorry, totally missed your comment here (until I noticed the email notification) I'd feel better if you do it, since you're the expert! Thank you!
Could maybe be included in #518 ?

PR #524, opened 2025-05-28
Added Row Seperation Factor/2

patrickstowell:
ID integration diagrams consist of 5 patterns of 8 rows of PMTs, with a final single pattern of 6 rows of PMTs for the bottom. 

The whole geometry was shifted by one PMT cell width in the vertical direction to align it with the extent of the old geofile (~70cm), but this should have been actually half that distance. The result is the ID PMTs are slightly shifted away from the top cap which is only really noticeable when full PMT drawing is turned on.

Fix is to change to use RowSeperation/4 as this gives the correct offset.

**Before Fix**
Screenshot 2025-05-25 at 23 23 15

Screenshot 2025-05-25 at 23 28 34

Screenshot 2025-05-25 at 23 28 53

**After Fix**
Screenshot 2025-05-25 at 23 34 18

Screenshot 2025-05-25 at 23 34 23

### DIFF
--- a/src/WCSimConstructRealisticPlacement.cc
+++ b/src/WCSimConstructRealisticPlacement.cc
@@ -919,7 +919,9 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructRealisticPlacement()
     // Inprint Columns first then rings
     for (int i = 0; i < 5; i++){
       G4RotationMatrix* imprint_rot = new G4RotationMatrix();
-      G4ThreeVector imprint_pos = G4ThreeVector(0.0,0.0,-i*imprint_spacing + WCIDHeight/2-70*cm);
+      G4ThreeVector imprint_pos = \
+        G4ThreeVector(0.0, 0.0, -i*imprint_spacing + WCIDHeight/2 - RowSeperation/4);
+
       for (int j = 0; j < config.NBlocksPerRing; j++){
         block_assembly->MakeImprint(InnerDetectorLogic, imprint_pos, imprint_rot);
         odblock_assembly->MakeImprint(OuterDetectorLogic, imprint_pos, imprint_rot);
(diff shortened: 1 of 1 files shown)


tdealtry:
This looks great. Thank you!

Picking out some interesting features of [the validation](https://wcsim.github.io/Validation/d4fd0470bf90fb1a804138f86574e3c7ebfd0543/index.html)
- This [removes the remaining 45 geometry overlaps](https://wcsim.github.io/Validation/d4fd0470bf90fb1a804138f86574e3c7ebfd0543/24/wcsim_hkfdcomplete_SensitiveDetector_Only_NoTrigger_gun5MeVem_xdir_dark0_n1_bad.diff.txt) ✨ 
- ID-starting particle guns have NHits consistent within the usual 1%
  - [mPMTs @ 5 MeV electrons](https://wcsim.github.io/Validation/d4fd0470bf90fb1a804138f86574e3c7ebfd0543/8/2/index.html) are larger at ~+4%. But the photon statistics aren't there (~0-7 hits per event) to take a sensible conclusion from this. The mPMTs are consistent @ 500 MeV electrons & muons
- OD-barrel-starting inwards-going 500 MeV electrons have a ~7%-13% drop in 20" hits (~1%-9% drop for mPMT hits). This is the most concerning feature
  - The X-starting & Y-starting are now much more consistent (both 20" & mPMT)
  - There's a drop (to 0) in the [number of killed photons](https://wcsim.github.io/Validation/d4fd0470bf90fb1a804138f86574e3c7ebfd0543/2/wcsim_hkfdcomplete_SensitiveDetector_Only_NoTrigger_gun500MeVemODbarrelIn_startX_dark0_n200_bad.diff.txt) for startX (startY was already 0)

patrickstowell:
On that last point I think we need to check that the OD inward going tracks aren't passing directly through a PMT. Looking at the original geofiles there is a PMT on one wall at x=3.37cm and z=0.14cm, and another at the far wall at x=-3.37cm z=0.14cm. An OD barrelIn Xdir gun would fire the particle directly through the back face of these PMTs.

PR #525, opened 2025-05-29
Changed OD bottom row to match updated layouts.

patrickstowell:
Slight changes to integration diagrams mean bottom row is more filled in now for the OD.
Very subtle change that only affects very bottom row of the OD.
OD still has a gap at the top based on the integration diagrams that needs to be checked with the OD group.

image

Before Change
image

After Change
image


tdealtry:
Here's the number of OD PMTs
| Geom | Caps | Barrels | Total |
| --- | --- | --- | --- |
| New | 542 | 2484 | 3568 |
| Old | 542 | 2448 | 3532 |

The CI checks aren't so informative (this being a change to the bottom row - the checks being for particles going up or sideways, going down). So the results of ±2% change in number of true hits is just stat fluctuations

@federiconova I wonder if you could discuss with the OD group whether you want this change for the MC production (it's my recollection that 3532 PMTs is (or was at some point) the number of OD PMTs expected to be installed)

federiconova:
> Here's the number of OD PMTs
> Geom 	Caps 	Barrels 	Total
> New 	542 	2484 	3568
> Old 	542 	2448 	3532
> 
> The CI checks aren't so informative (this being a change to the bottom row - the checks being for particles going up or sideways, going down). So the results of ±2% change in number of true hits is just stat fluctuations
> 
> @federiconova I wonder if you could discuss with the OD group whether you want this change for the MC production (it's my recollection that 3532 PMTs is (or was at some point) the number of OD PMTs expected to be installed)

Hi, the current exact numbers for the OD should be:
endcaps: 538
barrel: 2484
total: 3560
so quite close to the "NEW" row of the table. In fact we are still trying to add more PMTs on the endcaps, so it might be ok to keep 542.
Thanks!
Federico

tdealtry:
Thanks for the quick reply @federiconova. It sounds like this is closer to expected "reality", so will merge this & issue (another...) release

Thanks for the PR @patrickstowell!

PR #526, opened 2025-06-09
Optical process for Raman scattering in water

arturof:
Added a new optical process to simulate Raman scattering in water. It is a copy of the Rayleigh process with a different mean free path and a wavelength shift after the interaction based on "APPLIED OPTICS / Vol. 37, No. 15 / 20 May 1998".

The Raman process is disabled in this commit, but it can be enabled by commenting out the line added to jobOptions.mac.

Also added the ability to use the Rayleigh scattering mean free path as calculated in Geant4 internally, instead of the hardcoded values in WCSimConstructMaterials.cc from SKdetsim that are probably tuned to SK data. I use Geant4's Rayleigh mean free path when enabling the Raman process to avoid some double counting as the hardcoded values may contain a Raman component.


kmtsui:
@arturof Thanks for the development! We do want to have Raman scattering available for a long time.

To validate your implementation, can you make some plots showing the effect from the new Raman model? For example, can you do some mono-energetic photon bomb simulation, and check the photon energy & fraction of scattered photons when being detected at the PMTs? 

And now because photon energy can change during propagation (maybe even multiple times), I think it's necessary to record the photon energy in some way. Either store the initial and final energy in `WCSimRootCherenkovHitTime`, or the sequence of energy change in `WCSimRootCherenkovHitHistory`.

src/WCSimConstructMaterials.cc
```diff
@@ -1615,7 +1615,7 @@ void WCSimDetectorConstruction::ConstructMaterials()
    myMPT1->AddProperty("RINDEX", ENERGY_water, RINDEX1, NUMENTRIES_water);
    myMPT1->AddProperty("ABSLENGTH",ENERGY_water, ABSORPTION_water, NUMENTRIES_water);
    // M Fechner: new, don't let G4 compute it.
-   myMPT1->AddProperty("RAYLEIGH",ENERGY_water,RAYLEIGH_water,NUMENTRIES_water);
+   if (RAYFF>0) myMPT1->AddProperty("RAYLEIGH",ENERGY_water,RAYLEIGH_water,NUMENTRIES_water);
```
kmtsui, line 1618: I think this is related to your comment of `Also added the ability to use the Rayleigh scattering mean free path as calculated in Geant4 internally, instead of the hardcoded values in WCSimConstructMaterials.cc from SKdetsim that are probably tuned to SK data.`

Better to add explain in `macros/tuning_parameters.mac` on what would happen when `/WCSim/tuning/rayff 0` is used

src/WCSimConstructMaterials.cc
```diff
@@ -1678,7 +1678,7 @@ void WCSimDetectorConstruction::ConstructMaterials()
    G4MaterialPropertiesTable *SilGelPropTable = new G4MaterialPropertiesTable();
    SilGelPropTable->AddProperty("RINDEX", ENERGY_water, RINDEX_SilGel, NUMENTRIES_water);
    SilGelPropTable->AddProperty("ABSLENGTH",ENERGY_SilGel, ABSORPTION_SilGel, 18); //ToDo: get measurement of optical properties of the optical gel. From slides: better than 40cm above 350nm.
-   SilGelPropTable->AddProperty("RAYLEIGH",ENERGY_water,RAYLEIGH_water,NUMENTRIES_water); //ToDo: get actual Rayleigh scattering in gel
+   if (RAYFF>0) SilGelPropTable->AddProperty("RAYLEIGH",ENERGY_water,RAYLEIGH_water,NUMENTRIES_water); //ToDo: get actual Rayleigh scattering in gel
```
kmtsui, line 1681: This is for silicon gel. Do not couple with water implementation.

src/WCSimOpRaman.cc
```diff
@@ -0,0 +1,289 @@
...
+G4PhysicsOrderedFreeVector* 
+WCSimOpRaman::CalculateRamanMeanFreePaths( const G4Material* material ) const
+{
+  G4MaterialPropertiesTable* materialProperties = 
+                                       material->GetMaterialPropertiesTable();
+
+  // If the material doesn't have a RINDEX property vector then return
+  G4MaterialPropertyVector* rIndex = materialProperties->GetProperty("RINDEX");
+  if ( rIndex == NULL ) return NULL;
+
```
kmtsui, line 214: Does that mean Raman scattering is enabled for all materials with `RINDEX` table? Is this intended? Or should it be restricted to Water/GdWater?

src/WCSimOpRaman.cc
```diff
@@ -0,0 +1,289 @@
...
+//
+G4VParticleChange*
+WCSimOpRaman::PostStepDoIt(const G4Track& aTrack, const G4Step& aStep)
+{
+        aParticleChange.Initialize(aTrack);
+
+        const G4DynamicParticle* aParticle = aTrack.GetDynamicParticle();
+
+        if (verboseLevel>0) {
+                G4cout << "Scattering Photon!" << G4endl;
```
kmtsui: Just for more clear verbose
```
G4cout << "Raman scattering Photon!" << G4endl;
```

src/WCSimOpRaman.cc
```diff
@@ -0,0 +1,289 @@
...
+  if ( rIndex == NULL ) return NULL;
+
+  G4PhysicsOrderedFreeVector* ramanMeanFreePaths =
+                                             new G4PhysicsOrderedFreeVector();
+
+  for( size_t uRIndex = 0; uRIndex < rIndex->GetVectorLength(); uRIndex++ )
+  {
+     const G4double energy = rIndex->Energy( uRIndex );
+     const G4double lambda = h_Planck * c_light / energy;
+     const G4double meanFreePath = (3703.7037*m) * pow( lambda / (488*nm) , 5.3);
```
kmtsui, line 222: Can you plot mean free path/cross section for visualization?

src/WCSimConstructMaterials.cc
arturof: Yes, it is related to that. I will add an explanation in `macros/tuning_parameters.mac`.

src/WCSimConstructMaterials.cc
arturof: I am also wondering why the Rayleigh mean free path for water is used for silicon gel. Somebody probably concluded that it is better than not having Rayleigh scattering in silicon gel at all.

The problem is that when RAYFF<=0, the photon entering the silicon gel will Rayleigh scatter all the time or generate an error. We could remove that line and put it back when the correct Rayleigh mean free path for silicon gel is available.

src/WCSimOpRaman.cc
arturof: Yes, it is currently enabled for all materials with `RINDEX` table. I overlooked that by mistake. I will restrict it to Water and GdWater.

src/WCSimOpRaman.cc
arturof: Ok, that makes sense.

src/WCSimOpRaman.cc
arturof: The attached plot compares the mean free path in water of the new Raman process with other optical processes. The absorption and Rayleigh labels refer to WCSim default values. G4Rayleigh to the Rayleigh values calculated by Geant4. 
wcsim_water_mfp

arturof:
To validate the Raman model, I extracted the total attenuation length by fitting the true interaction position distribution in a simulation of 10000 295nm photons starting at the barrel wall and pointing along the cylinder diameter. The expected value is 1/(1/Absorption + 1/Rayleigh + 1/Raman) = 1/(1/571m + 1/41.1m + 1/257m) = 33.4m and the fit result is close to it. 

WCSim_HK_pencil_295nm_RamanOn_ntuple_all_pos

I also checked the final photon wavelength at the PMTs. The Raman model shifts the initial wavelength according to 1/lambda_final = 1/lambda_initial - 3400cm^-1 (it increases by ~30nm at 295nm and ~40nm at 320nm). The first peak in the plot contains photons that didn't scatter or did Rayleigh scattering only. The second peak contains photons that did 1 Raman interaction and the third peak 2 Raman interactions. The first peak is suppressed due to the low PMT QE at 295nm.

WCSim_HK_pencil_295nm_RamanOn_allPMT_histos_Lambda10_scat

arturof:
I am storing the initial and final photon energy in `WCSimRootCherenkovHitTime` and only the total number of Raman interactions in `WCSimRootCherenkovHitHistory` for now. Do I need to increase the version number of these classes in include/WCSimRootEvent.hh?

I plan to store the photon position, energy and interaction type of each step in `WCSimRootCherenkovHitHistory`. Is this reasonable? Is `WCSimRootCherenkovHitHistory` the correct place to do it? In addition to keep track of the energy changes due to the Raman interaction, I need this information to be able to re-weight the absorption, Rayleigh and Raman interaction probabilities.

src/WCSimWCSD.cc
```diff
@@ -412,6 +424,8 @@ G4bool WCSimWCSD::ProcessHits_boundary(G4Step* aStep, G4TouchableHistory*)
         (*hitsCollection)[PMTHitMap[replicaNumber]-1]->AddTrackID(trackID);
         (*hitsCollection)[PMTHitMap[replicaNumber]-1]->AddParentID(parentSavedTrackID);
         (*hitsCollection)[PMTHitMap[replicaNumber]-1]->AddPhotonStartTime(photonStartTime);
+        (*hitsCollection)[PMTHitMap[replicaNumber]-1]->AddPhotonStartEnergy(photonStartEnergy);
+        (*hitsCollection)[PMTHitMap[replicaNumber]-1]->AddPhotonStartEnergy(photonEndEnergy);
```
kmtsui: Bug here

src/WCSimWCSD.cc
```diff
@@ -424,6 +438,8 @@ G4bool WCSimWCSD::ProcessHits_boundary(G4Step* aStep, G4TouchableHistory*)
...
```
kmtsui: Bug here

kmtsui:
> I am storing the initial and final photon energy in `WCSimRootCherenkovHitTime` and only the total number of Raman interactions in `WCSimRootCherenkovHitHistory` for now. Do I need to increase the version number of these classes in include/WCSimRootEvent.hh?

Yes it is.

> I plan to store the photon position, energy and interaction type of each step in `WCSimRootCherenkovHitHistory`. Is this reasonable? Is `WCSimRootCherenkovHitHistory` the correct place to do it? In addition to keep track of the energy changes due to the Raman interaction, I need this information to be able to re-weight the absorption, Rayleigh and Raman interaction probabilities.

I agree.

src/WCSimWCSD.cc
kmtsui: seems these two bugs cause segmentation fault in WCTE tests.

arturof: Thanks for catching that. Somehow it didn't crash when I run it using the HK geometry. I fixed both bugs.

macros/tuning_parameters.mac
```diff
@@ -1,4 +1,6 @@
 # Set tuning parameters
+# Set rayff>0 to scale the Rayleigh mean free path in water hard-coded in WCSimContructMaterials.cc
+# Set rayff<=0 to use the Rayleigh mean free path in water calculated by Geant4
```
tdealtry:
```suggestion
# Set rayff>0 to scale the Rayleigh mean free path in water hard-coded in WCSimContructMaterials.cc, based on SK data
# Set rayff<=0 to use the Rayleigh mean free path in water calculated by Geant4
# If activating Raman scattering, using the Geant4 value is the current recommendation (the SK measurement of Rayleigh scattering may have a Raman effect)
```

include/WCSimOpRaman.hh
```diff
@@ -0,0 +1,188 @@
...
+// Created:     1996-05-31
+// Author:      Juliet Armstrong
+// Updated:     2014-08-20 allow for more material types
+//              2005-07-28 add G4ProcessType to constructor
+//              1999-10-29 add method and class descriptors
+//              1997-04-09 by Peter Gumplinger
+//              > new physics/tracking scheme
+// mail:        gum@triumf.ca
+//
+////////////////////////////////////////////////////////////////////////
```
tdealtry: From what I can tell, you just copied the Rayleigh class, and renamed the class / rewrote the source
This was instead of setting up a brand new class that inherits from `G4VDiscreteProcess` directly
Both would give the same results (one is a bit quicker!) so I don't think this header is required
```suggestion
```

include/WCSimOpRaman.hh
```diff
@@ -0,0 +1,188 @@
...
+        G4bool IsApplicable(const G4ParticleDefinition& aParticleType);
+        // Returns true -> 'is applicable' only for an optical photon.
+
+        void BuildPhysicsTable(const G4ParticleDefinition& aParticleType);
+        // Build thePhysicsTable at a right time
+
+        G4double GetMeanFreePath(const G4Track& aTrack,
+				 G4double ,
+                                 G4ForceCondition* );
+        // Returns the mean free path for Rayleigh scattering in water.
```
tdealtry:
```suggestion
        // Returns the mean free path for Raman scattering in water.
```

include/WCSimOpRaman.hh
```diff
@@ -0,0 +1,188 @@
...
+        // --- Not yet implemented for other materials! ---
+
+        G4VParticleChange* PostStepDoIt(const G4Track& aTrack,
+                                       const G4Step&  aStep);
+        // This is the method implementing Rayleigh scattering.
```
tdealtry:
```suggestion
        // This is the method implementing Raman scattering.
```

include/WCSimOpRaman.hh
```diff
@@ -0,0 +1,188 @@
...
+        }
+}
+
+inline G4PhysicsTable* WCSimOpRaman::GetPhysicsTable() const
+{
+  return thePhysicsTable;
+}
+
+
+#endif /* G4OpRayleigh_h */
```
tdealtry:
```suggestion
#endif /* WCSimOpRaman_h */
```

include/WCSimOpRaman.hh
```diff
@@ -0,0 +1,188 @@
...
+        WCSimOpRaman& operator=(const WCSimOpRaman &right);
+
+public:
+
+        ////////////
+        // Methods
+        ////////////
+
+        G4bool IsApplicable(const G4ParticleDefinition& aParticleType);
+        // Returns true -> 'is applicable' only for an optical photon.
```
tdealtry:
```suggestion
        /// Returns true -> 'is applicable' only for an optical photon.
        G4bool IsApplicable(const G4ParticleDefinition& aParticleType);
```
Since we use doxygen, and I guess they use something bespoke, it'd be good to do this. Essentially
- move the comment line(s) above, rather than below the function
- add an extra `/` to make it a triple `///`

This is a general comment for this (and other) files. But if it's a massive pain, then don't worry

src/WCSimOpRaman.cc
```diff
@@ -0,0 +1,293 @@
...
+//              > allow for positiv and negative CosTheta and force the
+//              > new momentum direction to be in the same plane as the
+//              > new and old polarization vectors
+//              2001-01-29 by Peter Gumplinger
+//              > fix calculation of SinTheta (from CosTheta)
+//              1997-04-09 by Peter Gumplinger
+//              > new physics/tracking scheme
+// mail:        gum@triumf.ca
+//
+////////////////////////////////////////////////////////////////////////
```
tdealtry: Again, I think you can delete this
```suggestion
```

tdealtry (changes requested):
Thanks @arturof Looks great! Thanks for bringing in a new feature like this!

Please increment `ClassDef`
- [Here](https://github.com/WCSim/WCSim/blob/develop/include/WCSimRootEvent.hh#L198)
- [And here](https://github.com/WCSim/WCSim/blob/develop/include/WCSimRootEvent.hh#L226)

Is it required to add new Raman enums [here](https://github.com/tdealtry/WCSim/blob/fc6b3a8cb8104157aad52f437e4536d216c57510/include/WCSimEnumerations.hh#L60)? e.g. `kOpRaman` (or `fOpRaman` - I'm confused given WCSimOpProcessSubType)

tdealtry:
Thanks for implementing my suggestions @arturof 

Think there's just the ClassDef request & enum question, then this can be merged. Thanks again for your work on this!

arturof:
Thanks for your input @tdealtry

I don't think it is required to have the Raman process in the EProcessType enumeration. It seems that it is used only to save the process that created a photon that created a hit, but the Raman process doesn't create photons. However, I added the Raman process anyway because the other optical processes (Rayleigh, Mie, absorption and boundary which don't create photons either) are included.

I incremented ClassDef as requested, but also added an additional feature that I mentioned before and was planing to include in this PR: I added the photon scattering and reflection positions to `WCSimRootCherenkovHitHistory` to be able to re-weight the mean free path of the Raman process and the other optical processes. I could open another PR for it if needed.

tdealtry (approved):
Thanks @arturof

PR #528, opened 2025-07-21
Make geometry boundary wall container work with old root

tdealtry:
Use int instead of an enum in the map. Older versions of ROOT (e.g. 6.20.04) can write the file, but see a std::bad_alloc when they try to read it...


kmtsui (approved):
Looks good!

PR #529, opened 2025-07-21
Make LI generator work with old C++ standard

tdealtry:
Pass two values by reference, instead of using C++17 feature of creating a pair from a function return. This is an issue for the old HK container

@ekneale @menailj please take a look!

### DIFF
--- a/src/WCSimLIGen.cc
+++ b/src/WCSimLIGen.cc
@@ -13,10 +13,11 @@
 using json = nlohmann::json;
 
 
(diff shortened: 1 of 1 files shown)


ekneale:
This looks fine to me but I'll let @menailj comment as this part of the code was authored by her.

tdealtry:
Thanks Liz. Given the urgency of this (need another tag for the HK MC production) I'm going to merge now.
I've verified (via printouts) that the costheta/phi numbers inside `SampleFromInterpolatedSurface()` are identical to the ones in `GeneratePhotons()`, so it should be fine. But of course, please shout if you see issues!

PR #530, opened 2025-07-22
Add WCSim version that was used to generate the file to WCSimRootOptions

tdealtry:
This is taken from the git tag & git hash, from `git describe --always --long --tags --dirty` 
For example, a _clean_ git status is `v1.12.23-1-g3a5d2087` which shows
- the latest tag
- the number of commits since the last tag
- the current hash (after a trailing `g`)
When a modification to a tracked file occurs, then we get an extra `-dirty`: `v1.12.23-1-g3a5d2087-dirty`

Why like this? IWCD are already using it in their Settings tree, as I rediscovered yesterday. It's not quite what we spoke about Mathieu, but it actually combines both options (version & hash) into one. Let me know what you think - it'll be easy to modify what's actually put in the variable

### DIFF
--- a/include/WCSimRootOptions.hh
+++ b/include/WCSimRootOptions.hh
@@ -59,6 +59,10 @@ public:
   virtual ~WCSimRootOptions();
   void Print(Option_t *option = "") const;
 
+  //WCSim version gets
+  string GetWCSimVersion() {return WCSimVersion;}
+  //WCSim version sets
+  void SetWCSimVersion(string iWCSimVersion) {WCSimVersion = iWCSimVersion;}
   //WCSimDetector* gets
   void SetDetectorName(string iDetectorName) {DetectorName = iDetectorName;}
   void SetGeomHasOD(bool iGeomHasOD) {GeomHasOD = iGeomHasOD;}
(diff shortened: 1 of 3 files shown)


guiguem:
This is sooo great! thanks!!

guiguem (approved):

guiguem (commented):
Maybe you need to increment line 236 of WCSimRootOptions?

nickwp:
I just came across this PR. It doesn't really matter, but was it known that this exact feature already existed in WCSim for years and now is duplicated? The git hash has has always been saved in the Settings tree (see [here](https://github.com/WCSim/WCSim/blob/develop/src/WCSimRunAction.cc#L132-L135)) and now it's saved in both the Settings tree and the WCSimRootOptionsT tree. Is it needed in both?

nickwp:
Ah I guess this must have been known, since it seems the addition to save it in the WCSimRootOptionsT uses the same implementation

tdealtry:
The Settings tree only exists for "nuprism" detectors, so we needed to add something for other detectors

I guess we could remove it from the Settings tree now? But this could mess up some IWCD analysis scripts for very little disk space gain (one string leaf in a one-entry TTree), so I'm inclined to leave things as is

nickwp:
Ah I see. I didnt know that the Settings tree was only for "nuprism" detectors.
I highly doubt that there are any automatic scripts that use the git hash from the Settings tree, but only used manually for tracing back which version was used to generate some files (in fact I did this myself yesterday). So probably it can be removed, but like you say I guess it's not too much of a problem to leave it there.

One benefit of the way it's stored in the Settings tree is that you don't need to load the library to read it (which is particularly useful if you don't know which WCSim version was used to make the file - so don't know which library version to load). You can just open the file in ROOT, ignore the warnings, and show the first entry in the tree, which tells you the version used to make the file. I'm not sure if that's possible when it's embedded within the WCSimRootOptions class.

PR #531, opened 2025-07-29
RooTracker generator position improvements & checks

sk1806:
Currently IWCD events from the current and previous IWCD NEUT vectors will not go through WCSim because they fail the check on the coordinates designed to ignore events that fall outside of the detector.  There are a few reasons for this.

1) The unit conversion in WCSim assumes the neutrino interaction vertex (EVtVtx) is in [cm], but it is in [m], which is corrected in this PR

2) When writing the output in nuPRISMGeom it adds the plane position back on to the neutrino interaction vertex (rather than leaving it so the centre of the detector is the origin of the coordinate system).  And in WCSim the y offset is left off when correcting this back.  I expect this may be something to do with how the generation of different planes was done in the past maybe?  But it means when trying to process this production, it doesn't work.  I have added that back on in this PR.

3) There is an explicit check that the nu vertex is inside the detector, but it checks the ID not the full detector.   I checked myDetector->GetWaterTubeLength and this does seem to give the correct full length when compared to here:  https://github.com/WCSim/WCSim/pull/522 so we could use that instead.  But I can't find anything obvious in that class to get the full radius. 
Another worry is that these position checks rely on the x, y, z definition, but this will be different for HKFD where y and z are swapped... so I am not sure how to reconcile that.
We could remove the check, but it does feel like a nice safety net to have in place (especially since it is already proving to be an obvious point of failure).
And also  y position has this subtracted,  myDetector->GetWCIDVerticalPosition(), which always seems to be zero... again I guess left over from previous treatment of different vertical planes?  So I think we should remove that too?  

I also noticed that even with these fixes, I am getting some events outside the 'y' by a small margin  (actually outside my world box around the det which is confusing), so I need to check what is going on there, but it overshoots in +y and -y so I think this is just something from neutgeom that I need to look into.

The final point is actually that I think we should just remove the bit in nuPRSIMGeom where it adds the plane position on
https://github.com/hyperk/nuPRISMGeom/blob/fb630ce959ffb3888f93988b63db1e0e8b8a1f10/nuPRISMGeom/TNeutOutput.cc#L835
This happens at the end of nuPRISMGeom when writing the output and then as soon as it is read in WCSim it is subtracted back off again.  So I don't see the point (we anyway have the plane position in the output file if needed for something else, e.g. tracking which vertical plane we are in).  This currently creates a point where IWCD and ND280 diverge and seeing as I would like to merge the two, leaving the coordinates centred on the geom seems like a sensible thing to do, provided IWCD group are happy with that?  (please comment)
So then, that would actually change this PR.  However, it would be nice to have an official release so I can process current vectors from nuPRISMGeom as a reference, and then do a new PR to process vectors after nurPRISMGeom and neutgeom merge and validate.  I guess I don't really need an official release for that if it is hassle though, so I'll leave that up to Tom.


tdealtry:
Many thanks for this Sophie! I'll let @kmtsui give the final sign off for this (along with the associated nuPRISMGeom suggestion), but generally looks good to me

#### On IWCD vs FD
I think we'll just need to use something like this in the appropriate place
```
if(mydetector->GetIsNuPrism()) {
  // IWCD-specific stuff
}
else {
  // FD-specific stuff
}
```
but I think doing the IWCD in this PR and having a different PR for FD is fine.

#### Remaining offset
This is weird, and shows that a check is really required! I guess I can suggest plotting vertex positions from the input & output, and applying offsets until they completely overlap (but @kmtsui may have ideas about what's really going on)

#### Crazy idea
but do we make the "where is the vertex?" check more complicated? Add a new config option, that can hopefully be a vector of string (if G4 supports it)
```
/mygen/vertexVolume ID deadspace OD rockshell
```
then do a check to see if the vertex is within one of the allowed regions.

For HKFD, the different water regions [have different geometry names](https://github.com/patrickstowell/WCSim/blob/develop/src/WCSimConstructRealisticPlacement.cc#L52), so we can do some kind of lookup to see the volume name we're inside
- unless this breaks when we're inside a PMT...
- or actually, we may need to do the below IWCD option. Whilst the water volume "cylinders" aren't actually cylinders (they're many-side polygons because of the superframes in the PMT support structure, and so to tell if you're really in a region you need to know the volume name, rather than average radius), I'm pretty sure the rootracker vertex distributions will be "true" cylinder based

For IWCD, I think we need to use the radius/length numbers. I'd suggest something like
```
vector<float> dims = mydetector->GetBoundaryWallDimensions()[kBoundaryWallIDBlacksheet];
radius = dims[0];
full_length = dims[1];
```
checking [here](https://github.com/tdealtry/WCSim/blob/fc6b3a8cb8104157aad52f437e4536d216c57510/include/WCSimEnumerations.hh#L7) for the other boundary positions. I think this will mean that we need to add the rock shell dimensions to that list

sk1806:
OK, so I have updated my PR based on Tom's suggestions.

I added a new command to WCSimPrimaryGeneratorMessenger, so that in the mac file you can specify
/mygen/region REGION
where REGION can be:   ID, OD_INNER, OD_OUTER

And this is then used in WCSimPrimaryGeneratorAction, where its handled with an enum class.
The dimensions for IDBlacksheet, ODInnerTyvek and ODOuterTyvek dimensions are retrieved, and the location of the event is checked against whichever region you specified. 
If you don't specify, it will default to ID.

IWCD and HKFD should now also switch y and z automatically to have the correct axis when determining if an event is in the correct region.
IWCD: axis = y, radius (x,z)
HKFD: axis = z, radius (x,y)

All of this only applies to RooTracker input.

Caveat - the only check I have done is to run this over an IWCD neut vector, using the new command, which worked.  I should also check an HKFD but I don't have a file to hand, but so far things look OK.  Let me know what you think.

RELEASENOTES.md
```diff
@@ -520,6 +520,12 @@ Update
 * Pull Request [#308](https://github.com/WCSim/WCSim/pull/308): Update validation tests
 * Pull Request [#309](https://github.com/WCSim/WCSim/pull/309): Dynamically expand `WCSimRootTrigger`s `TObjArray`
 
+************************************************************
```
kmtsui: mis-insert here?

src/WCSimPrimaryGeneratorMessenger.cc
```diff
@@ -219,6 +219,15 @@ WCSimPrimaryGeneratorMessenger::WCSimPrimaryGeneratorMessenger(WCSimPrimaryGener
   mPMTLEDIdCmd2->SetGuidance("Set LED id for mPMT LED source position, dTheta and dPhi for LED direction. Defaults to 0, 0.0, 0.0 ");
   mPMTLEDIdCmd2->SetParameterName("mPMTLEDId2","mPMTLEDId2_dTheta","mPMTLEDId2_dPhi", true);
   mPMTLEDIdCmd2->SetDefaultValue(G4ThreeVector(0,0.0,0.0));
+
+  RegionCmd = new G4UIcmdWithAString("/mygen/region", this);
```
kmtsui, line 230: Can you add and explain this command in `WCSim.mac`?

sk1806:
OK, so I was looking into why that README file had those changed when I pulled in recent changes before updating my PR... I merged master into my branch, but I should have merged develop ... apologies, not usually a WCSim developer!   Eugh... OK, let me fix this.

sk1806:
That should have tidied it up.

sk1806:
Reading back over Tom's comments (probably should have recapped that before I ploughed in and started making changes) ...   If I understood Tom's comment, there was a suggestion to check which volume we are in for HKFD, but for IWCD to use the average radius, but then why the different approach for the two dets ?   
In anycase I just did the radius approach for both as I'd forgotten that comment, but with the point about the geom actually being a many sided polygon, then maybe this isn't the best approach?  But I'm still a bit confused why we need to treat HKFD and IWCD with these two different approaches?

Anyway, IWCD is the priority right now, and I think the suggested approach for that has been implemented, and we can then deal with HKFD when we get to it, so I'm happy to have feedback and maybe we can just start testing and see where we are at...

kmtsui:
I don't see an obvious problem and am happy with merging it and start testing with some production

tdealtry:
Thanks @sk1806

PR #532, opened 2025-08-01
Update Radon generator

gpronost:
This pull request is an update of the Radon model to the version develop in May 2024 (sorry for the quite late update here). 

- The model is heavily simplified compared with the previous one, in order to make it easier to change its parameter (currently, only the input water Rn concentration can be modified)
- The number of scaling scenario from SK to HK is decreased to 1 (scenario A)

I also added a modification in WCSim.cc, in order to prevent compilation failure when G4VIS_USE is not defined


src/WCSimPrimaryGeneratorMessenger.cc
```diff
@@ -612,6 +626,11 @@ void WCSimPrimaryGeneratorMessenger::SetNewValue(G4UIcommand * command,G4String
     {
       myAction->SetRadonSymmetry(radonGeoSymCmd->GetNewIntValue(newValue));
     }
+
```
tdealtry: I _think_ there's some uncessary repeating going on here...

https://github.com/WCSim/WCSim/blob/0572ad2814287c6e5fa9fc8f77572d3d99bc410e/src/WCSimPrimaryGeneratorMessenger.cc#L574-L592

Looks remarkably similar to

https://github.com/WCSim/WCSim/blob/0572ad2814287c6e5fa9fc8f77572d3d99bc410e/src/WCSimPrimaryGeneratorMessenger.cc#L615-L633

src/WCSimPrimaryGeneratorMessenger.cc
```diff
@@ -771,8 +790,7 @@ void WCSimPrimaryGeneratorMessenger::RadonScalingCommand(G4String newValue)
   G4String scenario = next();
```
tdealtry, line 769: Making this the scaling command a `G4UIcmdWithAnInteger` (rather than a string) would, I think, allow this function to be removed

tdealtry: Another benefit would be if people are using an old mac file with A or B in it, having it be an `int` would fail the config reading, forcing people to engage with the fact that the model has been changed

src/WCSimPrimaryGeneratorAction.cc
```diff
@@ -137,8 +137,9 @@ WCSimPrimaryGeneratorAction::WCSimPrimaryGeneratorAction(
   // Radioactive and Radon generator variables:
   radioactive_sources.clear();
   myRn222Generator	= 0;
-  fRnScenario		= 0;
+  fRnScenario		= 1;
   fRnSymmetry		= 1;
+  fRnWaterConc		= 2.63;
```
tdealtry: It'd be good to specifiy what units this is in, and where the number came from

src/WCSimPrimaryGeneratorMessenger.cc
```diff
@@ -196,19 +196,27 @@ WCSimPrimaryGeneratorMessenger::WCSimPrimaryGeneratorMessenger(WCSimPrimaryGener
...
-  radonScalingCmd->SetCandidates("0 A B");
+  radonScalingCmd->SetGuidance("     SCENARIO : 0, 1");
+  radonScalingCmd->SetCandidates("0 1");
   param = new G4UIparameter("SCENARIO",'s',true);
-  param->SetDefaultValue("A");
+  param->SetDefaultValue("1");
   radonScalingCmd->SetParameter(param);
   
   radonGeoSymCmd = new G4UIcmdWithAnInteger("/mygen/radon_symmetry",this);
   radonGeoSymCmd->SetGuidance("Select scalling scenario");
```
tdealtry: Typo
```suggestion
  radonGeoSymCmd->SetGuidance("Select scaling scenario");
```

src/WCSimPrimaryGeneratorMessenger.cc
```diff
@@ -196,19 +196,27 @@ WCSimPrimaryGeneratorMessenger::WCSimPrimaryGeneratorMessenger(WCSimPrimaryGener
...
+  radonGeoSymCmd->SetGuidance("[usage] /mygen/radon_symmetry SYMMETRY ");
   radonGeoSymCmd->SetGuidance("     SYMMETRY : 1 ... ");
   param = new G4UIparameter("SYMMETRY",'d',true);
   param->SetDefaultValue("1");
   radonGeoSymCmd->SetParameter(param);
+  
+  radonWaterConcCmd = new G4UIcmdWithADouble("/mygen/radon_water_concentration",this);
+  radonWaterConcCmd->SetGuidance("Select scalling scenario");
+  radonWaterConcCmd->SetGuidance("[usage] /mygen/radon_water_concentration WATCONC ");
+  radonWaterConcCmd->SetGuidance("     WATCONC : (double) activity in input water ");
```
tdealtry: Add the units

tdealtry (changes requested):
Thanks @gpronost! Some mostly minor suggestions

Please also update [the radon section in WCSim.mac](https://github.com/WCSim/WCSim/blob/develop/WCSim.mac#L385) to reflect the new options.
It'd also be very useful to add some information there on how to convert from radon options to total time simulated (I was going to ask you in an email about this today actually). Is it as simple as `radioactive_time_window x number of events`? Or something more complicated?

If you want to put a plot (or some other validation you've done) on the PR, then that'd be welcome

gpronost:
> It'd also be very useful to add some information there on how to convert from radon options to total time simulated (I was going to ask you in an email about this today actually). Is it as simple as `radioactive_time_window x number of events`? Or something more complicated?

Thanks for the comments, I updated the PR with the requested changes.

The total time simulated is more complex. It actually depends on the model (so if the input water concentration is changed, if the dimension of detector change, or if the number of PMT change, then the result will be different). The generator should debug something about `Mean activity in the full ID:`, in mBq. Using `number of event / activity` give the time equivalent duration in seconds. 

`radioactive_time_window` needs to be used if `decayBiasProfile` is used, it's a simulation trick, which induce a bias in the decay simulation by forcing the decay to happen within the specified time windows. If it's not used, then the timing of the decay can become very large (hours or days), which WCSim was not handling properly before, this was why I introduced the `/DAQ/RelativeHitTime true` option at this time. I'm not sure what is the situation currently.

tdealtry (approved):
Thanks @gpronost Will merge when CI passes (just a random failure related to `yum install`, I think)

PR #534, opened 2025-08-20
extend wavelength range down to 200 nm (hard coded)

federiconova:
extend wavelength range down to 200 nm (needed by OD)

### DIFF
--- a/src/WCSimPMTQE.cc
+++ b/src/WCSimPMTQE.cc
@@ -40,7 +40,7 @@ G4double WCSimDetectorConstruction::GetPMTQE(G4String CollectionName, G4double P
   // return 0 for wavelength outside the range
   if (flag==1){
     /*ODCOMMENT -- We use photons < 280nm so we need ->*/ //if (PhotonWavelength <= low_wl || PhotonWavelength >= high_wl ){
-    if (PhotonWavelength <= low_wl || PhotonWavelength >= high_wl || PhotonWavelength <=280 || PhotonWavelength >=660){
+    if (PhotonWavelength <= low_wl || PhotonWavelength >= high_wl || PhotonWavelength <=200 || PhotonWavelength >=660){
       return 0;
     }
   }else if (flag==0){
--- a/src/WCSimStackingAction.cc
+++ b/src/WCSimStackingAction.cc
@@ -54,10 +54,10 @@ G4ClassificationOfNewTrack WCSimStackingAction::ClassifyNewTrack
 	if(!DetConstruct->GetIsCombinedPMTCollectionDefined()) {
 	  //Geometry without OD
 	  if (DetConstruct->GetPMT_QE_Method()==1){
-	    wavelengthQE  = DetConstruct->GetPMTQE(WCIDCollectionName,photonWavelength,1,240,660,ratio);
+	    wavelengthQE  = DetConstruct->GetPMTQE(WCIDCollectionName,photonWavelength,1,200,660,ratio);
 	    //wavelengthQE  = DetConstruct->GetPMTQE(volumeName,photonWavelength,1,240,660,ratio);
 	  }else if (DetConstruct->GetPMT_QE_Method()==2){
-	    wavelengthQE  = DetConstruct->GetPMTQE(WCIDCollectionName,photonWavelength,0,240,660,ratio);
+	    wavelengthQE  = DetConstruct->GetPMTQE(WCIDCollectionName,photonWavelength,0,200,660,ratio);
 	    //wavelengthQE  = DetConstruct->GetPMTQE(volumeName,photonWavelength,0,240,660,ratio);
 	  }else if (DetConstruct->GetPMT_QE_Method()==3 || DetConstruct->GetPMT_QE_Method() == 4){
 	    wavelengthQE = 1.1;
@@ -66,9 +66,9 @@ G4ClassificationOfNewTrack WCSimStackingAction::ClassifyNewTrack
 	else {
 	  //Geometry with OD
 	  if (DetConstruct->GetPMT_QE_Method() == 1){
-	    wavelengthQE  = DetConstruct->GetStackingPMTQE(photonWavelength,1,240,660,ratio);
+	    wavelengthQE  = DetConstruct->GetStackingPMTQE(photonWavelength,1,200,660,ratio);
 	  }else if (DetConstruct->GetPMT_QE_Method() == 2){
-	    wavelengthQE  = DetConstruct->GetStackingPMTQE(photonWavelength,0,240,660,ratio);
+	    wavelengthQE  = DetConstruct->GetStackingPMTQE(photonWavelength,0,200,660,ratio);
 	  }else if (DetConstruct->GetPMT_QE_Method() == 3 || DetConstruct->GetPMT_QE_Method() == 4){
 	    wavelengthQE = 1.1;
 	  }
--- a/src/WCSimWCSD.cc
+++ b/src/WCSimWCSD.cc
@@ -202,12 +202,12 @@ G4bool WCSimWCSD::ProcessHits(G4Step* aStep, G4TouchableHistory*)
     photonQE = 1.1;
   }else if (fdet->GetPMT_QE_Method()==2){
     // maxQE = fdet->GetPMTQE(WCIDCollectionName,wavelength,0,200,700,ratio);
-    maxQE = fdet->GetPMTQE(WCCollectionName,wavelength,0,240,660,ratio);
-    photonQE = fdet->GetPMTQE(volumeName, wavelength,1,240,660,ratio);
+    maxQE = fdet->GetPMTQE(WCCollectionName,wavelength,0,200,660,ratio);
+    photonQE = fdet->GetPMTQE(volumeName, wavelength,1,200,660,ratio);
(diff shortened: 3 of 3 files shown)


tdealtry:
Hi Federico. Thanks for this. Sorry for the delay

We discussed this PR at the software meeting yesterday. There are 2 important questions
- The number of hits doesn't change in any of the CI (e.g. [here](https://wcsim.github.io/Validation/9593000faa5cef3612d4ce3febc693dc7d674500/2/3/in)). This is of course exactly what we want for the ID PMTs. But the OD PMTs should surely have an increased hit rate?
- How does the runtime of WCSim change with this? The CI system is not appropriate to do this (random selection of runners of different powers), so I wonder if you can run the old/new version on your system, and provide the runtime. WCSim does this automatically (techincally Geant4) providing printout like this at the end of the event loop (where the times are total, rather than per event). This is strongly preferable to anything externally timing the running of WCSim, as this neglects the startup/shutdown, which can be 10s of seconds.
```
Run Summary
  Number of events processed : 100
  User=0.21s Real=0.32s Sys=0.11s
```

tdealtry:
To add one thing, I know we've spoken in the past about where the min/max wavelengths should be hard-coded or set in a config file. I would say
- if ID hit rate and runtime are unchanged, then keeping it hardcoded is fine
- if one of them is changed, then we need to make it a config option

I know you've had a look into this already, and see crashes when running. If we do need it, then, at that point, please share the branch name with me where you've tried this, and I can try to debug. It may be that the option needs to go in an earlier file than WCSim.mac (jobOptions.mac or tuning_parameters.mac)

federiconova:
Hi, I will address the comments.
(1) about timing
Time to run from this merge request:
Run Summary
  Number of events processed : 10
  User=0.38s Real=0.39s Sys=0s
vs time from develop branch:
Run Summary
  Number of events processed : 10
  User=0.39s Real=0.4s Sys=0.01s
Seems ok?

federiconova:
(2) about the OD hits
Well let's see. What kind of WLS plate is this tested on? Eljen? Kuraray? Actually none of them is a good test, the real WLS plate we will use is manifactured by INR institute in Russia. I have added a modeling of this plate, but that merge request has not been approved yet. I am not sure exactly how the WLS plate is modeled in the develop branch, but I know I have changed the modeling substantially in my merge request. Anyway, if more photons are added in but no more hits are detected, in my opinion this points to a mistake in the modeling of the WLS plate in the develop branch?

federiconova:
(3) about hard-coded values
I did add the parameters to tuning rather than to the mac file. I no longer have those modifications - I deleted them in order to make this merge request acceptable.

tdealtry:
Thanks for the quick replies. I am worried that this isn't having any physics output effect at all - it could be that there is a missing piece somewhere that's causing the extended range photons not to correctly propagate. So I think I'd prefer to put this merge on hold, add in the geometry changes, then come back to this. Does that sound reasonable?

1. In terms of the test - thank you. But 10 events is too low stats for a test like this. I'm also interested in the physics used - were a lot of 200-240 nm photons expected to be produced? I'd suggest that events with a lot of photons would be useful. _Maybe_ muons, but _maybe_ photon bombs with flat 200-660 nm wavelengths (I don't actually know the wavelength of Cherenkov emmission, and whether a different distribution would be more useful), and since we're testing how things slow in the non-OD part of the detector that doesn't require the 200-240 nm photons, then the photon bombs should be in the ID. But to reiterate, I'd strongly suggest holding off on doing this until after other PRs are added

2. In terms of WLS used, it's [EljenEJ286](https://github.com/WCSim/WCSim/blob/develop/src/WCSimDetectorConfigs.cc#L1000). Unfortunately, I don't see your INR WLS PR in [your list of PRs](https://github.com/WCSim/WCSim/issues?q=is%3Apr%20author%3Afedericonova)

3. Ok no worries. We'll see how this goes 1 goes later, and whether the config option needs to be added

federiconova:
the Cerenkov spectrum goes like 1/lambda, so yes you have lots of UV - more than anything else!

I will make a merge request to store the wavelength information. this is in my mind the best way to check which wavelengths are proagated.

Yes, believe it or not have added INR modeling ; didn't pull it yet, cause what is the point if every pull request gets years for approval?

tdealtry:
Ok thank you. Anything that shows that this is working as expected would be great.

But I would remind you of what I sent in the email last week. The photon energy [is now stored in WCSimRootCherenkovHitTime](https://github.com/WCSim/WCSim/blob/develop/include/WCSimRootEvent.hh#L195) I would expect that this is what you need?

federiconova:
> Ok thank you. Anything that shows that this is working as expected would be great.
> 
> But I would remind you of what I sent in the email last week. The photon energy [is now stored in WCSimRootCherenkovHitTime](https://github.com/WCSim/WCSim/blob/develop/include/WCSimRootEvent.hh#L195) I would expect that this is what you need?

mmm. did you just commit this? it does not compile for me:

[100%] Linking CXX executable WCSim
/usr/bin/ld: libWCSimCore.so.1.12.24: undefined reference to `WCSimOpRaman::WCSimOpRaman(G4String const&, G4ProcessType)'
collect2: error: ld returned 1 exit status

Also, can you explain how I access this information? I have a hit object of class WCSimRootCherenkovDigiHit. How do I get the photon energy from it?

federiconova:
> mmm. did you just commit this? it does not compile for me:
> 
> [100%] Linking CXX executable WCSim /usr/bin/ld: libWCSimCore.so.1.12.24: undefined reference to `WCSimOpRaman::WCSimOpRaman(G4String const&, G4ProcessType)' collect2: error: ld returned 1 exit status

nevermind, I managed to compile using hk-hyperk

tdealtry:
Hi Federico. I reproduced your error locally. The issue is that the presence of the new class files requires cmake to be run for the `make` command to be aware it has to build a new file. I presume there's a nice way to do it, but the nuclear option is to `rm -rf $WCSIM_BUILD_DIR/*`, before running your cmake & make commands again

In terms of going from a digitised hit->true hit, that information is available. See https://github.com/WCSim/WCSim/blob/develop/sample-root-scripts/sample_readfile.C#L293 as an example (I would note that your addition in #468 was [also in WCSimRootCherenkovHitTime](https://github.com/WCSim/WCSim/pull/468/files#diff-d0fbde985ecc6e7fa70b62e7663b05ba309acb889cb399d1bc576337047e9bfeR173), so any testing scripts you have for that will _almost_ work - just need e.g. `GetPhotonStartEnergy()` (or `GetPhotonEndEnergy()`) instead of `GetWavelength()`)

federiconova:
> Hi Federico. I reproduced your error locally. The issue is that the presence of the new class files requires cmake to be run for the `make` command to be aware it has to build a new file. I presume there's a nice way to do it, but the nuclear option is to `rm -rf $WCSIM_BUILD_DIR/*`, before running your cmake & make commands again
> 
> In terms of going from a digitised hit->true hit, that information is available. See https://github.com/WCSim/WCSim/blob/develop/sample-root-scripts/sample_readfile.C#L293 as an example (I would note that your addition in #468 was [also in WCSimRootCherenkovHitTime](https://github.com/WCSim/WCSim/pull/468/files#diff-d0fbde985ecc6e7fa70b62e7663b05ba309acb889cb399d1bc576337047e9bfeR173), so any testing scripts you have for that will _almost_ work - just need e.g. `GetPhotonStartEnergy()` (or `GetPhotonEndEnergy()`) instead of `GetWavelength()`)

yes thanks, I've found it as well
I'm testing now...

federiconova:
OK this will be more convinging. I made two branches in which I artificially extended the QE of PMTs down to 200 nm, so the new photons can actually be detected. I also made the QE flat (= 100%), so we see the Cerenkov spectrum without distortion. Then I've run it with and without extending the photon propagations down to 200 nm. Result:

Screenshot 2025-08-26 at 18 11 53

You can clearly see the 1/lambda shape of the Cerenkov spectrum. You can also clearly see the hard cut at 280 nm (develop branch) being moved down to 200 nm (my branch).

Is this convicing enough?

federiconova:
as for the timing. Develop branch:
  Number of events processed : 1000
  User=381.73s Real=391.58s Sys=0.81s

my branch:
Run Summary
  Number of events processed : 1000
  User=377.24s Real=386.75s Sys=0.72s

tdealtry:
Thank you for this Federico. Sorry I was going to ask a question about your orignal plot that had a worrying normalisation difference (then was sucked into a 3 hours meeting...). But this plot is a great smoking gun that this works as expected

Last question - you've done 1000 events for the benchmark test. What was the setup for these events?

federiconova:
the geometry is realistic and I shot 2.2 MeV electrons from the center of the tank in the x direction

federiconova:
I suspect you'd like to see the time performance for higher light level, so I tried simulating 100 muons, of 600 MeV energy.

Develop branch:
  Number of events processed : 100
  User=1481.2s Real=1485.56s Sys=0.06s

extended wavelength range:
  Number of events processed : 100
  User=1269.91s Real=1274.11s Sys=0.05s

tdealtry:
Thanks very much for these tests! I'm confused as to why the new code runs faster... but you've done lots of tests (as has the CI) and this seems to be real 🤷

PR #536, opened 2025-09-01
create feature/add-kuraray-plate

federiconova:
add a class for Kuraray WLS plate (on top of the existing Eljen WLS plate)
and set the Kuraray plate as the default choice for the FD OD


src/WCSimConstructMaterials.cc
```diff
@@ -1551,7 +1551,11 @@ void WCSimDetectorConstruction::ConstructMaterials()
   // OpWaterWLSSurface->SetSigmaAlpha(0.1); // TODO: What's this?
 
   // MATERIAL properties
+#if 0
```
kmtsui: I don't think this is a good implementation. Making a configurable command like those in `macros/tuning_parameters.mac` is better in my opinion.

src/WCSimDetectorConfigs.cc
```diff
@@ -445,7 +445,11 @@ void WCSimDetectorConstruction::SetHyperKWithODGeometry()
   isCombinedPMTCollectionDefined=true;
 
   // TEST WLS collection for stacking action
+#if 0
```
kmtsui: The `WLSType` value can also be specified by `macros/tuning_parameters.mac` in a similar way

src/WCSimConstructMaterials.cc
federiconova: well, the thing i am submitting does work. I don't know how to make it better. I have tried, but couldn't. Although the eljen type was there before, it was not assigned via tuning parameters. So I am really just replicating what was there before.
Can you suggest how to actually change it? Or else please just approve it.

src/WCSimDetectorConfigs.cc
federiconova: how?

src/WCSimConstructMaterials.cc
kmtsui: if you are completely replacing Eljen by Kuraray (and not using Elijen again), that it works. 
But if someone needs to use Elijen afterwards then one needs to change the hard-coded part and recompile it, which makes it super inconvenient. 

By creating a new command in the `WCSimTuningMessenger` class, this config can be loaded with `macros/tuning_parameters.mac` and be accessed here (and in `src/WCSimDetectorConfigs.cc`)

federiconova: yes I tried but I couldn't do it. Can you actually try to implement it?
It should have been done even before for the eljen class, so I don't really think it's my job to fix it

tdealtry:
Thanks for this @federiconova [Checking the CI](https://wcsim.github.io/Validation/dabf00a09e245a9f047a45c7cb7bc3b267813e8b/index.html)
- HKFD is
  - unchanged for the ID-starting particles (that don't go into the OD). Great!
  - Basically unchanged for the ID for OD-starting particles going up (the one scatter from OD into ID doesn't happen anymore). Great!
  - 1-3% drop in OD hits for OD-starting particles (going up & in). Is this what you expect Federico?
  - [9% drop in ID 20" raw hits for OD-starting electron at +x](https://wcsim.github.io/Validation/dabf00a09e245a9f047a45c7cb7bc3b267813e8b/2/1/index.html) Looking at the code, I think it's just photon properties that are different between the two plates - energy loss by electrons/muons shouldn't be effected - is that right? The stats aren't great either (200 events), so could just be that.
- IWCD is changed, with small (sub % level) changes in ID hits for ID-starting particles. I presume this is random number offset (these CI particles do get in the OD, causing the random number stream to change -> difference of ~1%. Seen this kind of thing before)
  - What is worrying at first glance is the [ID-starting electron OD hits](https://wcsim.github.io/Validation/dabf00a09e245a9f047a45c7cb7bc3b267813e8b/15/3/index.html) being reduced dramatically (38%). However, this is on events with ~no OD hits (mean 7.5->4) & low stats on events that actually do have hits (small fraction of 800 events, it looks like. So I think this is fine

As to why IWCD changes output, I think this shows that there is an issue with the implementation in this PR (the DetectorConfigs for IWCD still say Eljen, but the output is different - I believe that this is because the ConstructMaterials is set to Kuraray). It _might_ not matter, but it might... Sorry for not knowing the answer to this. @kmtsui what should the IWCD WLS plate model used be? Eljen? Kuraray? Something different to be added later?

Depending on the answer, there might have to be changes to the PR
- if the IWCD & FD WLS plates are the same, then the implementation is fine
- if the IWCD & FD WLS plates are different, then I think a new material needs to be added, such that the IWCD WLS plates can use one material, and FD WLS plates can use the other

In terms of whether the option to swap between Kuraray & Eljen is required to be a config option, or compilation option, I agree that it's only to be config if there is still uncertainty in what's bought / studies to be done on different plate types. But the fact that there are changes to be made in 2 files, and can be confusing to know exactly where (see the 4 similarly named detector geometries for an example!) then I'd suggest the compilation option could be improved. e.g.
- adding something like this to a new `include/Build.h` (or some better named file)
```
#ifndef HKFD_OD_WLS_PLATE_TYPE
#define HKFD_OD_WLS_PLATE_TYPE Kuaray
//#define HKFD_OD_WLS_PLATE_TYPE EljenEJ286
#endif
``` 
- `#include`ing this new file in DetectorConfigs & ConstructGeometry
- using string comparison on `HKFD_OD_WLS_PLATE_TYPE` to decide what to construct

kmtsui:
> I believe that this is because the ConstructMaterials is set to Kuraray

Yes I agree this is the cause.

> what should the IWCD WLS plate model used be? 

We will try to use the same WLS plate as the FD if possible, but of course the final decision depends on the physics performance. That's why it's better to have a easy switch between the two (or more) options.

federiconova:
Hello, I'm answering inline

> Thanks for this @federiconova [Checking the CI](https://wcsim.github.io/Validation/dabf00a09e245a9f047a45c7cb7bc3b267813e8b/index.html)
> 
>     * HKFD is
>       
>       * unchanged for the ID-starting particles (that don't go into the OD). Great!
>       * Basically unchanged for the ID for OD-starting particles going up (the one scatter from OD into ID doesn't happen anymore). Great!
>       * 1-3% drop in OD hits for OD-starting particles (going up & in). Is this what you expect Federico?

Yes or at least it does not look wrong. The Kuraray plate is less efficient than the Eljen plate. The reason why we discarded the Eljen plate is cost.

>       * [9% drop in ID 20" raw hits for OD-starting electron at +x](https://wcsim.github.io/Validation/dabf00a09e245a9f047a45c7cb7bc3b267813e8b/2/1/index.html) Looking at the code, I think it's just photon properties that are different between the two plates - energy loss by electrons/muons shouldn't be effected - is that right? The stats aren't great either (200 events), so could just be that.

Yes, but also nothing should change at all for the ID (as there are no WLS plates in the ID).

> 
>     * IWCD is changed, with small (sub % level) changes in ID hits for ID-starting particles. I presume this is random number offset (these CI particles do get in the OD, causing the random number stream to change -> difference of ~1%. Seen this kind of thing before)
>       
>       * What is worrying at first glance is the [ID-starting electron OD hits](https://wcsim.github.io/Validation/dabf00a09e245a9f047a45c7cb7bc3b267813e8b/15/3/index.html) being reduced dramatically (38%). However, this is on events with ~no OD hits (mean 7.5->4) & low stats on events that actually do have hits (small fraction of 800 events, it looks like. So I think this is fine
> 

Sorry I am a bit lost with the nomenclature. If I understand correctly, these are events in which an electron (of 500 MeV?) is started in the ID; and you are saying that the number of OD hits is reduced by 38%? But also you are saying these are events with almost no OD hits. Indeed, there should only be dark noise hits in the OD if the electron is in the ID. Notice that in reality the WLS plate does have an impact on dark noise (it increases it), but I think this is not included in the simulations in any way. I suppose this can only be statistics? Or else there is some leakage of light from ID into OD? (which again is something we are concerned about in real life, but should not be there in simulations I think). Or maybe sometimes the electron does make it from the ID into the OD? But it seems the vertex is always in the tank center, so that's hard to believe.

> 
> As to why IWCD changes output, I think this shows that there is an issue with the implementation in this PR (the DetectorConfigs for IWCD still say Eljen, but the output is different - I believe that this is because the ConstructMaterials is set to Kuraray). It _might_ not matter, but it might... Sorry for not knowing the answer to this. @kmtsui what should the IWCD WLS plate model used be? Eljen? Kuraray? Something different to be added later?
> 

Let me comment on this. I can speak for the FD OD, but as I understand IWCD will try to replicate the FD OD anyway.
So, the Eljen plate is not going to be used, for cost issues. We have two candidates at the moment: the Kuraray plate (which is still costly because it comes from an external company), versus some custom-made plates produced by our colleagues at INR institute in Moscow. The INR choice looks much more likely and I would say almost definitive. However, for the INR plates I do not have as complete modeling as for the Kuraray plate. So my intention is, after this merge is approved, to make another merge request in which INR plates are also added, and perhaps taken as the default choice; although, as I say, the modeling of these is not very complete.


> Depending on the answer, there might have to be changes to the PR
> 
>     * if the IWCD & FD WLS plates are the same, then the implementation is fine
> 
>     * if the IWCD & FD WLS plates are different, then I think a new material needs to be added, such that the IWCD WLS plates can use one material, and FD WLS plates can use the other
> 
> 
> In terms of whether the option to swap between Kuraray & Eljen is required to be a config option, or compilation option, I agree that it's only to be config if there is still uncertainty in what's bought / studies to be done on different plate types. But the fact that there are changes to be made in 2 files, and can be confusing to know exactly where (see the 4 similarly named detector geometries for an example!) then I'd suggest the compilation option could be improved. e.g.
> 

My opinion is that it would be much preferable to be able to switch from the config file (the mac file or the tuning file). But I just can't manage to implement it - I have tried, but failed. If more experienced people can do it, this would be the preferred choice. But I just don't know how to do it. I know how to pass parameters into the Detector Construction, but I can't make it work for the WLS plates construction.

>     * adding something like this to a new `include/Build.h` (or some better named file)
> 
> 
> ```
> #ifndef HKFD_OD_WLS_PLATE_TYPE
> #define HKFD_OD_WLS_PLATE_TYPE Kuaray
> //#define HKFD_OD_WLS_PLATE_TYPE EljenEJ286
> #endif
> ```
> 
>     * `#include`ing this new file in DetectorConfigs & ConstructGeometry
> 
>     * using string comparison on `HKFD_OD_WLS_PLATE_TYPE` to decide what to construct

mmm ok well this would still require recompiling, but at least it would be centralized in one place, so I can see it is better.
I have tried to implement this, but can you give me a little help?
I have created the Build.h file, but how do I make the string comparison for this pre-compiled string? The line:
if( HKFD_OD_WLS_PLATE_TYPE == "EljenEJ286" )
does not work.
And my other question; I have included Build.h into WCSimDetectorConfigs.cc, but which file do you mean by "ConstructGeometry" ?
Thanks
Federico

tdealtry:
Thanks Federico. So the reason things are tricky with the config option is when things are read. The material construction is done before WCSim.mac (or similar) is read, hence the need to do it in a different part of the code. So it should be done in `tuning_parameters.mac`, which is read in via `WCSimTuningMessenger`. How to _nicely_ pass the info from there to the `WCSimDetectorConstruction` class, I don't know...

>  And my other question; I have included Build.h into WCSimDetectorConfigs.cc, but which file do you mean by "ConstructGeometry" ?
Ahh sorry. `WCSimConstructMaterials.cc`

In terms of the compilation way of doing things, this should be string friendly
Build.h
```
#pragma once
#include <string>
extern const std::string HKFD_OD_WLS_PLATE_TYPE;  // Declaration only
```
Build.cc
```
#include "Build.h"
const std::string HKFD_OD_WLS_PLATE_TYPE = "Kuraray";  // Definition
```
WCSimConstructMaterials.cc & WCSimDetectorConfigs.cc
```
#include <Build.h>
```
and in the relavant place(s)
```
if(HKFD_OD_WLS_PLATE_TYPE == "Kurary")
 //do stuff...
else if(HKFD_OD_WLS_PLATE_TYPE == "Eljen")
 //do other stuff...
```

On the IWCD plate model, there is actually another consideration - that we're currently part way through the IWCD MC production. I presume it'd be easier to analyse if the IWCD OD didn't change between MC files (@kmtsui please correct me if I'm wrong), so I think it would be benefitical to have the two different materials defined separately, such that we can easily use Eljen for IWCD MC & Kurarray for FD MC. (Even if long term the materials are likely to be the same)

federiconova:
> Thanks Federico. So the reason things are tricky with the config option is when things are read. The material construction is done before WCSim.mac (or similar) is read, hence the need to do it in a different part of the code. So it should be done in `tuning_parameters.mac`, which is read in via `WCSimTuningMessenger`. How to _nicely_ pass the info from there to the `WCSimDetectorConstruction` class, I don't know...

Yeah, that's where I got stuck as well...

> 
> > And my other question; I have included Build.h into WCSimDetectorConfigs.cc, but which file do you mean by "ConstructGeometry" ?
> > Ahh sorry. `WCSimConstructMaterials.cc`
> 
> In terms of the compilation way of doing things, this should be string friendly Build.h
> 
> ```
> #pragma once
> #include <string>
> extern const std::string HKFD_OD_WLS_PLATE_TYPE;  // Declaration only
> ```
> 
> Build.cc
> 
> ```
> #include "Build.h"
> const std::string HKFD_OD_WLS_PLATE_TYPE = "Kuraray";  // Definition
> ```
> 
> WCSimConstructMaterials.cc & WCSimDetectorConfigs.cc
> 
> ```
> #include <Build.h>
> ```
> 
> and in the relavant place(s)
> 
> ```
> if(HKFD_OD_WLS_PLATE_TYPE == "Kurary")
>  //do stuff...
> else if(HKFD_OD_WLS_PLATE_TYPE == "Eljen")
>  //do other stuff...
> ```
> 

Ok great many thanks, I'm trying to implement it this way. If it works I will commit it.

> On the IWCD plate model, there is actually another consideration - that we're currently part way through the IWCD MC production. I presume it'd be easier to analyse if the IWCD OD didn't change between MC files (@kmtsui please correct me if I'm wrong), so I think it would be benefitical to have the two different materials defined separately, such that we can easily use Eljen for IWCD MC & Kurarray for FD MC. (Even if long term the materials are likely to be the same)

I am entirely convinced that it is beneficial to have different materials defined separately; in fact we should have three (Eljen, Kuraray, INR). Likewise I will add FDOD 3 inch PMTs (separate from the IWCD ones, and of two different types because again we have 2 candidates for the FDOD), and also different Tyveks. This will all come (hopefully) in separate merge requests that I am pushing serially one after the other.

But, about the IWCD, I undestand that you don't want to change mid-way the simulations, but I also would  advise @kmtsui and IWCD people to wait for better modeling before doing their MC campaign. I mean, having lots of files simulated with the wrong plate will not be as useful as having one month later files with the right plate. Anyway, this is the strategy the FDOD is taking and that is why the OD group is pressuring me (and us) to get these merges done with.

tdealtry:
Great thanks for this @federiconova 

So I think the last thing that's required for this to be merged is to separate the WLS material into IWCD/FD WLS material

On the MC production, I'll send an email to gather some more expertise

federiconova:
> Great thanks for this @federiconova
> 
> So I think the last thing that's required for this to be merged is to separate the WLS material into IWCD/FD WLS material
> 
> On the MC production, I'll send an email to gather some more expertise

Thanks, I've seen your email..
Ah I didn't understand it. So should I also add another precompiled string to control the IWCD plate type?

federiconova:
> > Great thanks for this @federiconova
> > So I think the last thing that's required for this to be merged is to separate the WLS material into IWCD/FD WLS material
> > On the MC production, I'll send an email to gather some more expertise
> 
> Thanks, I've seen your email.. Ah I didn't understand it. So should I also add another precompiled string to control the IWCD plate type? (notice that the current string already says FDOD)

Opps I think I closed this by mistake (was just trying to submit the comment)

federiconova:
> Great thanks for this @federiconova
> 
> So I think the last thing that's required for this to be merged is to separate the WLS material into IWCD/FD WLS material
> 
> On the MC production, I'll send an email to gather some more expertise

allright, I've added the option for the IWCD, and for now it is left as Eljen by default.
I am not sure where it should be used in ConstructMaterial (or elsewhere), I think it's better if IWCD people do this...
Let me know if this is ready to go or how the CI's go
thanks

src/WCSimConstructMaterials.cc
```diff
@@ -1551,10 +1553,48 @@ void WCSimDetectorConstruction::ConstructMaterials()
...
+  G4double* WLSRIndex;
+  G4double* WLSPhotonEnergy_ABS;
+  G4double* WLSAbs;
+  G4int WLSNumEntries_ABS;
+  G4double* WLSPhotonEnergy_EM;
+  G4double* WLSEm;
+  G4int WLSNumEntries_EM;
+  G4int WLSNumEntries;
+  
+  if( HKFD_OD_WLS_PLATE_TYPE == "EljenEJ286" ){
```
kmtsui, line 1644: I think this `if else`also enforce makes IWCD WLS to take the same choice as FD

src/WCSimDetectorConstruction.cc
```diff
@@ -654,6 +654,10 @@ WCSimWLSProperties *WCSimDetectorConstruction::CreateWLSObject(G4String WLSType)
     WCSimWLSProperties* WLS = new EljenEJ286;
```
kmtsui, line 654: On second thought, why not move the `WLS_PVT` material definition and `WlsPlasticMPT` table from `src/WCSimConstructMaterials.cc` to this `if else`. Then one can just change the `WLSType` variable in `src/WCSimDetectorConfigs.cc`. 
Although one still needs to recompile to change the WLS option, at least FD and IWCD are decoupled.

src/WCSimConstructMaterials.cc
federiconova: allright guys, look
three days ago I made a merge request from this hash id: dabf00a09e245a9f047a45c7cb7bc3b267813e8b
This version was compiling and working, it would give Kuraray plate to FDOD and Eljen plate to IWCD. Can we please go back to that version and approve it?

I really don't care about the style issue of implementing precompiled strings; also I really don't think it's my job to do it, because stuff was hard-coded even before. If you feel strongly about it, you should make your own merge request and fix it. What I care about is have a code that works and move quickly to the next merge requests that I need to do.

tdealtry: Hi Federico. The issue is that this PR (both currently & in the hash you mention) changes the FD (as you want) but also changes the IWCD (as we don't). Why did we not have the option to use separate models for the IWCD & FD? We didn't need this functionality, because just one model was defined. So it is a required new feature of this PR

Since you've had problems with implementation, I've implemented in federiconova/WCSim#1. So please take a look at that, and merge if you like it. That should trigger the CI here, so we will be able to do the last check before merging, hoping to see
- no changes for IWCD
- the same changes as seen previously for FD
As we've discussed this isn't the perfect solution (config file would be better - it'd be more user friendly, and we wouldn't need to worry about multiple materials), but it'll do for now (we'll make things better in GHOST)

federiconova: Ok done; I merged, git pulled, and it seems to be running ok on my installation

federiconova: > Hi Federico. The issue is that this PR (both currently & in the hash you mention) changes the FD (as you want) but also changes the IWCD (as we don't). Why did we not have the option to use separate models for the IWCD & FD? We didn't need this functionality, because just one model was defined. So it is a required new feature of this PR

I see, only now I understand the issue. You are right about this.

tdealtry:
Ok great this is now doing what's expected - changing things for the FD, and not for IWCD. Thanks for this @federiconova

PR #537, opened 2025-09-03
Update IWCD geometry

kmtsui:
Re-implement some IWCD geometry update from PR https://github.com/WCSim/WCSim/pull/518

Major changes

1. Update ID and OD PMT tables
   - Use in-situ mPMT instead of default one
   - Increase barrel mPMT vertical separation from 67.4 to 71 cm
   - Make ID and OD PMT numbers the same
2. Change cap volume from tube to Polyhedra, to avoid overlap with OD Tyvek
3. Debug Tyvek overlap check
4. Debug Tyvek dimensions


kmtsui:
Detector display
image

kmtsui:
Due to change in cap construction, there is less than 1% change in ndigihits in WCTE, which is within statistical fluctuations.

For IWCD, there is one `GeomNav1002` navigation warning, but I don't see any warning from overlap checks, so I think it is mostly fine.

tdealtry:
Thanks for this @kmtsui. The code changes look good to me, so generally happy to merge.
However... given that we're part way through the IWCD MC production, I _think_ we don't want to replace the default IWCD geometry with this one. Not only would we need to respin the particle gun MC already produced (@sk1806 please correct me if I'm wrong), given the mPMT separation has changed, I presume that fiTQun would need retuning too (@bquilain please correct me if I'm wrong)
So I'd suggest that this actually becomes a new geometry (e.g. IWCD_Sept2025 & IWCD_Sept2025_With_OD), rather than replacing the current IWCD geometries. I understand that there will still be some small statistical change to the output for IWCD (as WCTE has seen)

kmtsui:
Right I forgot about the ongoing production. The PMT positions are specified by two txt files, so I have reverted the changes and create two separated txt for the new geometry. This should preserve the old results (within statistical fluctuations).

kmtsui:
aargh the Tyvek dimension debug changes the OD dimensions and PMT positions, so there is no way to recover the exact old geometry. For the sake of production, maybe it's better to merge this after the production is launched. 

@sk1806 any opinions?

PR #538, opened 2025-09-08
add INR plate

federiconova:
this is adding the option of a WLS plate produced by INR institute. This is the most likely to be chosen for HK.

I submit the merge request now, but we are discussing in the OD group if indeed the modeling I submit is the best one. It may have to be slightly changed later. But at least the mechanics of having another plate candidate is there.


tdealtry:
Looks good. Thanks. Assuming the last CI job finishes as green, I'd be happy to merge now. Then you can submit another PR to modify just the model inside WCSimWLSProperties.cc (and maybe activating it as default in Build.cc) later if you need. Does this sound like the best plan?

federiconova:
> Looks good. Thanks. Assuming the last CI job finishes as green, I'd be happy to merge now. Then you can submit another PR to modify just the model inside WCSimWLSProperties.cc (and maybe activating it as default in Build.cc) later if you need. Does this sound like the best plan?

Hi Tom, thanks. Yes that is exactly my plan. It's possible that we won't actually need to modify it, but this is under discussion and if that happens it can be a separate PR. And yes I agree INR should be set as default when the model is finalised. I am running some studies...
Thanks again

tdealtry:
Perfect. Thanks!

PR #539, opened 2025-09-12
Feature/fix inr modeling

federiconova:
this merge request is mainly to fix the modeling of some properties of the WLS plate for the FDOD (though some more changes could come in the future as more measurements are done)

It also sets the INR plate as the default standard for the FDOD (instead of the Kuraray plate), reflecting the most likely configuration


src/WCSimWLSProperties.cc
```diff
@@ -707,21 +707,33 @@ void Kuraray::SethEm() {
...
+      3.5424,
+      3.75709,
+      3.8745,
+      4.06505,
+      4.1328,
+      6.1992
```
tdealtry: These should have a `*eV` for consistency (the way it currently is _could_ be a bug)

tdealtry (changes requested):
Thanks @federiconova. Just the one code comment

CI looks a little up/down on the OD hits. In your studies, are the number of hits for Kurary->Inr plate as you/OD group expects?
- otherwise the non HKFD are identical, & HKFD ID changes for OD-starting particles look reasonable

src/WCSimWLSProperties.cc
federiconova: good catch

federiconova:
> CI looks a little up/down on the OD hits. In your studies, are the number of hits for Kurary->Inr plate as you/OD group expects?
> 

yes naturally the modeling will change the n of hits. I have a fresh set of slides about all the changes produced these merges; many details are carefully looked at and seem ok. I could show them at a software meeting if there is interest?
Thanks

macros/tuning_parameters_old.mac
```diff
@@ -0,0 +1,23 @@
+# Set tuning parameters
```
kmtsui: Is this file necessary?

macros/tuning_parameters.mac
```diff
@@ -17,7 +17,7 @@
 # Input file to set photocathode parameters
 /WCSim/tuning/PMTCathodePara data/CathodeParameters.txt
 
-/WCSim/tuning/WCODWLSCladdingReflectivity 0.90
+/WCSim/tuning/WCODWLSCladdingReflectivity 0.99
```
kmtsui: Is this just for INR plate, or applicable for all plates?

macros/tuning_parameters_old.mac
federiconova: Hi 
I thought this would be useful. It just sets the tyvek reflectivity to a lower value. It is what we use to simulate the aged detector (after 10 years of running). We need to assess how the detector will perform after ageing.
I can take the file out if it bothers people.

macros/tuning_parameters.mac
federiconova: This applies to all plates. It reflects the very good reflectivity of the film we will use. Anyway the effect is marginal.

Screenshot 2025-09-16 at 05 55 49

tdealtry:
This is good discussion about the tuning parameter files. Thanks for catching it!

A suggestion on how to proceed
- Add a `tuning_parameters_iwcd.mac` - this should be what the original version of `tuning_parameters.mac` is, so can be used in MC prods, and IWCD can modify this independently of FD
- Add a `tuning_parameters_hkfd.mac` - this should be this PRs version of `tuning_parameters.mac` i.e. with higher cladding reflectivity
- Rename `tuning_parameters_old.mac` as `tuning_parameters_hkfd_aged.mac`
- WCTE folk may want to use the IWCD file, or may want to create their own.
- Remove `tuning_parameters.mac`

The main point here is to put in a breaking change (removal of `tuning_parameters.mac`) such that WCSim users will have to actively choose which tuning file to use

@kmtsui does this sound reasonable?

federiconova:
> This is good discussion about the tuning parameter files. Thanks for catching it!
> 
> A suggestion on how to proceed
> 
>     * Add a `tuning_parameters_iwcd.mac` - this should be what the original version of `tuning_parameters.mac` is, so can be used in MC prods, and IWCD can modify this independently of FD
> 
>     * Add a `tuning_parameters_hkfd.mac` - this should be this PRs version of `tuning_parameters.mac` i.e. with higher cladding reflectivity
> 
>     * Rename `tuning_parameters_old.mac` as `tuning_parameters_hkfd_aged.mac`
> 
>     * WCTE folk may want to use the IWCD file, or may want to create their own.
> 
>     * Remove `tuning_parameters.mac`
> 
> 
> The main point here is to put in a breaking change (removal of `tuning_parameters.mac`) such that WCSim users will have to actively choose which tuning file to use
> 
> @kmtsui does this sound reasonable?

sounds good to me, thanks, I can do it if approved

kmtsui:
Yes the proposed changes in tuning files are good and clear

federiconova:
done
come to think about it, since tuning_parameters.mac has been renamed, all the CI will fail (they now need to point to tuning_parameters_hkfd.mac or tuning_parameters_iwcd.mac)

kmtsui:
The build process failed due to the missing `tuning_parameters.mac`, which can be solved by adding and removing the relevant lines around [here](https://github.com/WCSim/WCSim/blob/b3d3f2d8bdd66ab22263d078ee761346550ef424/CMakeLists.txt#L204).

On the other hand the CI has its [own tuning parameter file](https://github.com/WCSim/Validation/blob/master/Generate/macReference/tune/tuning_parameters.mac). In principle we don't need to change it as this is only used in the validation process. But @tdealtry can decide.

kmtsui:
And there are conflicts in `include/WCSimDetectorConstruction.hh` and `src/WCSimDetectorMessenger.cc` due to your previous PR, it's better to pull the develop branch and resolve them.

federiconova:
> And there are conflicts in `include/WCSimDetectorConstruction.hh` and `src/WCSimDetectorMessenger.cc` due to your previous PR, it's better to pull the develop branch and resolve them.

done

tdealtry:
Great, this builds now!

So [the CI](https://wcsim.github.io/Validation/3ff6d42ac84151a04e896dd255127419a1f21c54/index.html) shows that
- Everything except HKFD, with OD starting particles is identical. Great!
- 20" & mPMTs for HKFD with OD starting particles are the same within ~2-3%. This is reasonable (it's low stats, designed for testing OD Nhits, etc)
- ~20% drop in the number of OD true hits (closer to 17% for digitised OD hits). Is this what you expect?

Note that this uses [this tuning file](https://github.com/WCSim/Validation/blob/master/Generate/macReference/tune/tuning_parameters.mac)
```
/WCSim/tuning/WCODWLSCladdingReflectivity 0.90
/WCSim/tuning/WCODTyvekReflectivity 0.90
```

federiconova:
> Great, this builds now!
> 
> So [the CI](https://wcsim.github.io/Validation/3ff6d42ac84151a04e896dd255127419a1f21c54/index.html) shows that
> 
>     * Everything except HKFD, with OD starting particles is identical. Great!
> 
>     * 20" & mPMTs for HKFD with OD starting particles are the same within ~2-3%. This is reasonable (it's low stats, designed for testing OD Nhits, etc)
> 
>     * ~20% drop in the number of OD true hits (closer to 17% for digitised OD hits). Is this what you expect?
> 
> 
> Note that this uses [this tuning file](https://github.com/WCSim/Validation/blob/master/Generate/macReference/tune/tuning_parameters.mac)
> 
> ```
> /WCSim/tuning/WCODWLSCladdingReflectivity 0.90
> /WCSim/tuning/WCODTyvekReflectivity 0.90
> ```

I think this works as expected. About the change in n of hits, well it depends; you tell me that this is obtained having 90% reflectivity, but what was the value for the reference run against which you are comparing? If it was, for example, 95% (i.e. the default value), than it certainly makes sense.

tdealtry:
Ahh sorry yes. 0.90 both before & after (the validation `tuning_parameter` files haven't changed). So this _just_ takes into account the change from Kuraray -> INR (with fixed model)

federiconova:
> Ahh sorry yes. 0.90 both before & after (the validation `tuning_parameter` files haven't changed). So this _just_ takes into account the change from Kuraray -> INR (with fixed model)

I see. Yes I get lost as to what has been merged yet and what hasn't.
So yes I did studies and INR performs a bit worse than Kuraray, so I think this is expected and can be merged.
By the way, would there be any benefit in me showing a presentation to the software group about all these changes?
Thanks
Federico

tdealtry (approved):
Perfect thanks for confirming.

Yes, if you could give a talk, that'd be wonderful! Thank you. You could contact Miura-san to see if there's space in the premeeting (if it doesn't clash with your other premeetings, which is an issue I have...)

PR #540, opened 2025-09-13
add FDOD PMTs and set PMT3inchR14374_FDOD as default choice

federiconova:
add the two candidates for FDOD PMTs

also, set the default OD choice to be 3 inch Hamamatsu PMT with 30 cm WLS plate


tdealtry:
Thanks for this @federiconova!

I'm very happy that you've updated the defaults to be 3" & 30cm. This'll avoid a lot of confusion in the future (and it's nice that's there are still the options for OD experts to modify the settings for studies)

In terms of CI, this is doing what I think it should be
- IWCD & WCTE are unchanged
- number of FD ID hits is unchanged, other than statistical fluctuations
  - except for OD starting particles, aiming into the ID ([here](https://wcsim.github.io/Validation/45cfcafb207885cf788c6b7a43fc01e0a01637c5/2/1/index.html) and [here](https://wcsim.github.io/Validation/45cfcafb207885cf788c6b7a43fc01e0a01637c5/3/1/index.html)), where number of true hits change by ~10%. This seems fine to me.
    - the path the particle travels on will be different (e.g. previously could have gone through a PMT or WLS, can miss that entirely now)
    - the test is a bad one (as I've mentioned for a while)
- number of FD OD hits is reduced, by ~70%. I won't ask you what the expected reduction is (because that is arbitrary). But it would be good for you to confirm that the number of hits for some known simulation of yours is what you/OD group expectes

Other than that one question, and confirming that this PR is ready, I'm happy to merge this. Thanks!

federiconova:
> Thanks for this @federiconova!
> 
> I'm very happy that you've updated the defaults to be 3" & 30cm. This'll avoid a lot of confusion in the future (and it's nice that's there are still the options for OD experts to modify the settings for studies)
> 
> In terms of CI, this is doing what I think it should be
> 
>     * IWCD & WCTE are unchanged
> 
>     * number of FD ID hits is unchanged, other than statistical fluctuations
>       
>       * except for OD starting particles, aiming into the ID ([here](https://wcsim.github.io/Validation/45cfcafb207885cf788c6b7a43fc01e0a01637c5/2/1/index.html) and [here](https://wcsim.github.io/Validation/45cfcafb207885cf788c6b7a43fc01e0a01637c5/3/1/index.html)), where number of true hits change by ~10%. This seems fine to me.
>         
>         * the path the particle travels on will be different (e.g. previously could have gone through a PMT or WLS, can miss that entirely now)
>         * the test is a bad one (as I've mentioned for a while)
> 
>     * number of FD OD hits is reduced, by ~70%. I won't ask you what the expected reduction is (because that is arbitrary). But it would be good for you to confirm that the number of hits for some known simulation of yours is what you/OD group expectes
> 
> 
> Other than that one question, and confirming that this PR is ready, I'm happy to merge this. Thanks!

Hi
Yes that the OD hits are reduced makes sense. The detecting unit is now 25% in size, so it may well detect only 25% of the light.
I think it's ready to merge.
Thanks

PR #543, opened 2025-09-17
assign correct thicknesses to FDOD WLS plates

federiconova:
this is to fix the default values of the thickness of the different types of WLS plates for the FDOD
Hopefully it's implemented in a way to change nothing for the IWCD

### DIFF
--- a/src/WCSimDetectorConfigs.cc
+++ b/src/WCSimDetectorConfigs.cc
@@ -979,6 +979,17 @@ void WCSimDetectorConstruction::SetHyperK_HybridmPMT_WithOD_Realistic_Geometry()
   WCODDeadSpace            = 600.*mm;
   WCODTyvekSheetThickness  = 1.*mm; // Quite standard I guess
   WCODWLSPlatesLength      = 30.*cm; //
+  G4String WLSType="";
+  if( HKFD_OD_WLS_PLATE_TYPE == "EljenEJ286" ){
+	WLSType = "EljenEJ286";
+	WCODWLSPlatesThickness   = 1.*cm; //
+  }else if( HKFD_OD_WLS_PLATE_TYPE == "Kuraray" ){
+	WLSType = "Kuraray";
+	WCODWLSPlatesThickness   = 0.6*cm; //
+  }  else if( HKFD_OD_WLS_PLATE_TYPE == "Inr" ){
+	WLSType = "Inr";
+	WCODWLSPlatesThickness   = 0.7*cm; //
+  }
   WCODDiameter             = WCIDDiameter + 2*(WCBlackSheetThickness+WCODDeadSpace+WCODTyvekSheetThickness+WCODWLSPlatesThickness);
 
   // OD PMTs //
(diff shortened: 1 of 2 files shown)


src/WCSimDetectorConfigs.cc
```diff
@@ -1012,12 +1011,16 @@ void WCSimDetectorConstruction::SetHyperK_HybridmPMT_WithOD_Realistic_Geometry()
...
-  else if( HKFD_OD_WLS_PLATE_TYPE == "Kuraray" )
+	WCODWLSPlatesThickness   = 1.*cm; //
+  }else if( HKFD_OD_WLS_PLATE_TYPE == "Kuraray" ){
 	WLSType = "Kuraray";
-  else if( HKFD_OD_WLS_PLATE_TYPE == "Inr" )
+	WCODWLSPlatesThickness   = 0.6*cm; //
+  }  else if( HKFD_OD_WLS_PLATE_TYPE == "Inr" ){
 	WLSType = "Inr";
+	WCODWLSPlatesThickness   = 0.7*cm; //
+  }
```
tdealtry: `WCODWLSPlatesThickness` is used to calculate `WCODDiameter`, but isn't set until after it's used

Basically, need to move this block up to line 981 again

src/WCSimDetectorConfigs.cc
```diff
@@ -1131,12 +1133,16 @@ void WCSimDetectorConstruction::SetHyperK_HybridmPMT_IDonly_Realistic_Geometry()
...
-  else if( HKFD_OD_WLS_PLATE_TYPE == "Kuraray" )
+	WCODWLSPlatesThickness   = 1.*cm; //
+  }else if( HKFD_OD_WLS_PLATE_TYPE == "Kuraray" ){
 	WLSType = "Kuraray";
-  else if( HKFD_OD_WLS_PLATE_TYPE == "Inr" )
+	WCODWLSPlatesThickness   = 0.6*cm; //
+  }else if( HKFD_OD_WLS_PLATE_TYPE == "Inr" ){
 	WLSType = "Inr";
+	WCODWLSPlatesThickness   = 0.7*cm; //
+  }
```
tdealtry: Same comment

tdealtry (changes requested):
Thanks @federiconova 

Found a subtle bug

Also, please add comments to Build.cc, something like

Under the IWCD bit
```
All IWCD WLS plates are currently set to 1 cm by default. These defaults can be overridden with /WCSim/HyperKOD/ODWLSPlatesThickness
```

And under the HKFD bit
```
HKFD WLS plate thickness is automatically set based on the chosen plate type. This cannot be overridden when using HyperK_HybridmPMT_WithOD_Realistic & HyperK_HybridmPMT_NoOD_Realistic
```

federiconova:
> Thanks @federiconova
> 
> Found a subtle bug
> 
> Also, please add comments to Build.cc, something like
> 
> Under the IWCD bit
> 
> ```
> All IWCD WLS plates are currently set to 1 cm by default. These defaults can be overridden with /WCSim/HyperKOD/ODWLSPlatesThickness
> ```
> 
> And under the HKFD bit
> 
> ```
> HKFD WLS plate thickness is automatically set based on the chosen plate type. This cannot be overridden when using HyperK_HybridmPMT_WithOD_Realistic & HyperK_HybridmPMT_NoOD_Realistic
> ```

ok done. but I removed the "cannot be overridden" part, because the thickness still can be overwritten using mac instructions (for both IWCD and FDOD)

src/WCSimDetectorConfigs.cc
federiconova: ah very good catch

src/WCSimDetectorConfigs.cc
federiconova: done

tdealtry:
Looks good to me. Just the usual question [on the CI](https://wcsim.github.io/Validation/3a1c0750d087218ed768ba122b8e92a8e9f1787a/index.html): are the up to ~3% changes in the number of OD raw hits expected for this? I guess the change is 1cm -> 0.7cm on this branch

federiconova:
> Looks good to me. Just the usual question [on the CI](https://wcsim.github.io/Validation/3a1c0750d087218ed768ba122b8e92a8e9f1787a/index.html): are the up to ~3% changes in the number of OD raw hits expected for this? I guess the change is 1cm -> 0.7cm on this branch

yes. The change should be small because the absorption length is smaller than 7 mm anyway in the UV, and much larger in the visible.

tdealtry (approved):
Great thanks

PR #547, opened 2025-11-26
Fixes photon timing and theta binning.

ekneale:
Updates photon time sampling to sample from Gaussian distribution with FWHM pulsewidth.

Also fixes theta binning so that the number of bins is not dependent on whether the LI is in the OD or ID, since this resulted in extremely long sampling times for the OD.


ekneale:
In fact, I have noticed that the changes made to the sampling for the diffusers won't work for the collimator sims. These involve the theta and phi binning and so I will push the changes to this PR since they are related changes and override the current change to the theta binning.

ekneale:
New sampling method added to deal with profiles measured in angle from the collimator axis in x and y. Seems to be no need for the extra step of projecting onto the sphere. Sampling from x and y direct gives the image below for a collimator simulated in the ID at (3600,0,0) pointing in the (-1,0,0) direction. The distance from the xy plane of ~300 cm at a distance of 70 m from the collimator corresponds to a half angle of ~ 2.5 degrees .

 image

ekneale:
Marking as ready for review following validation and debugging by @ekneale and @menailj.

menailj:
Do all collimator scans use the xyAngle method, and all diffuser scans use the thetaPhi method? If so, it might be easier not to have the extra command option in the mac file to avoid confusion

ekneale:
Unfortunately not, as some of the diffuser scans will have the new format.

tdealtry (approved):
Thank you very much for this @ekneale @menailj

PR #548, opened 2025-12-17
Addition of TOF (T5) scintillator to beam pipe

jerenner:
Places the TOF (T5) scintillator in the WCTE beam pipe

The TOF (T5) scintillator consists of 8 scintillator bars stacked in the y-dimension with varying lengths in the x-dimension:
- 6.4 mm thickness in Z-dimension
- 16.25 mm height in Y-dimension
- lengths of {41, 94, 112, 123, 123, 112, 94, 41} mm in X-dimension

(Code written by Google Gemini 3 and subsequently revised)

### DIFF
--- a/src/WCSimConstructBeamPipe.cc
+++ b/src/WCSimConstructBeamPipe.cc
@@ -108,58 +108,150 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructBeamPipe()
   //////////////////////////////////////////////////
   /// 3) T5 TOF Scintillator                     ///
   //////////////////////////////////////////////////
-
-  // Get EJ-228 Material (Polyvinyltoluene based)
-  G4NistManager* nist = G4NistManager::Instance();
-  G4Material* scintMaterial = nist->FindOrBuildMaterial("G4_PLASTIC_SC_VINYLTOLUENE");
   
-  // Dimensions
+  // -- Materials --
+  G4NistManager* nist = G4NistManager::Instance();
+  G4Material* scintMaterial = nist->FindOrBuildMaterial("G4_PLASTIC_SC_VINYLTOLUENE"); // EJ-228 equivalent
+  G4Material* vinylMaterial = nist->FindOrBuildMaterial("G4_POLYVINYL_CHLORIDE"); // PVC
+  G4Material* mylarMaterial = nist->FindOrBuildMaterial("G4_MYLAR");
+  G4Material* airMaterial   = G4Material::GetMaterial("Air1"); // For the gaps inside the assembly
+
+  // -- Dummy materials properties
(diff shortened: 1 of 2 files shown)


jerenner:
Initial placement of scintillators:

T5_beam_pipe
T5_beam_pipe_zoom
T5_beam_pipe_zoom2

src/WCSimConstructBeamPipe.cc
```diff
@@ -103,8 +105,66 @@ G4LogicalVolume* WCSimDetectorConstruction::ConstructBeamPipe()
                     0,
                     checkOverlaps);
 
+  //////////////////////////////////////////////////
+  /// 3) T5 TOF Scintillator                     ///
+  //////////////////////////////////////////////////
+
+  // Get EJ-228 Material (Polyvinyltoluene based)
+  G4NistManager* nist = G4NistManager::Instance();
+  G4Material* scintMaterial = nist->FindOrBuildMaterial("G4_PLASTIC_SC_VINYLTOLUENE");
```
kmtsui: There are errors from validation CI.
```
> "Optical photon is killed because of missing refractive index" 2
```
This is very probably because optical properties (refractive index, etc.) are not defined for this `scintMaterial`.
Since it is placed into the beam pipe, those photons would not escape into the detector, so it's fine to define some dummy properties like [here](https://github.com/WCSim/WCSim/blob/43b3dae41cdc6cdb6676956d44ae474b2fb88959/src/WCSimConstructMaterials.cc#L1771). And also move the material definition to the same place.

kmtsui: And are we happy with the material density, etc.? At least I want to make sure it does not introduce too much interaction material that would dissipates too much energy or produces too much secondaries.

kmtsui:
Just add the detector display instructions for reference
```
Ok I've added the detector display figures to the pull request. To make these, I used the VRML export option (after fighting for awhile trying to get OGL to work through the Docker container and not succeeding):

/vis/open VRML2FILE
/vis/drawVolume
/vis/viewer/refresh

The resulting .wrl file took awhile to export and was quite large (~460MB), but it worked, and I was able to view it using this program: https://castle-engine.io/castle-model-viewer
```

kmtsui:
Two requests here:
1. can you update the detector displays, with the Mylar and vinyl layers?
2. can you move the material definition to src/WCSimConstructMaterials.cc?

jerenner:
Updated displays of T5 scintillator:

beam_pipe_T5_0
beam_pipe_T5_2
beam_pipe_T5_1

kmtsui:
@jerenner thanks! I am going to merge it, unless there are additional components to be added within the beam pipe

PR #549, opened 2025-12-18
NiCf Simulation Partial Integration

DiegoCostas97:
Neutron generation from Cf Spontaneous Fission must be simulated from standalone Geant4 application since Geant4.10 does not have SF implemented.

I have a Geant4.11 application for this. It could be uploaded somehow too.

### DIFF
--- a/src/WCSimPrimaryGeneratorAction.cc
+++ b/src/WCSimPrimaryGeneratorAction.cc
@@ -9,6 +9,7 @@
 #endif
 
 #include "WCSimDetectorConstruction.hh"
+#include "WCSimDetectorMessenger.hh"
 #include "WCSimPrimaryGeneratorMessenger.hh"
 #include "G4RunManager.hh"
 #include "G4Event.hh"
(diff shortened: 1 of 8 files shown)


kmtsui:
The Ni ball construction looks fine. Can you show some detector displays and do some simulations with, e.g. geantino to make sure the placement is in the desired position?

kmtsui:
About the particle generation, it is a bit tricky because we need some external simulation to provide the particle spectra. 
If you are the only one doing NiCf simulation, then it might be fine. Otherwise it would be better if you have a stable storage for such input file (with large enough amount of events for large statistics simulation)

kmtsui:
The `custom` event generator is basically generating particles from a single vertex. I suggest you make your input into [Nuance_MC_Format](https://github.com/WCSim/WCSim/blob/develop/doc/Nuance_MC_Format.md) instead of making a new generator.

kmtsui:
Add the example usage into `macros/WCTE.mac`

src/WCSimPrimaryGeneratorAction.cc
tdealtry, line 1: I totally agree with @kmtsui that using the nuance format will be a good idea (this will mean you don't need any changes to the primary generator action or messenger)

But just to say that if you do want to use a new format you've defined for whatever reason, then "custom" is a bad name.

PR #554, opened 2026-02-23
Adding compatibility with Geant4 version > 11.0.0

nadrino:
This PR is a collection of commits I've implemented to get the code running with the latest Geant4. A few deprecated methods were replaced accordingly.

This PR also preserves the original implementation for version below 11.0.0. This is done thanks to pre-processor macros that redirect the compiler depending on the G4 version it's compiling in with.

### DIFF
--- a/include/G4Utils.hh
+++ b/include/G4Utils.hh
@@ -1,0 +1,15 @@
+//
+// Created by Nadrino on 23/02/2026.
+//
+
+#ifndef WCSIM_G4UTILS_H
+#define WCSIM_G4UTILS_H
+
+#include "G4Version.hh"
+
+#define G4VERSION_ENCODE(major,minor,patch) ((major)*100 + (minor)*10 + (patch))
+
+#define G4VERSION_IS_GREATER_EQUAL(major,minor,patch) (G4VERSION_NUMBER >= G4VERSION_ENCODE(major,minor,patch))
+
+
+#endif //WCSIM_G4UTILS_H
--- a/include/G4Utils.hh
+++ b/include/G4Utils.hh
@@ -6,10 +6,24 @@
 #define WCSIM_G4UTILS_H
 
 #include "G4Version.hh"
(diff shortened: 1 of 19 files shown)


kmtsui:
Thanks for the hard work! It is a good sign that the CI with older geant4 runs perfectly.
I remember there were discussions of migration to newer geant4 for the hyper-k software package. I think it is the best if someone (@tdealtry ?) can test this PR with the new software environment.

nadrino:
Thank you very much @kmtsui. Is this repo still up to date or is it being deprecated?

kmtsui:
Yes we are still developing it for WCTE and Hyper-K MC production. 
Of course Hyper-K is moving towards GHOST but I think it will take some time so WCSim will still be maintained for a while.

include/G4Utils.hh
tdealtry, line 1: For consistency, could you rename this `WCSimG4Utils.hh`?

tdealtry (changes requested):
Thanks for this and sorry for the delay (I've been away). This looks good to me. Just the minor comment inline, and a request to please pull from the develop branch, just so I can check the CI works for everything (my assumption is that the 2 failures were that you worked from an older version of the code base at the time)

vish11rs:
Hi Adrien,
Thanks for the Geant4 11 compatibility updates. I tested your branch with Geant4 11.4.0, and WCSim now builds successfully.

While testing, I ran into another issue when running:
bin/WCSim macros/WCSim_1part_HK.mac macros/tuning_parameters_hkfd.mac
WCSim aborts during material construction with:
G4Exception : mat205
Attempting to create a new material property vector FASTCOMPONENT
The exception seems to come from WCSimConstructMaterials.cc when constructing the BGO material properties table (BGO_mpt->AddProperty("FASTCOMPONENT", ...)).

I'm just starting to work with WCSim, so I may be missing something. Has this already been fixed somewhere, or is it another Geant4 11 compatibility issue?

PR #556, opened 2026-04-16
Save all Michel electrons from the muMinusCaptureAtRest process

tdealtry:
The current default is to only save Michel electrons from mu- in the WCSimTracks output container, if the Michel electron goes on to create Cherenkov photons that hit PMTs (unless they have the `Decay` process, which is not prevelant for mu-. It dominates for mu+). There are therefore some edge cases where the Michel electron isn't saved for mu-.

There has been a request from the LBL group to always save the produced Michel electron. This will catch the edge cases e.g. where the Michel is created & remains in the deadspace. This solves that

### DIFF
--- a/src/WCSimTrackingAction.cc
+++ b/src/WCSimTrackingAction.cc
@@ -19,8 +19,7 @@ WCSimTrackingAction::WCSimTrackingAction()
   ProcessList.insert("Decay") ;                         // Michel e- from pi+ and mu+
   ProcessList.insert("conv") ;                         // Products of gamma conversion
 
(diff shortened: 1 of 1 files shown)


nickwp:
Note that I think this behaviour can already be enabled by [adding the following command](https://github.com/WCSim/WCSim/blob/ed6c5f5f7f1d1f018f1336b7d3f83eb0a31aa18b/src/WCSimTrackingMessenger.cc#L28-L30) to the mac file:

```
/Tracking/trackProcess muMinusCaptureAtRest
```

If this doesn't work, that's probably a bug that should be fixed!

(I'm not objecting to also changing the defaults through this PR, so that this process is _always_ tracked, since it would make sense as the Decay process is always tracked already)

tdealtry:
Thanks for the comment @nickwp. Yeah I agree it's just a little niceity, but there's a new release for the MC production, so it's a good time to get this in now (people using older versions of the code, use Nick's mac command!)

Merging despite CI failing
- WCTE CI failing since a geometry update earlier this year
- CI with mu minuses are failing because more true tracks are being saved. That's the whole point of this PR

PR #557, opened 2026-04-20
Feature/individual pmt dark rate

federiconova:
allow to set different average dark rate levels for ID PMTs
levels are read in from a text file (similar to what is done in IWCD geometry to set indivudal PMT positions)
only works for realistic geometry


src/WCSimDetectorConfigs.cc
```diff
@@ -1024,6 +1024,7 @@ void WCSimDetectorConstruction::SetHyperK_HybridmPMT_WithOD_Realistic_Geometry()
   BuildODWLSCladding = true;
   CreateWLSObject(WLSType);
   isRealisticPlacement=true;
+  SetPMTDarkRateInput(wcsimdir_path+"/data/PMT_darkrate.txt");
```
tdealtry: My feeling is that it would be easier for users if this was a Messenger option

Concrete use case why this would help? Person wants to test `n` different dark rate files on a batch system. The soft link case means this needs to be done one after the other, or they need to clone & compile multiple copies of WCSim

Another benefit would be different PMT types can have different dark noises

doc/PMTDarkRateTableFormat.md
tdealtry, line 1: Thank you for documenting this format!

include/WCSimDetectorConstruction.hh
```diff
@@ -135,6 +135,10 @@ public:
   G4int    GetTotalNumPmts2() {return totalNumPMTs2;}//For the hybrid config
   G4int    GetTotalNum_mPmts2() {return totalNum_mPMTs2;}//For the hybrid config         
   G4int    GetTotalNumODPmts() {return totalNumODPMTs;}
+  G4bool   GetreadDarkRateFromTable() {return readDarkRateFromTable;}
+  G4int    GetnPMTsReadForDarkRate() {return nPMTsReadForDarkRate;}
+  std::vector<G4int> GetpmtId() {return pmtId;}
+  std::vector<G4double> GetpmtDarkRate() {return pmtDarkRate;}
```
tdealtry:
```suggestion
  G4bool   GetReadDarkRateFromTable() {return readDarkRateFromTable;}
  G4int    GetNPMTsReadForDarkRate() {return nPMTsReadForDarkRate;}
  std::vector<G4int> GetpmtId() {return pmtId;}
  std::vector<G4double> GetPMTDarkRate() {return pmtDarkRate;}
```
I'm sorry, I know you're using the same capitalisation as the variables themselves, but having them start with lower case after the `Get` is unpleasantly unreadable and inconsistent with other methods

Alternatively could do a `Get_pmtDarkRate` etc

tdealtry (changes requested):
Thanks for this @federiconova! This has been a missing feature for a long time

A couple of minor comments inline. A more major/conceptual one is mixed up with where the information is read in
- I've not run the code, but I presume that the dark rates for mPMTs & OD PMTs will be set to the same value as for the 20" PMTs (e.g. everything @ 10k in the example you added to `data/`) It'd be good if they could be set to unique values instead. Dark noise creation is expensive computationally (e.g. disc space), so if you're creating a lot of mPMT & OD noise (that you can't even use because they're too noisy), it's inefficient
- This reads dark noise options in via WCSimDetectorMessenger (and I think only works for one geometry due to the filename setting in the DetectorConfigs file?). Other dark noise options are read in using WCSimDarkRateMessenger, and so will confuse users & developers
- Actually if you do read in the options using WCSimDarkRateMessenger, and since each PMT type has its own copy of WCSimWCAddDarkNoise, can actually use 3 different files for the 3 different PMT types rather easily (and work for all geometries)
- So I'd suggest taking a look at porting your changes from the detector/geometry classes & porting it into the dark noise classes. If I've missed something and this won't work nicely, then happy for you to ignore this suggestion. But I think the rewrite shouldn't take too long (an hour at most?) and will give more flexibility to the feature

One more request. If you have a script to create your dark noise files, then it'd be nice to add to `sample-root-scripts` to make it easier for people to get started with this new feature

FYI the [CI is failing for the FDOD geometry](https://github.com/WCSim/WCSim/actions/runs/24659521396/job/72102351942?pr=557#step:12:7773). I think it's because the new files in data need adding to [here in `CMakeLists.txt`](https://github.com/WCSim/WCSim/blob/develop/CMakeLists.txt#L180)

include/WCSimDetectorConstruction.hh
federiconova: fixed

src/WCSimDetectorConfigs.cc
federiconova: I see your point about parallel batch jobs. I was basically using as a template what the IWCD group has done. They pass to WCSim a similar txt file with the list of PMT ids and positions. They call it like this:
  SetPMTPositionInput(wcsimdir_path+"/data/mPMT_Position_IWCD.txt");
So I was just replicating this method. Anyway, it is possible to overwrite the name of the file via the mac file. See the method SetPMTDarkRateInput in DetectorMessenger. The macfile line is:

/WCSim/PMT/DarkRateFile data/PMT_darkrate.txt

So I think it's fine?

About the different PMT types, this is not an issue. Different files can be separately provided for the different PMT types. Again this matches what the IWCD group does, for example they have:
  SetODPMTPositionInput(wcsimdir_path+"/data/mPMT_Position_IWCD_OD.txt");

federiconova:
Hello, let me answer inline to these points.

> A couple of minor comments inline. A more major/conceptual one is mixed up with where the information is read in
> 
>     * I've not run the code, but I presume that the dark rates for mPMTs & OD PMTs will be set to the same value as for the 20" PMTs (e.g. everything @ 10k in the example you added to `data/`) It'd be good if they could be set to unique values instead. Dark noise creation is expensive computationally (e.g. disc space), so if you're creating a lot of mPMT & OD noise (that you can't even use because they're too noisy), it's inefficient

no, the value from the text file is only set for ID 20 inch PMTs, because of this line in WCSimWCAddDarkNoise:

    if( myDetector->Get_readDarkRateFromTable() && detectorElement=="tank" ){

So mPMTs and OD PMTs are not touched. It would be easy to extend the method and provide a separate text file for these other PMT types.

> 
>     * This reads dark noise options in via WCSimDetectorMessenger (and I think only works for one geometry due to the filename setting in the DetectorConfigs file?). Other dark noise options are read in using WCSimDarkRateMessenger, and so will confuse users & developers
> 

Yes currently this only works for the Realistic geometry. Yes the options are set via DetectorMessenger. I have tried to change this and set the options via DarkRateMessenger, and went some way, but I got stuck at the following problem. Currently, in WCSimConstructGeometryTables there is a call to:
ReadDarkRateTableFromFile()
This is only executed once and the values are read in, in the same way as the ReadGeometryTableFromFile() method for the IWCD simulations. This block of code, however, belongs to DetectorConstruction and so is driven by DetectorMessenger. So, I have tried to move lines from DetectorMessenger to DarkRateMessenger, but in this case where should I move the ReadDarkRateTableFromFile() method? I think it should not go into WCAddDarkNoise, because then it would be executed at each event rather than only once at the beginning?

>     * Actually if you do read in the options using WCSimDarkRateMessenger, and since each PMT type has its own copy of WCSimWCAddDarkNoise, can actually use 3 different files for the 3 different PMT types rather easily (and work for all geometries)

I can easily extend the method to assign dark rate values to the 3 PMT types separately via different txt files. I will do it once we agree in which Messenger the stuff should stay.


>     * So I'd suggest taking a look at porting your changes from the detector/geometry classes & porting it into the dark noise classes. If I've missed something and this won't work nicely, then happy for you to ignore this suggestion. But I think the rewrite shouldn't take too long (an hour at most?) and will give more flexibility to the feature
> 
> 
> One more request. If you have a script to create your dark noise files, then it'd be nice to add to `sample-root-scripts` to make it easier for people to get started with this new feature

I don't really have a script for this. In fact one problem I have is that, in the input text file, there need to be as many lines as there are ID PMTs, but this number needs to be magically known to the user as the PMTs are then separately populated in detector construction. This is different from what the IWCD group does, as they provide a txt file which also tells the code how many PMTs to build and where to put them.

> FYI the [CI is failing for the FDOD geometry](https://github.com/WCSim/WCSim/actions/runs/24659521396/job/72102351942?pr=557#step:12:7773). I think it's because the new files in data need adding to [here in `CMakeLists.txt`](https://github.com/WCSim/WCSim/blob/develop/CMakeLists.txt#L180)

added

tdealtry:
Thank you for looking into the option to move things into the dark rate messenger. I had missed that complication. My feeling is that we don't spend a lot of time on this (given the long term prospects of GHOST), but just do something to make the initial use case possible (e.g. if people want to modify dark rate with a txt file for multiple PMT types at once, then we cross that bridge later)

What the issue is that needs to be fixed before this can be merged is that currently this MR forces users to use the `.txt` file format for 20" PMTs inside the HKFD realistic geometry, in order to change dark rate
- This turns dark noise off
```
/DarkRate/SetDetectorElement tank
/DarkRate/SetDarkRate 0 MHz
```
- This runs the values in `data/PMT_darkrate.txt`, no matter what the non-zero value is set to (e.g. 0.1 kHz, 1 MHz) and is the same if the `SetDarkRate` line is commented out
```
/DarkRate/SetDetectorElement tank
/DarkRate/SetDarkRate 4.2 kHz
```

How to proceed?
- I've had a quick play running, and it seems like removing `SetPMTDarkRateInput()` in `WCSimDetectorConfigs.cc` is all that's required. This seems to allow the `/DarkRate/SetDarkRate` option to work as intended again
- One can then use `/WCSim/PMT/DarkRateFile` before `/WCSim/Construct` to set a custom PMT dark rate file for the "tank" PMTs (20" in HKFD - works, mPMT in IWCD - untested)
- I think it'd be good to add some description in the .mac file you've added (rather than WCSim.mac, because it is confusing) along the lines of
```
# Use the /WCSim/PMT/DarkRateFile command to specify a text file containing the dark rate for each PMT type in the "tank" (e.g. 20" in HKFD)
# When using this, /DarkRate/SetDarkRate will not work for the "tank" element, other than setting it to 0
# e.g. if you specify /DarkRate/SetDarkRate 0, dark noise is turned off for all tank PMTs, independent of values in your DarkRateFile
# e.g. if you specify /DarkRate/SetDarkRate 0.1 Hz or /DarkRate/SetDarkRate 1 MHz (or leave it uncommented), these values will be ignored & dark noise will be added according to your DarkRateFile
# This command *must* come between your /WCSim/WCgeom command and /WCSim/Construct
```

federiconova:
> Thank you for looking into the option to move things into the dark rate messenger. I had missed that complication. My feeling is that we don't spend a lot of time on this (given the long term prospects of GHOST), but just do something to make the initial use case possible (e.g. if people want to modify dark rate with a txt file for multiple PMT types at once, then we cross that bridge later)
> 
> What the issue is that needs to be fixed before this can be merged is that currently this MR forces users to use the `.txt` file format for 20" PMTs inside the HKFD realistic geometry, in order to change dark rate
> 
>     * This turns dark noise off
> 
> 
> ```
> /DarkRate/SetDetectorElement tank
> /DarkRate/SetDarkRate 0 MHz
> ```
> 
>     * This runs the values in `data/PMT_darkrate.txt`, no matter what the non-zero value is set to (e.g. 0.1 kHz, 1 MHz) and is the same if the `SetDarkRate` line is commented out
> 
> 
> ```
> /DarkRate/SetDetectorElement tank
> /DarkRate/SetDarkRate 4.2 kHz
> ```
> 
> How to proceed?
> 
>     * I've had a quick play running, and it seems like removing `SetPMTDarkRateInput()` in `WCSimDetectorConfigs.cc` is all that's required. This seems to allow the `/DarkRate/SetDarkRate` option to work as intended again
> 
>     * One can then use `/WCSim/PMT/DarkRateFile` before `/WCSim/Construct` to set a custom PMT dark rate file for the "tank" PMTs (20" in HKFD - works, mPMT in IWCD - untested)
> 
>     * I think it'd be good to add some description in the .mac file you've added (rather than WCSim.mac, because it is confusing) along the lines of
> 
> 
> ```
> # Use the /WCSim/PMT/DarkRateFile command to specify a text file containing the dark rate for each PMT type in the "tank" (e.g. 20" in HKFD)
> # When using this, /DarkRate/SetDarkRate will not work for the "tank" element, other than setting it to 0
> # e.g. if you specify /DarkRate/SetDarkRate 0, dark noise is turned off for all tank PMTs, independent of values in your DarkRateFile
> # e.g. if you specify /DarkRate/SetDarkRate 0.1 Hz or /DarkRate/SetDarkRate 1 MHz (or leave it uncommented), these values will be ignored & dark noise will be added according to your DarkRateFile
> # This command *must* come between your /WCSim/WCgeom command and /WCSim/Construct
> ```

federiconova:
> > Thank you for looking into the option to move things into the dark rate messenger. I had missed that complication. My feeling is that we don't spend a lot of time on this (given the long term prospects of GHOST), but just do something to make the initial use case possible (e.g. if people want to modify dark rate with a txt file for multiple PMT types at once, then we cross that bridge later)
> > What the issue is that needs to be fixed before this can be merged is that currently this MR forces users to use the `.txt` file format for 20" PMTs inside the HKFD realistic geometry, in order to change dark rate
> > ```
> > * This turns dark noise off
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > /DarkRate/SetDetectorElement tank
> > /DarkRate/SetDarkRate 0 MHz
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > * This runs the values in `data/PMT_darkrate.txt`, no matter what the non-zero value is set to (e.g. 0.1 kHz, 1 MHz) and is the same if the `SetDarkRate` line is commented out
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > /DarkRate/SetDetectorElement tank
> > /DarkRate/SetDarkRate 4.2 kHz
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > How to proceed?
> > ```
> > * I've had a quick play running, and it seems like removing `SetPMTDarkRateInput()` in `WCSimDetectorConfigs.cc` is all that's required. This seems to allow the `/DarkRate/SetDarkRate` option to work as intended again
> > 
> > * One can then use `/WCSim/PMT/DarkRateFile` before `/WCSim/Construct` to set a custom PMT dark rate file for the "tank" PMTs (20" in HKFD - works, mPMT in IWCD - untested)
> > 
> > * I think it'd be good to add some description in the .mac file you've added (rather than WCSim.mac, because it is confusing) along the lines of
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > # Use the /WCSim/PMT/DarkRateFile command to specify a text file containing the dark rate for each PMT type in the "tank" (e.g. 20" in HKFD)
> > # When using this, /DarkRate/SetDarkRate will not work for the "tank" element, other than setting it to 0
> > # e.g. if you specify /DarkRate/SetDarkRate 0, dark noise is turned off for all tank PMTs, independent of values in your DarkRateFile
> > # e.g. if you specify /DarkRate/SetDarkRate 0.1 Hz or /DarkRate/SetDarkRate 1 MHz (or leave it uncommented), these values will be ignored & dark noise will be added according to your DarkRateFile
> > # This command *must* come between your /WCSim/WCgeom command and /WCSim/Construct
> > ```

This was a good suggestion thanks. I've implemented it. Let's see what the CI says.

tdealtry (approved):
Thanks for this @federiconova. The CI failure smells of this branch needing updating to develop (to pickup these changes)
- more tracks now saved, due to muoncaptureatrest process tracks being saved by default
- small changes to IWCD+OD geometry results due to updated geometry

When the CI passes, I'll merge this

federiconova:
> Thanks for this @federiconova. The CI failure smells of this branch needing updating to develop (to pickup these changes)
> 
>     * more tracks now saved, due to muoncaptureatrest process tracks being saved by default
> 
>     * small changes to IWCD+OD geometry results due to updated geometry
> 
> 
> When the CI passes, I'll merge this

I've updated, please let me know if it's ok now or something else is wrong...
Thanks

tdealtry:
Excellent. Many thanks for this @federiconova

federiconova:
> Excellent. Many thanks for this @federiconova

Great many thanks for merging this in. I will have some more updates about this (I hope to talk about it at the DAQ meeting on Friday) but for now this is very good progress and we can have an extra incremental pull request later.
Thanks!

PR #558, opened 2026-05-05
separate tyvek reflectivities

federiconova:
This allows setting separate tyvek reflectivities for the OD in-wall and out-wall.
It is a feature long time needed for OD simulations as the baseline values are very different (95% inwall, 87% outwall).
I've tried to set it up so that IWCD simulations aren't changed.


src/WCSimTuningParameters.cc
```diff
@@ -32,7 +32,8 @@ WCSimTuningParameters::WCSimTuningParameters()
  topveto = false;
 
  WCODWLSCladdingReflectivity   = 0.99; //
- WCODTyvekReflectivity   = 0.90; //
+ WCODTyvekReflectivity   = 0.95; //
+ WCODTyvekOutwallReflectivity   = 0.87; //
```
tdealtry:
```suggestion
 WCODTyvekReflectivity   = 0.90; //
 WCODTyvekOutwallReflectivity   = 0.90; //
```

The CI is failing. I see two possibilities
1. Your branch is a bit outdated. Please pull from develop (but the changes from the commits you don't have I don't think actually change the physics that runs in the CI)
2. More likely the culprit is these two lines. The CI has its own [tuning_parameters.mac](https://github.com/WCSim/Validation/blob/master/Generate/macReference/tune/tuning_parameters.mac) (used for both IWCD & HKFD). Which means changing the defaults here will modify the CI output (the new Outwall will be 0.87 instead of 0.9).


To ensure this PR is adding the extra surface/properties correctly, I'd like a CI run with the old default values (so please accept this suggestion). This will ensure no bugs have been introduced

Then when that passes, you can update the default values like this (again, sorry). The CI will fail, by design

I'll then modify the validation CI mac file to set the new `WCODTyvekOutwallReflectivity` explicitly, (which will allow default values to change without making the CI fail in future)

Then I can merge this PR

tdealtry (changes requested):
Thanks for this @federiconova 

You've been inconsistent between using `Inwall` vs `<nothing>` in variable names for the inner wall (`Outwall` seems to be consistently used). For clarity, it'd be good to use `Inwall` consistently. Hopefully can do this with a handful of clever find-replace commands (if not, please comment the header files to say explicitly that these parameters are for the OD inner wall)

Please also modify `tuning_parameters_iwcd.mac` to set the outer wall reflectivity to 0.9

Otherwise this looks good. Other than the CI... but hope we can sort this in the right order (see comment inline)

src/WCSimTuningParameters.cc
federiconova: Hi, I think I implemented all the required changes:
- explicitely say Inwall and Outwall everywhere; 
- update IWCD code accordingly; 
- add config line to IWCD macfile; 
- set default values back to 90%
Please take a look and let's see what the CI says

tdealtry: Thanks @federiconova. Again updated this branch so we have a more accurate CI output for comparison

federiconova: updated, let me know if it's ok now or what is wrong please...

tdealtry: So now the only change is in the FD OD (rather than also in geometries without an OD). And only for our OD-starting particle guns (which we know are of limitted use). For reference, this PR has [dropped the number of OD hits by 10-16%](https://wcsim.github.io/Validation/44aca2687631b9df467c82e82e84c07f20dffd2d/index.html)

As a last hurrah, I'm going to swap the defaults in TuningMessenger to 0.9 & see what happens

If this doesn't work, I'd suggest one of
- You posting a link to some slides presented at an OD meeting to show this is working as you expect
- An explanation as to why the CI is turning out different (I can't see why it should)

federiconova: Hi many thanks for checking
Yes there is a very good reason why things change in the OD, even though the parameters in the mac file are both set at 0.90 which is what the single parameter was before.
The underlying reason is that in the main branch a single TYVEK model is used for both in-wall and out-wall. In this pull request, instead, two different TYVEK materials are defined, having different reflectivity spectra, for the two walls. The lines in the mac file simply scale the spectra up or down setting the maximum value at 0.90. But the shape of the spectra is still different. This is correct because different TYVEK types are actually used in the in-wall and out-wall. I am attaching some slides from today's OD meeting describing this.
[2026_06_02_OD_tyveks.pdf](https://github.com/user-attachments/files/28519373/2026_06_02_OD_tyveks.pdf)

It should be added that I have put both values to 0.90 (as per your request) to sort of minimize the change, but in reality the two values should also be different: at 0.95 for the in-wall and 0.87 for the out-wall. So, two different spectra, with two different maxima. You suggested to set the values at 0.90 for now, merge it in, and then have a separate pull request to change the values to 0.95 and 0.87. I am ok with proceeding this way. Hope it makes sense!

tdealtry: Ahh yes of course. This makes sense. Thank you. I'll merge now

tdealtry (approved):

PR #559, opened 2026-05-12
changed OD related size

ryom2020:
changed WLS plate size and type, OD size and blacksheet thickness in IWCD_mPMT_WithOD


tdealtry (approved):
Thanks @ryom2020. If the IWCD group are happy that this is a more realistic geometry & the results look ok, then I'll merge & issue a new release (I'll be at the meeting in 40 minutes so can hopefully finalise there)

PR #560, opened 2026-05-19
Standardise sampling method for li sources

menailj:
Sampling method TSpline3 vs Monotonic used for collimators and diffusers. No reason they should differ. 
Also TSpline3 method breaks when running interactively. 
Restructured code to share function related to setting up splines.

Also renamed one of the OD collimator profiles to ID to allow code to work for both types.


tdealtry:
Thanks @menailj. Looks good to me. I presume you've presented results in the calibration group & they're happy that the output is as expected?

menailj:
Yes I discussed mainly with Liz but also the LI group and there was no issue with this method.

PR #562, opened 2026-05-20
Update radioactivity numbers

menailj:
Update the numbers for radioactivity to better reflect real measured results of HK PMTs.
Numbers based on discussion with Yano-san, and measurements as of 12/25.

### DIFF
--- a/WCSim.mac
+++ b/WCSim.mac
@@ -401,10 +401,11 @@
 #/mygen/radioactive_time_window 1.e-4
 ## The following list the different isotope to take into account, where they decay,
 ## and what are their activities in the full ID detector (for Water) or in the PMT (for PMT)
-#/mygen/isotope Tl208 PMT 2.8
-#/mygen/isotope Bi214 PMT 27.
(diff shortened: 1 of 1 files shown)


tdealtry:
Thanks for this @menailj. If you can link a HK note or meeting planner link in the mac file for future reference, that'd be great (but if this is just from discussions/emails, then obviously not. Let me know)

tdealtry:
I wonder if it it'd also be possible to modify the default value of this variable too? @takatomiyano @gpronost
```
/mygen/radon_water_concentration
```

gpronost:
Maybe yes, default water concentration for HK should be 50, for SK it is 2.63 (current value)

menailj:
> Thanks for this @menailj. If you can link a HK note or meeting planner link in the mac file for future reference, that'd be great (but if this is just from discussions/emails, then obviously not. Let me know)

There are no references for this other than an email chain. This should be updated with reference from FD1 group once they are happy all measurements are completed.

PR #563, opened 2026-06-04
set default reflectivity values for FDOD TYVEK

federiconova:
Just a tiny incremental pull request to set the default values for the reflectivity of the FDOD to:
95% (in-wall)
87% (out-wall)


tdealtry:
Thanks for this @federiconova 

This modifies the HKFD output (as expected) in the CI jobs that induce OD hits

It also modifies the IWCD output in the CI jobs that induce OD hits. This confused me at first, but is simply because the validation scripts don't set these values explictly (so take the new defaults)
```
/WCSim/tuning/WCODTyvekInwallReflectivity
/WCSim/tuning/WCODTyvekOutwallReflectivity
```

Will merge & tag

PR #564, opened 2026-06-10
Consistent saving of PMT radii

tdealtry:
OD PMT radius was being saved in mm, whilst others used cm. Now use cm consistently

Additionally the radius for ID PMTs of the second type (aka mPMTs in FD) was previously hardcoded


tdealtry:
I should add that this only applies to the output (geofile & ROOT file). Internally (e.g. for building the geometry) the size has not changed at all (and was always consistent)

tdealtry:
CI has just got changes to the geofile, as expected (physics identical). So will merge & tag

PR #565, opened 2026-07-22
Bugfix nRooTracker truth variables

mscott201:
Added code to copy neutrino and target truth information from nRooTracker files to WCSim RootEvent output.

New code is on line 619 -> 679.


nickwp:
After getting all the CI except the "finalise" to run (and the rest pass, although not 100% sure if correct or not) I can now merge this.
Need access to WCSim/Validation to get the CI to fully work but I don't have access yet, so I'm not sure how this will affect future CI checks.

PR #568, opened 2026-08-04
Add missing WCSIM_HEPMC3_ENABLED preprocessor definition

Abdoelabassi:
## Problem
   When building with `-DWCSIM_HEPMC3_ENABLED=ON`, `find_package(HepMC3)`
   succeeds and the HepMC3 include dirs/libraries are correctly added to
   the build — but the `WCSIM_HEPMC3_ENABLED` preprocessor macro used to
   gate `#ifdef` blocks in `WCSimPrimaryGeneratorAction.hh`/`.cc` is never
   actually defined at compile time. As a result, the HepMC3 reader code
   path is silently compiled out, and at runtime you get:
[WARNING] : HepMC3 events requested, but interface not compiled.
: Use -DWCSISM_HEPMC3_ENABLED=ON at compile time.
even though HepMC3 was found and linked correctly. This manifests as
   the `hepmc3` generator producing zero primary particles, and therefore
   zero digitized hits/triggers downstream.

   ## Fix
   I Added the missing `add_definitions(-DWCSIM_HEPMC3_ENABLED)` call inside
   the `if (WCSIM_HEPMC3_ENABLED)` block in the top-level `CMakeLists.txt`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,6 +58,7 @@ if (NOT ${WCSim_WCSimRoot_only}) # only looking for HepMC3 if WCSim_WCSimRoot_on
   endif (NOT HepMC3_FOUND)
   include_directories (${HEPMC3_INCLUDE_DIR})
   LIST(APPEND PUBLIC_EXT_LIBS ${HEPMC3_LIBRARIES})
+  add_definitions(-DWCSIM_HEPMC3_ENABLED)
 endif ()
 
 endif()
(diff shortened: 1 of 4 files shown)


Abdoelabassi:
Also, there's a typo in -DWCSI<S>M_HEPMC3_ENABLED, it's should writes: ``` -DWCSIM_HEPMC3_ENABLED```

